PR #12, opened 2021-03-06
Add export resource and default role support.

joefruland:


genesyscloud/resource_genesyscloud_tf_export.go
```diff
@@ -0,0 +1,594 @@
...
+	"sync"
+
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+const defaultTfJSONFile = "genesyscloud.tf.json"
```
korbjt, line 22: Minor style thing: this would usually be written as a const block
`const (
    defaultTFJSONFile = "genesyscloud.tf.json"
    defaultTFStateFile  = "terraform.tfstate"
)`

genesyscloud/resource_genesyscloud_tf_export.go
```diff
@@ -0,0 +1,594 @@
...
+			wildcardAttr = prevAttr + "." + "*"
+		}
+
+		if currAttr == "id" {
+			// Strip off IDs from the root resource
+			delete(configMap, currAttr)
+			continue
+		}
+
+		switch val.(type) {
```
korbjt, line 474: It might be worth adding a default entry for unrecognized values, if you want to emit an warning or fail.

genesyscloud/sdk_client_pool.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/platformclientv2"
+)
+
+// SDKClientPool holds a pool of client configs for the Genesys Cloud SDK. One should be
+// acquired at the beginning of any resource operation and released on completion.
+// This has the benefit of ensuring we don't issue too many concurrent requests and also
+// increases throughput as each token will have its own rate limit.
+type SDKClientPool struct {
+	pool chan *platformclientv2.Configuration
```
korbjt, line 18: Did you consider just using a sync.Pool?

genesyscloud/resource_genesyscloud_tf_export.go
```diff
@@ -0,0 +1,594 @@
...
+		// Removes zero values and sets proper reference expressions
+		sanitizeConfigMap(resource.Type, jsonResult, "", exporters, includeStateFile)
+
+		if resourceTypeJSONMaps[resource.Type] == nil {
+			resourceTypeJSONMaps[resource.Type] = make(map[string]jsonMap)
+		}
+		resourceTypeJSONMaps[resource.Type][resource.Name] = jsonResult
+	}
+
+	providerSource := "registry.terraform.io/mypurecloud/genesyscloud"
```
korbjt, line 132: I'd move this stuff off into another function e.g. `func sourceForVersion(version string) string`. Doesn't feel right to have this in the middle of this function.

genesyscloud/sdk_client_pool.go
joefruland: > Did you consider just using a sync.Pool?

I looked at it, but it didn't appear to have a way to set a size limit.

genesyscloud/resource_genesyscloud_tf_export.go
joefruland: There can be ints and floats, there was just nothing to do with them here.

PR #44, opened 2021-05-07
Feat/integrations

WitLi07:
Added support for genesys cloud integration


genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"intended_state": {
+				Description: "Integration state.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "DISABLED",
+			},
+			"integration_type": {
```
joefruland, line 106: The integration types and their descriptions are predefined, correct? Can this just be a string called integration_type_id?

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+
+	return resources, nil
+}
+
+func integrationExporter() *ResourceExporter {
+	return &ResourceExporter{
+		GetResourcesFunc: getAllWithPooledClient(getAllIntegrations),
+		RefAttrs: map[string]*RefAttrSettings{
+			// TODO: Since now it uses jsonencode, might need to change how export works
+			"config.credentials.*": {RefType: "genesyscloud_integration_credentials"},
```
joefruland, line 76: config.credentials is just a string, so I don't think this is actually going to work as-is if it is JSON. You can probably just remove this reference for now.

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+				Optional:    true,
+				Computed:    true,
+			},
+			"advanced": {
+				Description: "Integration advanced config.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+			},
+			"credentials": {
```
joefruland, line 39: Is credentials always a map of string to string? If so, this could be a schema.Map  https://www.terraform.io/docs/extend/schemas/schema-types.html#typemap

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "Integration name.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"intended_state": {
```
joefruland, line 100: Can we define the options for intended_state in the description and add a validator to make sure the config specifies one of them?

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+	intendedState := d.Get("intended_state").(string)
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	integrationAPI := platformclientv2.NewIntegrationsApiWithConfig(sdkConfig)
+
+	if d.HasChange("intended_state") {
+
+		log.Printf("Updating integration %s", name)
+
+		diagErr := retryWhen(isVersionMismatch, func() (*platformclientv2.APIResponse, diag.Diagnostics) {
```
joefruland, line 263: Do the integrations potentially return version mismatch errors? The directory APIs (users, skills, etc.) have this issue due to their database schema, but I would assume integrations are completely different and wouldn't return the same error.

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+
+func deleteIntegration(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	name := d.Get("name").(string)
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	integrationAPI := platformclientv2.NewIntegrationsApiWithConfig(sdkConfig)
+
+	log.Printf("Deleting integration %s", name)
+	return retryWhen(isVersionMismatch, func() (*platformclientv2.APIResponse, diag.Diagnostics) {
+		// Directory occasionally returns version errors on deletes if an object was updated at the same time.
```
joefruland, line 296: I don't believe directory handles integration APIs.

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+				Description: "Integration config name.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"notes": {
+				Description: "Integration config notes.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"properties": {
```
joefruland, line 27: Do the properties and advanced attributes properly handle diffs when the config JSON string doesn't actually change (whitespace, attribute order, etc)? You may want to write a test case to verify. There is a DiffSuppress method `suppressEquivalentJsonDiffs` I added recently for JSON attributes that may help.

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,495 @@
...
+							credentialId := value.(string)
+							credentialMap[k] = platformclientv2.Credentialinfo{Id: &credentialId}
+						}
+					}
+				}
+				config.Credentials = &credentialMap
+
+			}
+
+			diagErr := retryWhen(isVersionMismatch, func() (*platformclientv2.APIResponse, diag.Diagnostics) {
```
joefruland, line 482: Same as above. I wouldn't expect us to need to handle version mismatch errors for the integration APIs.

joefruland (commented):
Please also add this resource to the `generateTfExportResource` method in resource_genesyscloud_tf_export_test.go so we can verify exports work.

genesyscloud/resource_genesyscloud_integration.go
WitLi07: Yep this makes it easier to use. Changed to only use integration type id and removed other attributes.

genesyscloud/resource_genesyscloud_integration.go
WitLi07: I used schema.TypeMap at first, then love jsonencode too much and changed it. 
Changed it back and now it's a Map type.

genesyscloud/resource_genesyscloud_integration.go
WitLi07: Added options and validate function

genesyscloud/resource_genesyscloud_integration.go
WitLi07: Added diffsuppress function, and another testcase to test Raw JSON string

genesyscloud/resource_genesyscloud_integration.go
joefruland: Can you actually set the integration state to "DELETED"? I see it listed in the API doc, but it may be a state that would be returned from a GET after you've called DELETE. Worth trying out to see if we should include it here.

genesyscloud/resource_genesyscloud_tf_export_test.go
```diff
@@ -167,7 +167,8 @@ func generateTfExportResource(
 			"genesyscloud_routing_queue",
 			"genesyscloud_routing_skill",
 			"genesyscloud_routing_utilization",
-			"genesyscloud_routing_wrapupcode"
+			"genesyscloud_routing_wrapupcode",
+			"genesyscloud_integration"
```
joefruland, line 171: Please maintain the alphabetical order of this list.

go.sum
```diff
@@ -268,6 +268,15 @@ github.com/hashicorp/terraform-json v0.10.0 h1:9syPD/Y5t+3uFjG8AiWVPu1bklJD8QB8i
 github.com/hashicorp/terraform-json v0.10.0/go.mod h1:3defM4kkMfttwiE7VakJDwCd4R+umhSQnvJwORXbprE=
 github.com/hashicorp/terraform-plugin-docs v0.4.0 h1:xJIXsMzBFwBvC1zcjoNz743GL2tNEfYFFU9+Hjp4Uek=
 github.com/hashicorp/terraform-plugin-docs v0.4.0/go.mod h1:fKj/V3t45tiXpSlUms/0G4OrBayyWpbUJ4WtLjBkINU=
+github.com/hashicorp/terraform-plugin-go v0.2.1 h1:EW/R8bB2Zbkjmugzsy1d27yS8/0454b3MtYHkzOknqA=
```
joefruland, line 271: If you run `go mod tidy` I think it should clean up the duplicate versions in here.

genesyscloud/provider.go
```diff
@@ -86,6 +86,7 @@ func New(version string) func() *schema.Provider {
 				"genesyscloud_tf_export":               resourceTfExport(),
 				"genesyscloud_user":                    resourceUser(),
 				"genesyscloud_user_roles":              resourceUserRoles(),
+				"genesyscloud_integration":             resourceIntegration(),
```
joefruland, line 89: Please maintain the alphabetical order here

genesyscloud/resource_exporter.go
```diff
@@ -136,6 +136,7 @@ func getResourceExporters(filter []string) map[string]*ResourceExporter {
 		"genesyscloud_routing_wrapupcode":      routingWrapupCodeExporter(),
 		"genesyscloud_user":                    userExporter(),
 		"genesyscloud_user_roles":              userRolesExporter(),
+		"genesyscloud_integration":             integrationExporter(),
```
joefruland, line 139: Please maintain the alphabetical order here

genesyscloud/resource_genesyscloud_integration.go
joefruland: These are a map of any credential keys to credential IDs, correct? We should add this to the RefAttrs below so they are excluded from exports until the credentials type is defined. Since the keys are unknown, we can define the attribute as `credentials.*` and it should work as expected.

WitLi07: image
Added reference for integration config credentials.

examples/resources/genesyscloud_integration/resource.tf
```diff
@@ -0,0 +1,20 @@
+resource "genesyscloud_integration" "integration1" {
+  name             = "test_integration"
+  intended_state   = "DISABLED"
+  integration_type = "embedded-client-app"
+  config {
+    name = "Premium Client Application Example"
```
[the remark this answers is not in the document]
WitLi07: They are the same thing, same attribute. Yeah I also thought this is a little strange when I was testing it. But in the API, there are two names, one in integration and one in integration config, and I just kept it this way, to sync with API settings. Maybe I can get rid of the config name and just use the integration name?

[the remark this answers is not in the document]
joefruland: If they're the same attribute in the backend, we should only expose one.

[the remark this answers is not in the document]
WitLi07: okay, I'll change them. 
Just curious, the integration name, not the config name, actually doesn't do anything except gets updated every time after read. So even users modify that name, it won't do anything. The only way to update the name attribute is through the config name. So basically that integration name attr is a read-only attribute. Does that also fit in this situation and considered as an exposed attribute?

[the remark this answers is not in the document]
joefruland: IMO this is just a case of a confusing API, and since we are a layer above the API we have the ability to deviate from them when they are confusing. In order to make it simpler, we should have one attribute for "name" and when you change it in the config, the name on the backend (and therefore the UI) also changes. If you want to keep the name in the config object that is fine too, just as long as it exists in one place and allows modifying the name.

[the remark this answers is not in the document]
WitLi07: Removed the top-level "name" attribute. Now it only has the one in the integration config.

docs/resources/integration.md
```diff
@@ -0,0 +1,72 @@
...
+- **config** (List of Object) Integration config. Each integration type has different schema, use [GET /api/v2/integrations/types/{typeId}/configschemas/{configType}](https://developer.mypurecloud.com/api/rest/v2/integrations/#get-api-v2-integrations-types--typeId--configschemas--configType-) to check schema, then use the correct attribute names for properties. (see [below for nested schema](#nestedatt--config))
+- **id** (String) The ID of this resource.
+- **intended_state** (String) Integration state (ENABLED | DISABLED | DELETED). Defaults to `DISABLED`.
+
+<a id="nestedatt--config"></a>
+### Nested Schema for `config`
+
+Optional:
+
+- **advanced** (String)
```
joefruland, line 67: The descriptions here are missing because `schema.SchemaConfigModeAttr` is set on the `config` attribute. I think you can just remove `schema.SchemaConfigModeAttr` as it isn't needed here.

genesyscloud/resource_genesyscloud_integration.go
```diff
@@ -0,0 +1,366 @@
...
+			"integration_type": {
+				Description: "Integration type.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"config": {
+				Description: "Integration config. Each integration type has different schema, use [GET /api/v2/integrations/types/{typeId}/configschemas/{configType}](https://developer.mypurecloud.com/api/rest/v2/integrations/#get-api-v2-integrations-types--typeId--configschemas--configType-) to check schema, then use the correct attribute names for properties.",
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
```
joefruland, line 112: Since this contains required attributes like name, this block should also be required now.

WitLi07: I thought it's more like config attribute is optional, but when there is a config attr, then within it the name is required. I checked the user resource and found some [examples](https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/7cab741ec8778433fff441c7fa4bfe5341411f03/genesyscloud/resource_genesyscloud_user.go#L26) of that scenario.  Or I can just change the name to be optional if that doesn't work. Since users can still make request without specifying the name. The API will automatically assign a name anyway. The name attribute will be set when users want to change the name of the integration.

joefruland: Ah, the API docs say the name is required, but I guess that must be wrong: https://developer.genesys.cloud/api/rest/v2/integrations/#post-api-v2-integrations

We should add a test case without the config block set to confirm that works. Also if the API can set the name, the `name` attribute should be `Optional` and `Computed` so it can be excluded from the config if someone just wants to set `properties`

docs/resources/integration.md
WitLi07: Removed. Now it has descriptions for the config.

genesyscloud/resource_genesyscloud_integration.go
WitLi07: Made changes to the `name` attribute and also refactored the `updateIntegrationConfig` function a little bit. Added two more test cases, one with no config block and one with config block but all nullValues.

PR #55, opened 2021-06-01
Add Credential resource

WitLi07:
Added credential resource and tests. Also added credential-related integration tests.


examples/resources/genesyscloud_credential/resource.tf
```diff
@@ -0,0 +1,9 @@
+resource "genesyscloud_credential" "credential1" {
```
joefruland, line 1: Can we make this `genesyscloud_integration_credential` instead to match the API path?

genesyscloud/resource_genesyscloud_credential.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "Credential name.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"credential_type_name": {
+				Description: "Credential type name.",
```
joefruland, line 62: Does this have to come from a pre-defined list of names? If so, we should document where to get the list of possibilities.

genesyscloud/resource_genesyscloud_credential_test.go
```diff
@@ -0,0 +1,161 @@
...
+				),
+			},
+			{
+				// Update name
+				Config: generateCredentialResource(
+					credResource1,
+					strconv.Quote(credName2),
+					strconv.Quote(typeName1),
+					generateCredentialFields(
+						generateMapKeyValue(key1, strconv.Quote(val1)),
```
joefruland, line 59: Can we test updating the fields? That would be the most likely change I'm guessing.

genesyscloud/resource_genesyscloud_credential.go
WitLi07: I added similar notes down below in the description of fields and also included the LIST credential type API in the APIs list. But maybe also mention that in this description, as well?

genesyscloud/resource_genesyscloud_credential_test.go
WitLi07: Yeah I can add some key-value pairs and also change values for the same keyname. But I guess I can change them only to check if the program runs? Since it cannot be set by `d.Set()`, there is no way to actually test it

genesyscloud/resource_genesyscloud_credential.go
joefruland: I would mention it with a link to the API in the actual description, similar to what we did with integration types.

genesyscloud/resource_genesyscloud_credential.go
```diff
@@ -0,0 +1,189 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"fields": {
+				Description: "Credential fields. Different credential types require different fields. Missing any correct required fields will result API request failure. Check out the specific credential type schema to find out what fields are required. ",
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Computed:    true,
```
joefruland, line 70: Since this can contain passwords, we should also add `Sensitive: true` to prevent it from displaying in logs.

examples/resources/genesyscloud_credential/resource.tf
WitLi07: Name Changed

joefruland:
The `RefAttrs` for integrations also .need updated to reference the credential resource

PR #73, opened 2021-07-13
Update queue timeout maximum

Brombeersaft:
Update queue timeout maximum to 259200 seconds (3 days)


joefruland:
Good find. Thanks for the PR!

PR #75, opened 2021-07-19
Adding generic phones resource

ronanwatkins:
This is generic at the moment. I can change it to be specific to WebRTC if that seems better @carnellj


examples/resources/genesyscloud_telephony_providers_edges_phone/resource.tf
```diff
@@ -0,0 +1,20 @@
+resource "genesyscloud_telephony_providers_edges_phone" "test_phone" {
+  name                   = "test phone"
+  state                  = "active"
+  site_id                = "48382f1b-fbbe-4232-8dd6-42a4fa70c1b6"
```
ronanwatkins, line 4: Randomly generated GUIDs. Not sure if I can reference another resource's ID here

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -0,0 +1,393 @@
...
+				Description: "Web RTC User ID.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"capabilities": {
+				Description: "Phone Capabilities.",
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
+				ConfigMode:  schema.SchemaConfigModeAttr,
```
joefruland, line 68: I don't think this ConfigMode setting is needed here. I've used this in the past to allow specifying an empty list to clear a `Computed` attribute, but this attribute isn't computed. It also has the side effect of removing descriptions from the docs.

examples/resources/genesyscloud_telephony_providers_edges_phone/resource.tf
joefruland: That's fine for now. Eventually I would expect sites to be available in the provider, at which point this will be able to reference them.

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -0,0 +1,393 @@
...
+				Description: "Line Base Settings ID.",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"phone_meta_base_id": {
+				Description: "Phone Meta Base ID.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"web_rtc_user_id": {
```
joefruland, line 58: If this attribute references a user ID, it should be added to the `RefAttrs` for this resource so exports properly use a user reference rather than the raw ID.

genesyscloud/provider.go
```diff
@@ -69,34 +69,35 @@ func New(version string) func() *schema.Provider {
...
+				"genesyscloud_routing_email_route":             resourceRoutingEmailRoute(),
+				"genesyscloud_routing_language":                resourceRoutingLanguage(),
+				"genesyscloud_routing_queue":                   resourceRoutingQueue(),
+				"genesyscloud_routing_skill":                   resourceRoutingSkill(),
+				"genesyscloud_routing_utilization":             resourceRoutingUtilization(),
+				"genesyscloud_routing_wrapupcode":              resourceRoutingWrapupCode(),
+				"genesyscloud_tf_export":                       resourceTfExport(),
+				"genesyscloud_user":                            resourceUser(),
+				"genesyscloud_user_roles":                      resourceUserRoles(),
+				"genesyscloud_telephony_providers_edges_phone": resourcePhone(),
```
WitLi07, line 100: Might want to put it in alphabetical order, same goes for the list in exporter.

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone_test.go
```diff
@@ -0,0 +1,281 @@
...
+					capabilities,
+				),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("genesyscloud_telephony_providers_edges_phone."+phoneRes, "name", name2),
+					resource.TestCheckResourceAttr("genesyscloud_telephony_providers_edges_phone."+phoneRes, "state", stateActive),
+					resource.TestCheckResourceAttr("genesyscloud_telephony_providers_edges_phone."+phoneRes, "site_id", siteId),
+					resource.TestCheckResourceAttr("genesyscloud_telephony_providers_edges_phone."+phoneRes, "phone_base_settings_id", phoneBaseSettingsId),
+					resource.TestCheckResourceAttr("genesyscloud_telephony_providers_edges_phone."+phoneRes, "web_rtc_user_id", webRtcUserId2),
+				),
+			},
```
WitLi07, line 131: Might want to add an import/read test step

WitLi07:
Should also add the telephony resource into the list in `resource_genesyscloud_tf_exporter_test.go` file

examples/resources/genesyscloud_telephony_providers_edges_phone/resource.tf
carnellj: > This is generic at the moment. I can change it to be specific to WebRTC if that seems better @carnellj

@ronanwatkins Keep it generic for now.  We should do a blog post after the provider is released showing how to setup a WebRTC phone

joefruland (approved):

PR #80, opened 2021-07-21
CXCODE-20 Add DID Pool Resource

PrinceMerluza:


joefruland (approved):

joefruland:
Please resolve the merge conflicts, but overall this looks good to me.

PrinceMerluza:
> Please resolve the merge conflicts, but overall this looks good to me.

Thanks Joe. I resolved the merge conflicts.

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+		return diag.Errorf("Failed to read DID pool %s: %s", d.Id(), getErr)
+	}
+
+	if *didPool.State == "deleted" {
+		d.SetId("")
+		return nil
+	}
+
+	// d.Set("start_phone_number", *didPool.StartPhoneNumber)
```
ronanwatkins, line 138: Could comment this out

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,214 @@
...
+	// d.Set("end_phone_number", *didPool.EndPhoneNumber)
+
+	if didPool.StartPhoneNumber != nil {
+		d.Set("start_phone_number", *didPool.StartPhoneNumber)
+	} else {
+		d.Set("start_phone_number", nil)
```
ronanwatkins, line 144: Not sure if it's necessary to set properties to be nil. I could be wrong though

PrinceMerluza: Oh yea, this and the endphoneNumber should always be set. I did this when I was debugging an issue. I'll delete this and uncomment the Set lines for start and end phoneNumbers.

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil
+	}
+
+	d.Set("start_phone_number", *didPool.StartPhoneNumber)
+	d.Set("end_phone_number", *didPool.EndPhoneNumber)
+
+	if didPool.Description != nil {
+		d.Set("description", *didPool.Description)
+	} else {
+		d.Set("description", nil)
```
carnellj, line 144: Just out of curiosity why are we explicitly setting the resource value to nil?   Looks funny.  Could you provide a little bit more background on this.

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	if didPool.Comments != nil {
+		d.Set("comments", *didPool.Comments)
+	} else {
+		d.Set("comments", nil)
+	}
+
+	if didPool.Provider != nil {
+		d.Set("pool_provider", *didPool.Provider)
```
carnellj, line 154: Formatting looks off on this.  Make sure you run go fmt on the file.  Go normally ident so many spaces over.

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,202 @@
...
+
+	log.Printf("Read DID pool %s %s", d.Id(), *didPool.StartPhoneNumber)
+	return nil
+}
+
+func updateDidPool(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	startPhoneNumber := d.Get("start_phone_number").(string)
+	endPhoneNumber := d.Get("end_phone_number").(string)
+	description := d.Get("description").(string)
+	comments := d.Get("comments").(string)
```
carnellj, line 167: Not familiar with what you are doing with .string.   (Doesnt mean its wrong, I have just never seen it before :) ) Is that meant to cast the return object as a string.  If so, what happens if anyone of these values is nil.  Will you get a error and a stack dump.

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func deleteDidPool(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	startPhoneNumber := d.Get("start_phone_number").(string)
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	telephonyApi := platformclientv2.NewTelephonyProvidersEdgeApiWithConfig(sdkConfig)
+
+	log.Printf("Deleting DID pool with starting number %s", startPhoneNumber)
+	_, err := telephonyApi.DeleteTelephonyProvidersEdgesDidpool(d.Id())
```
carnellj, line 196: Consider combining your if then err checks into something like this.

if _,err := doStuff(); err != nil {
  // handle the error here
}

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,202 @@
...
+	_, _, err := telephonyApi.PutTelephonyProvidersEdgesDidpool(d.Id(), platformclientv2.Didpool{
+		StartPhoneNumber: &startPhoneNumber,
+		EndPhoneNumber:   &endPhoneNumber,
+		Description:      &description,
+		Comments:         &comments,
+		Provider:         &poolProvider,
+	})
+	if err != nil {
+		return diag.Errorf("Error updating DID pool %s: %s", startPhoneNumber, err)
+	}
```
carnellj, line 183: I would consider breaking the object initialization out of the  method call.  It makes the code easier to reader and then you can combine the err checking in a single line:

if _,_,err := telephonyApi.PutTelephonyProvidersEdgesDidpool(d.Id(),didPool); err != nil {
  // handle the error here
}

genesyscloud/resource_genesyscloud_telephony_didpool.go
```diff
@@ -0,0 +1,202 @@
...
+	endPhoneNumber := d.Get("end_phone_number").(string)
+	description := d.Get("description").(string)
+	comments := d.Get("comments").(string)
+	poolProvider := d.Get("pool_provider").(string)
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	telephonyApi := platformclientv2.NewTelephonyProvidersEdgeApiWithConfig(sdkConfig)
+
+	log.Printf("Updating DID pool %s", d.Id())
+	_, _, err := telephonyApi.PutTelephonyProvidersEdgesDidpool(d.Id(), platformclientv2.Didpool{
```
carnellj, line 174: @ronanwatkins Why is this function returning three values?  The normal golang paradigm is to do a return value for the function call and then the err code.  Just curious.

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/mypurecloud/platform-client-sdk-go/v48/platformclientv2"
+)
+
+func TestAccResourceDidPoolBasic(t *testing.T) {
+	var (
```
carnellj, line 14: The var here is extraneous and is usually not done in most go code I see.  I would do:

didPoolResource1         := "test-didpool1"
didPoolStartPhoneNumber1 := "+13175550000"
didPoolEndPhoneNumber1   := "+13175550005"
didPoolDescription1      := "Test description"
didPoolComments1         := "Test comments"
didPoolProvider1         := "PURE_CLOUD"

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
```diff
@@ -0,0 +1,115 @@
...
+		didPoolResource1         = "test-didpool1"
+		didPoolStartPhoneNumber1 = "+13175550000"
+		didPoolEndPhoneNumber1   = "+13175550005"
+		didPoolDescription1      = "Test description"
+		didPoolComments1         = "Test comments"
+		didPoolProvider1         = "PURE_CLOUD"
+	)
+
+	resource.Test(t, resource.TestCase{
```
carnellj, line 23: Love table style test cases.

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
```diff
@@ -0,0 +1,115 @@
...
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+		CheckDestroy: testVerifyDidPoolsDestroyed,
+	})
+}
+
+func generateDidPoolResource(
+	resourceID string,
```
carnellj, line 75: Once you start getting a large number of values being passed into a function you are better off defining a struct that wrappers all these values.  Otherwise it becomes unreadable :)

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
```diff
@@ -0,0 +1,115 @@
...
+		comments = %s
+		pool_provider = %s
+	}
+	`, resourceID, startPhoneNumber, endPhoneNumber, description, comments, poolProvider)
+}
+
+func testVerifyDidPoolsDestroyed(state *terraform.State) error {
+	telephonyAPI := platformclientv2.NewTelephonyProvidersEdgeApi()
+	for _, rs := range state.RootModule().Resources {
+		if rs.Type != "genesyscloud_telephony_did_pool" {
```
carnellj, line 94: From a readability perspective, GoLang is very big on avoiding nest if.. then..else statements.  Instead, look for how you can refactor the code below to use a single line of if..then statements that either break or return.  You should be able to see a straight line all the way down your code with your conditionals.  Take a look at this blog piece.

https://www.geeksforgeeks.org/writing-clean-else-statements/

carnellj (changes requested):
Overall really good first pass on your code @PrinceMerluza.  We will make a go coder out of you yet :)

genesyscloud/resource_genesyscloud_telephony_didpool.go
joefruland: I use this pattern in most places for consistency. Terraform recommends the read method is called at the end of create/updates to ensure the state is in sync with the server after an apply: https://learn.hashicorp.com/tutorials/terraform/provider-create?in=terraform/providers#add-read-function-to-create-function

In order to ensure the values returned through the ResourceData object match what is returned by the read, it sometimes means that a value that was previously set on the ResourceData object must be explicitly set to nil.

genesyscloud/resource_genesyscloud_telephony_didpool.go
joefruland: Yes, this casts to a string. Terraform does not have a concept of nil values, so it populates all attributes with their zero value in the ResourceData (empty string in this case), so I believe the only time this would return nil is if the attribute is not defined or it is not defined as a string.

carnellj: Cool!   Thanks @joefruland thats what I was looking for.

genesyscloud/resource_genesyscloud_telephony_didpool.go
carnellj: Cool.  Thats exactly what I needed to understand

genesyscloud/resource_genesyscloud_telephony_didpool.go
PrinceMerluza: Hi John, I don't know about the Golang best practice but the three return values are for the returned Genesys Cloud Object, the ApiResponse, and the error.

genesyscloud/resource_genesyscloud_telephony_didpool.go
PrinceMerluza: I might have misunderstood the question, Joe has a comment on this on why properties are being set to nil.

genesyscloud/resource_genesyscloud_telephony_didpool.go
PrinceMerluza: I ran gofmt for the files but didn't really notice any difference in indentation.

genesyscloud/resource_genesyscloud_telephony_didpool.go
PrinceMerluza: Nice, that's a very handy trick. Thanks

genesyscloud/resource_genesyscloud_telephony_didpool.go
PrinceMerluza: Initiated the request body outside the method.

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
PrinceMerluza: Hi John the var was for aligning the assignment operator but I changed it for now to remove the var grouping

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
PrinceMerluza: Good point. I created a struct and use a pointer parameter instead.

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
PrinceMerluza: Thanks for the tip John, I agree it's more readable that way. I've refactored them to separate 'if block' statements

PrinceMerluza:
> Overall really good first pass on your code @PrinceMerluza. We will make a go coder out of you yet :)

Thanks ! Though admittedly I have the advantage of learning from the other sample files already in the project.

genesyscloud/resource_genesyscloud_telephony_didpool_test.go
carnellj: No worries.  What you did wasnt wrong, it just isnt idiomatic go.

genesyscloud/resource_genesyscloud_telephony_didpool.go
carnellj: @PrinceMerluza That is why I directed it to Ronan.  I was curious as to why this was being done this within our SDK.  Not wrong, just curious :)

carnellj:
> > Overall really good first pass on your code @PrinceMerluza. We will make a go coder out of you yet :)
> 
> Thanks ! Though admittedly I have the advantage of learning from the other sample files already in the project.

Dont undersell yourself Prince.  I think you are a talented engineer and I love working with you.

carnellj (approved):

ronanwatkins (approved):

PR #84, opened 2021-08-18
CXCODE-19 Enable Provisioning of DIDs.

PrinceMerluza:
- Added IVR Config (inbound call routing) resource
- Modified Phone resource to enable a standalone configuration which requires DID per Line

No changes to User resource at it already supports DID assignment


joefruland:
Since this was merged from a fork instead of a branch, the acceptance tests did not run. This is on purpose to prevent external users from being able to submit PRs with malicious code that would have access to our GitHub secrets.

PR #85, opened 2021-08-24
CXCODE-19 DID Provisioning

PrinceMerluza:
A new PR since the old one was from a fork of the repo.
The files should still be the same.

- Add IVR config resource
- Modify Phone resource to allow Line Keys to be set for standalone phones


ronanwatkins (approved):

WitLi07 (approved):

AgnesCorpuz (approved):

examples/data-sources/genesyscloud_routing_email_domain/data-source.tf
```diff
@@ -1,4 +1,4 @@
 data "genesyscloud_routing_email_domain" "email_domain" {
```
ronanwatkins, line 1: Just realised you'll have to run `go generate` to generate files in the docs directory that will show how to use the data source and resource by combining their .tf file and apis.md files

genesyscloud/resource_genesyscloud_architect_ivr.go
```diff
@@ -0,0 +1,247 @@
...
+		d.Set("schedule_group_id", nil)
+	}
+
+	log.Printf("Read IVR config %s %s", d.Id(), *ivrConfig.Name)
+	return nil
+}
+
+func updateIvrConfig(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	name := d.Get("name").(string)
+	description := d.Get("description").(string)
```
ronanwatkins, line 193: It's best to set optional properties like description only if they're not empty strings to avoid overwriting a description with an empty one

genesyscloud/resource_genesyscloud_architect_ivr.go
```diff
@@ -0,0 +1,247 @@
...
+	openHoursFlowId := buildSdkDomainEntityRef(d, "open_hours_flow_id")
+	closedHoursFlowId := buildSdkDomainEntityRef(d, "closed_hours_flow_id")
+	holidayHoursFlowId := buildSdkDomainEntityRef(d, "holiday_hours_flow_id")
+	scheduleGroupId := buildSdkDomainEntityRef(d, "schedule_group_id")
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	architectApi := platformclientv2.NewArchitectApiWithConfig(sdkConfig)
+
+	ivrBody := platformclientv2.Ivr{
```
ronanwatkins, line 202: You'll need to fetch the latest version when doing an update. See https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/resource_genesyscloud_group.go#L235 for the pattern

genesyscloud/resource_genesyscloud_architect_ivr_test.go
```diff
@@ -0,0 +1,155 @@
...
+		if ivrConfig != nil && ivrConfig.State != nil && *ivrConfig.State == "deleted" {
+			continue
+		}
+
+		if ivrConfig != nil {
+			return fmt.Errorf("IVR config (%s) still exists", rs.Primary.ID)
+		}
+
+		if resp != nil && resp.StatusCode == 404 {
+			// DID pool not found as expected
```
ronanwatkins, line 120: Looks like a few comments left over from DID pools :)

examples/data-sources/genesyscloud_routing_email_domain/data-source.tf
PrinceMerluza: Added doc file

genesyscloud/resource_genesyscloud_architect_ivr.go
PrinceMerluza: Added check for empty string on create and update methods.

genesyscloud/resource_genesyscloud_architect_ivr.go
PrinceMerluza: Oh thanks for the catch. Looks like the API doesn't require version (unlike groups and users) when updating, so I forgot there is a version property. Added.

genesyscloud/resource_genesyscloud_architect_ivr_test.go
PrinceMerluza: Oops. Fixed.

carnellj (approved):

PrinceMerluza:
Resolved merge conflicts

PR #94, opened 2021-09-20
Adding trunks

ronanwatkins:
Trunks are created by assigning trunk base settings to an edge or edge group


genesyscloud/resource_genesyscloud_telephony_providers_edges_trunk.go
```diff
@@ -0,0 +1,202 @@
...
+		},
+	}
+}
+
+func getAllTrunks(ctx context.Context, sdkConfig *platformclientv2.Configuration) (ResourceIDMetaMap, diag.Diagnostics) {
+	resources := make(ResourceIDMetaMap)
+
+	edgesAPI := platformclientv2.NewTelephonyProvidersEdgeApiWithConfig(sdkConfig)
+
+	for pageNum := 1; ; pageNum++ {
```
carnellj, line 184: Why not use a range instead of this.  ^^^^

carnellj (approved):

genesyscloud/resource_genesyscloud_telephony_providers_edges_trunk.go
ronanwatkins: Yeah that could be done using a do while. For consistency it should be done to all resources if we are to go that route but there's not much need since the loop breaks when there are no more entities

PR #98, opened 2021-09-27
Adding phone base settings resource and data source

ronanwatkins:


genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -234,6 +246,37 @@ func readPhone(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 	return nil
 }
 
+func assignUserToWebRtcPhone(ctx context.Context, sdkConfig *platformclientv2.Configuration, userId string) diag.Diagnostics {
+	stationsAPI := platformclientv2.NewStationsApiWithConfig(sdkConfig)
+	stationId := ""
+
+	retryErr := withRetries(ctx, 15*time.Second, func() *resource.RetryError {
+		stations, _, getErr := stationsAPI.GetStations(100, 1, "", "", "", userId, "", "")
```
carnellj, line 254: Why is the value of 100 set here.  Even if it is within a function we should not be using"magic numbers".  Instead use a constant to describe what the value is for.

genesyscloud/resource_genesyscloud_telephony_providers_edges_phonebasesettings.go
```diff
@@ -0,0 +1,261 @@
...
+	if description != "" {
+		phoneBase.Description = &description
+	}
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	edgesAPI := platformclientv2.NewTelephonyProvidersEdgeApiWithConfig(sdkConfig)
+
+	phoneBaseSettings, resp, getErr := edgesAPI.GetTelephonyProvidersEdgesPhonebasesetting(d.Id())
+	if getErr != nil {
+		if resp != nil && resp.StatusCode == 404 {
```
carnellj, line 137: Go's http package has a set of constants for this.  I personally would use this rather then a 404.  Small thing.

genesyscloud/resource_genesyscloud_telephony_providers_edges_phonebasesettings.go
```diff
@@ -0,0 +1,261 @@
...
+	})
+}
+
+func getAllPhoneBaseSettings(ctx context.Context, sdkConfig *platformclientv2.Configuration) (ResourceIDMetaMap, diag.Diagnostics) {
+	resources := make(ResourceIDMetaMap)
+
+	edgesAPI := platformclientv2.NewTelephonyProvidersEdgeApiWithConfig(sdkConfig)
+
+	for pageNum := 1; ; pageNum++ {
+		phoneBaseSettings, _, getErr := edgesAPI.GetTelephonyProvidersEdgesPhonebasesettings(pageNum, 100, "", "", nil, "")
```
carnellj, line 237: see my earlier comments about magic numbers

carnellj (approved):

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
ronanwatkins: They are used for all calls to list resources (usually page number and size) so that change should be made across the board if we don't want magic numbers.

genesyscloud/resource_genesyscloud_telephony_providers_edges_phonebasesettings.go
ronanwatkins: Yeah fair point. I'll make a ticket to have magic numbers for HTTP codes and API calls removed

PR #122, opened 2021-10-14
CXCODE-38 Architect User Prompt Resource

PrinceMerluza:
CXCODE-38 Architect User Prompt Resource


genesyscloud/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -0,0 +1,1081 @@
...
+
+	return readUserPrompt(ctx, d, meta)
+}
+
+func readUserPrompt(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	architectAPI := platformclientv2.NewArchitectApiWithConfig(sdkConfig)
+
+	log.Printf("Reading User Prompt %s", d.Id())
+	userPrompt, resp, getErr := architectAPI.GetArchitectPrompt(d.Id())
```
ronanwatkins, line 827: Could you please add some retry to this read, you can look at how I added it to other resources. It avoids issues where the resource is created and the read returns 404 because of how dynamodb eventual consistency works.

genesyscloud/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -0,0 +1,1081 @@
...
+	if err != nil {
+		return diag.Errorf("Failed to update user prompt %s: %s", name, err)
+	}
+
+	diagErr := updatePromptResource(d, architectApi, sdkConfig)
+	if diagErr != nil {
+		return diagErr
+	}
+
+	log.Printf("Updated User Prompt %s", d.Id())
```
ronanwatkins, line 878: Also, a sleep of 5 seconds here so the read is guaranteed to return up to date data.

genesyscloud/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -0,0 +1,1081 @@
...
+	})
+}
+
+func uploadPrompt(uploadUri *string, filename *string, sdkConfig *platformclientv2.Configuration) (string, []byte) {
+	accessToken := sdkConfig.AccessToken
+
+	client := &http.Client{}
+	file, err := os.Open(*filename)
+	if err != nil {
+		log.Fatal(err)
```
ronanwatkins, line 914: It's probably better practice to have this function return the error and the caller handle it by returning it in the diag.Errorf function

genesyscloud/resource_genesyscloud_architect_user_prompt_test.go
```diff
@@ -0,0 +1,243 @@
...
+type userPromptResourceStruct struct {
+	language   string
+	tts_string string
+	text       string
+	filename   string
+}
+
+func TestAccResourceUserPromptBasic(t *testing.T) {
+	userPromptResource1 := "test-user_prompt_1"
+	userPromptName1 := "TestUserPrompt_1"
```
ronanwatkins, line 29: This should be concatenated with something like `strings.Replace(uuid.NewString(), "-", "", -1)` to avoid potential conflicts

genesyscloud/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -0,0 +1,1081 @@
...
+		DeleteContext: deleteWithPooledClient(deleteUserPrompt),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "Name of the user audio prompt.",
+				Type:        schema.TypeString,
+				Required:    true,
```
ronanwatkins, line 734: Pretty sure this should have ForceNew: true

PR #165, opened 2021-11-03
CXCODE-37 Evaluation Form

PrinceMerluza:
Add Evaluation Form resource + data source


genesyscloud/resource_genesyscloud_quality_forms_evaluation.go
```diff
@@ -0,0 +1,591 @@
...
+		})
+		if err != nil {
+			return diag.Errorf("Failed to publish evaluation form %s", name)
+		}
+	} else {
+		// If published property is reset to false, set the resource Id to the latest unpublished form
+		d.SetId(*form.Id)
+	}
+
+	log.Printf("Updated evaluation form %s %s", name, *form.Id)
```
ronanwatkins, line 332: It would be good to put in a sleep here of 5 seconds. The old form might be cached/or dynamo eventual consistency could cause the read to return stale data

PrinceMerluza: + Added 5s sleep

PR #180, opened 2021-12-29
Feat/extension pools

segestaconsult:
Added resource genesyscloud_telephony_providers_edges_extension_pool
It is also configured to be exportable via tf_export.
This is largely a copy of the did_pool resource, with some tweaking (API calls are a bit different).


genesyscloud/resource_genesyscloud_telephony_providers_edges_extension_pool_test.go
```diff
@@ -0,0 +1,143 @@
...
+	extensionPoolResource1 := "test-extensionpool1"
+	rand.Seed(time.Now().Unix())
+	n := rand.Intn(9)
+	extensionPoolStartNumber1 := fmt.Sprintf("1500%v", n)
+	extensionPoolEndNumber1 := fmt.Sprintf("1509%v", n+1)
+	err := authorizeSdk()
+	if err != nil {
+		t.Fatal(err)
+	}
+	deleteExtensionPoolWithNumber(extensionPoolStartNumber1)
```
ronanwatkins, line 32: Thanks for putting in this cleanup step!

ronanwatkins (approved):
This is great, thanks for contributing.

examples/resources/genesyscloud_telephony_providers_edges_extension_pool/resource.tf
```diff
@@ -0,0 +1,5 @@
+resource "genesyscloud_telephony_providers_edges_extension_pool" "test_extension_pool" {
+  start_phone_number = "1000"
```
ronanwatkins, line 2: There's a mismatch between the `start_phone_number` and `end_phone_number` values. They're defined in the schema as `start_number`/`end_number` and used in the tests as `start_phone_number`.
Also they are `startNumber` and `endNumber` in the data source example.

Using `start_phone_number`/`end_phone_number` would be best as it would align with DID pools.

segestaconsult: Thank you for the review. I have resolved this is last commit.
I didn't use phone_number as I wanted to make a clear distinction between phone_number (I see this as a number reachable from outside) and number (extension only usable inside). If you catch my reasoning ;-)

PR #191, opened 2022-01-24
Add Division ID

AgnesCorpuz:


genesyscloud/resource_genesyscloud_architect_schedulegroups.go
```diff
@@ -144,6 +157,7 @@ func readArchitectScheduleGroups(ctx context.Context, d *schema.ResourceData, me
 		}
 
 		d.Set("name", *scheduleGroup.Name)
+		d.Set("division_id", *scheduleGroup.Division.Id)
```
ronanwatkins, line 160: Is this always guaranteed to bet set at this point?

genesyscloud/resource_genesyscloud_architect_schedules_test.go
```diff
@@ -41,6 +41,7 @@ func TestAccResourceArchitectSchedules(t *testing.T) {
 					resource.TestCheckResourceAttr("genesyscloud_architect_schedules."+schedResource1, "start", start),
 					resource.TestCheckResourceAttr("genesyscloud_architect_schedules."+schedResource1, "end", end),
```
ronanwatkins, line 42: Could you add a test case creating a schedule with a division (genesyscloud_auth_division) to confirm that this change works?

genesyscloud/resource_genesyscloud_architect_schedulegroups.go
AgnesCorpuz: I haven't encountered any issue when testing with adding new division and with using home division.

genesyscloud/resource_genesyscloud_architect_schedules_test.go
AgnesCorpuz: Added a test case in recent commit.

ronanwatkins (approved):

ronanwatkins:
I think you should follow the test failure up with the architect team. I've used OAuth clients in different orgs with the `routing:schedule:add` permission but I keep getting the same error

AgnesCorpuz:
I was also getting the same permission error before but was able to resolve it by adding all divisions and future divisions in my Genesys Cloud user and the same with the OAuth client credentials I was using.

PR #200, opened 2022-02-04
CXCODE-119 - Adding support for web deployment resources

ininrobu:
This adds CX as Code support for web deployment configurations and deployments. They are typically used to set up web messaging on a customer's site, although they are used for more than just that - they are used to configure GPE's journey tracking and in the not-too-distant future will be used to configure cobrowse as well.

These resources are a bit unusual in that it's not your typical simple CRUD-based usage. A Configuration holds the bulk of the, well, configuration, and must be published before it can be used in a Deployment. That publish/versioning concept is tough to reconcile with TF's typical usage so I've taken some liberties. Rather than attempt to expose the versioning concept directly, only the latest version of a configuration is affected by TF. When a configuration is updated in any way, a new version is published automatically. Likewise, when defining a Deployment a Configuration must referenced, both its id and version. The expectation is that users will typically reference a Configuration resource directly, which will ensure the Deployment stays up to date, using the latest published Configuration. However, I chose not to hide the versioning concept completely in the context of the Deployment, so users do have the flexibility to reference a specific version should they choose.

Much of the unusual logic in this PR is a result of that reconciliation. That said, don't hesitate to speak up if you think there's a better way to represent and manage these resources. This is also my first foray into TF and Golang, so don't hesitate to pick nits - if I've failed to match coding styles or conventions or there are more idiomatic ways things could be done please let me know.


genesyscloud/resource_genesyscloud_webdeployments_configuration.go
```diff
@@ -0,0 +1,888 @@
...
+				ValidateFunc: validation.StringInSlice([]string{
+					"Pending",
+					"Active",
+					"Inactive",
+					"Error",
+					"Deleting",
+				}, false),
+				DiffSuppressFunc: validateConfigurationStatusChange,
+			},
+			"version": {
```
ronanwatkins, line 316: The version is usually left out of resources and the latest version is retrieved from the API using a GET call prior to updating it. You can see that pattern applied in other resources

genesyscloud/resource_genesyscloud_webdeployments_configuration.go
```diff
@@ -0,0 +1,888 @@
...
+
+	configuration, _, err := api.PostWebdeploymentsConfigurationVersionsDraftPublish(d.Id())
+	if err != nil {
+		return diag.Errorf("Error publishing web deployment configuration %s: %s", name, err)
+	}
+	d.Set("version", configuration.Version)
+	d.Set("status", configuration.Status)
+
+	log.Printf("Finished updating web deployment configuration %s", name)
+	//	time.Sleep(5 * time.Second)
```
ronanwatkins, line 721: This could be uncommented. Eventual consistency usually causes issues when we update and try to read a resource straight away

genesyscloud/resource_genesyscloud_webdeployments_configuration_test.go
```diff
@@ -0,0 +1,240 @@
...
+		configName        = "Test Configuration " + randString(8)
+		configDescription = "Test Configuration description " + randString(32)
+		fullResourceName  = "genesyscloud_webdeployments_configuration.basic"
+	)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
```
ronanwatkins, line 24: It's best to have one or more config blocks with different values in order to test that the resource can be successfully updated

genesyscloud/resource_genesyscloud_webdeployments_deployment.go
```diff
@@ -0,0 +1,313 @@
...
+		inputDeployment.Flow = flow
+	}
+
+	_, _, err = api.PutWebdeploymentsDeployment(d.Id(), inputDeployment)
+	if err != nil {
+		return diag.Errorf("Error updating web deployment %s: %s", name, err)
+	}
+
+	log.Printf("Finished updating web deployment %s", name)
+	//time.Sleep(5 * time.Second)
```
ronanwatkins, line 269: Same thing here regarding sleeping after updating

examples/resources/genesyscloud_webdeployments_deployment/resource.tf
```diff
@@ -0,0 +1,18 @@
...
+
+data "genesyscloud_webdeployments_configuration" "exampleConfiguration" {
+  name = "Example Web Deployment Configuration"
+}
+
+resource "genesyscloud_webdeployments_deployment" "exampleDeployment" {
+  name            = "Example Web Deployment"
+  description     = "This is an example of a web deployment"
+  allowed_domains = ["genesys.com"]
+  flow_id         = genesyscloud_flow.incomingMessageFlow.id
```
ronanwatkins, line 13: This should be data.incomingMessageFlow.id. Same for all examples

genesyscloud/resource_genesyscloud_webdeployments_deployment_test.go
```diff
@@ -0,0 +1,213 @@
+package genesyscloud
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
```
ronanwatkins, line 6: This log import is unused and giving me a compile error because Go is fussy like that :)

genesyscloud/resource_genesyscloud_webdeployments_configuration.go
ininrobu: I believe it's necessary in this case since it needs to be referenced by the deployment, but since it's computed it functions similarly to how you describe. The difference being that updates don't automatically use the latest version. I could switch it to do that though.

ronanwatkins: Oh yeah sorry, the other resources use versions a bit differently. Just put it as `Optional: true` and it'll be fine. I tried exporting some configurations using the current schema and then applying them and got an error because of that.

PR #221, opened 2022-03-15
Adding a solution for eventual consistency

ronanwatkins:
The biggest change is the addition of the consistency_checker.go file. This is used in the read hook of resources inside the withRetriesForRead retry logic. It checks to see if the state of a resource matches what it should be after reading the resource and returns a retryable error if the state doesn't match. This has allowed me to remove lots of sleeps after updating resources.
It doesn't work correctly for resources with nested Computed attributes and when the tests do a refresh or ImportStateVerify after the apply. It's probably possible to fix these issues but I don't have the time.
I've created the following tickets for them to be followed up on:
https://inindca.atlassian.net/browse/DEVENGAGE-1083
https://inindca.atlassian.net/browse/DEVENGAGE-1084


genesyscloud/sdk_client_pool.go
```diff
@@ -44,21 +44,41 @@ func InitSDKClientPool(max int, version string, providerConfig *schema.ResourceD
...
+	for i := 0; i < cap(p.pool); i++ {
 		sdkConfig := platformclientv2.NewConfiguration()
-		err := initClientConfig(providerConfig, version, sdkConfig)
-		if err != nil {
-			return err
-		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err := initClientConfig(providerConfig, version, sdkConfig)
```
carnellj, line 58: What are you doing here Ronan?  Is this taking a pre-configured list of clientConfigs from a pool?  If it is, Nice?

genesyscloud/consistency_checker/consistency_checker.go
```diff
@@ -0,0 +1,285 @@
...
+	diff, _ := c.r.SimpleDiff(c.ctx, c.d.State(), resourceConfig, c.meta)
+	if diff != nil {
+		//fmt.Println(diff)
+		for k, v := range diff.Attributes {
+			if strings.HasSuffix(k, "#") {
+				continue
+			}
+
+			//fmt.Println("!c.d.HasChange(k)", !c.d.HasChange(k))
+			//fmt.Println("isNilOrEmpty(c.d.Get(k))", isNilOrEmpty(c.d.Get(k)))
```
carnellj, line 241: We should remove this commented code.  If its is logging information we should use the Go log function

genesyscloud/sdk_client_pool.go
ronanwatkins: Basically it's got a channel with a list of SDK configs and it's authenticating each one. My updates to this code are to make it multi threaded to speed it up.

PR #237, opened 2022-04-05
Feat/intindy 2216 create trigger terraform provider

Tyler-Bath:
Creating the terraform provider for the Process Automation service trigger objects. This service is currently in Beta and is marked as such on the description of the resource.


genesyscloud/data_source_genesyscloud_process_automation_trigger.go
```diff
@@ -0,0 +1,100 @@
...
+
+			for _, trigger := range *processAutomationTriggers.Entities {
+				if trigger.Name != nil && *trigger.Name == triggerName {
+					d.SetId(*trigger.Id)
+					return nil
+				}
+			}
+
+			if processAutomationTriggers.NextUri == nil {
+                return resource.NonRetryableError(fmt.Errorf("no process automation triggers found with name: %s", getErr))
```
ronanwatkins, line 59: The formatting is off in most files. You can use [go fmt](https://go.dev/blog/gofmt) to format them

docs/data-sources/process_automation_trigger.md
```diff
@@ -0,0 +1,32 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "genesyscloud_process_automation_trigger Data Source - terraform-provider-genesyscloud"
```
ronanwatkins, line 3: The naming convention for resources only puts underscores between names if there is a forward slash between them in the API path, so this will need to be `genesyscloud_processautomation_trigger` in all files and references.

docs/resources/process_automation_trigger.md
```diff
@@ -0,0 +1,60 @@
...
+```terraform
+resource "genesyscloud_process_automation_trigger" "test-trigger" {
+  name       = "Test Trigger"
+  topic_name = "v2.detail.events.conversation.{id}.customer.end"
+  enabled    = true
+  target = jsonencode({
+    id : "ae1e0cde-875d-4d13-a498-615e7a9fe956",
+    type : "Workflow"
+  })
+  match_criteria = jsonencode([
```
ronanwatkins, line 34: Does match_criteria always have these 3 properties? Looks like it's a `MatchCriteria[]`. Therefore it should be of type Set containing a `schema.Resource` in the Elem value. You can see other nested objects like this in other resources.

Same goes for the `target` attribute.

Tyler-Bath: The matchCriteria can change, it will. have either a value field or a values field. Target should always have the same 2 fields

ronanwatkins: Yeah they should both be Sets then containing nested objects. value and values in matchCriteria can be optional

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -0,0 +1,583 @@
...
+                                    Operator:       &operator,
+                                    Values:         &values,
+                                }
+
+			        matchCriteriaObjectList = append(matchCriteriaObjectList, criteria)
+                }
+
+			}
+
+			log.Printf("LIST OF MATCH CRITERIA %v", matchCriteriaObjectList)
```
ronanwatkins, line 367: Best to leave this out

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -0,0 +1,583 @@
...
+
+                if(len(values) < 1){
+                    criteria := MatchCriteria{
+                                    JsonPath:       &jsonPath,
+                                    Operator:       &operator,
+                                    Value:          &value,
+                                }
+
+			        matchCriteriaObjectList = append(matchCriteriaObjectList, criteria)
+                }else{
```
ronanwatkins, line 355: Still need to run go fmt on all files

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -0,0 +1,583 @@
...
+                s[i] = v
+            }
+		    flattendedMatchCriteria["values"] = s
+		}
+		matchCriteriaSet.Add(flattendedMatchCriteria)
+	}
+	return matchCriteriaSet
+}
+
+func flattenTriggerJsonField(jsonField interface{}) (string, diag.Diagnostics) {
```
ronanwatkins, line 419: Looks like this is unused

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -0,0 +1,583 @@
...
+			return resp, diag.Errorf("Failed to update process automation trigger %s: %s match_criteria:%#v", name, err, buildMatchCriteria(d))
+		}
+		return resp, nil
+	})
+	if diagErr != nil {
+		return diagErr
+	}
+
+	log.Printf("Updated process automation trigger %s", name)
+	time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
ronanwatkins: For this resource you can initialise the consistency checker with `cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, resourceProcessAutomationTrigger())`

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -0,0 +1,583 @@
...
+        Schema: map[string]*schema.Schema{
+            "json_path": {
+                Description: "The json path of the topic event to be compared to match criteria value",
+                Type:           schema.TypeString,
+                Required:       true,
+            },
+            "operator": {
+                Description:  "The operator used to compare the json path against the value of the match criteria",
+                Type:           schema.TypeString,
+                Required:       true,
```
ronanwatkins, line 77: You can add a validationFunc for `validations.StringInSlice` and check for `"GreaterThanOrEqual",
                        "LessThanOrEqual",
                        "Equal",
                        "NotEqual",
                        "LessThan",
                        "GreaterThan",
                        "NotIn",
                        "In",
                        "Contains",
                        "All",
                        "Exists",
                        "Size"`

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -0,0 +1,583 @@
...
+type Target struct {
+    Type        *string             `json:"type,omitempty"`
+    Id          *string             `json:"id,omitempty"`
+}
+
+var (
+	target = &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"type": {
+				Description: "Type of the target the trigger is configured to hit",
```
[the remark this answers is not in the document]
Tyler-Bath: its open as a possibility, but no plans yet

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -234,7 +249,9 @@ func readProcessAutomationTrigger(ctx context.Context, d *schema.ResourceData, m
         }
 
         log.Printf("Read process automation trigger %s %s", d.Id(), *trigger.Name)
-        return nil
+
+        cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, resourceProcessAutomationTrigger())
```
ronanwatkins, line 253: This needs to be created up before you call d.Set on any attributes

genesyscloud/resource_genesyscloud_processautomation_trigger.go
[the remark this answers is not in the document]
ronanwatkins: Grand, it's best to do a validation.StringInSlice for `Workflow` and add more to the slice when more enums are added

ronanwatkins (approved):

PR #244, opened 2022-04-15
[DEVENGAGE-1104] Adding error handling for <nil> to map[string]interface{} type mismatch

charliecon:
This crash occurs when a resource field of type map, which has no required attributes, is defined and left empty. E.g. `genesyscloud_user.addresses`


ronanwatkins (approved):
Nice. Are there no other risky instances of accessing potentially nil maps?

charliecon:
> Nice. Are there no other risky instances of accessing potentially nil maps?

charliecon:
> Nice. Are there no other risky instances of accessing potentially nil maps?

@ronanwatkins No, I went through all of them. Most have at least one required field, so in that case the user will see a "Missing required field" error straight away. And in some cases the code was already handling it.

ronanwatkins:
> Details

Great. Will merge so

PR #248, opened 2022-04-20
Adding substitutions map for genesyscloud_flow resource

ronanwatkins:


carnellj (approved):
Looked through the code.  Downloaded the branch and am actually using it.  Works like a champ.  Thanks for the quick turnaround on this.

PR #250, opened 2022-04-20
Feat/intindy 2216 create trigger terraform provider

Tyler-Bath:
Updating the handling of the event_ttl_seconds field to allow for it to be optional. neeeded to update to string so that terraform UI did not show a bad response that would confuse users on gets when event ttl was not included.
Adding validation for event_ttl_seconds.


Tyler-Bath:
The schema.ResourceData that we use to return the values does not allow us to remove a field from it even when the field is null from the backed API. This resulted in us showing bad information when running the terraform apply from cli. It would show that event ttl was set to 0, even though it really was null in the backend. In order to avoid the confusion for the time being we decided to use a string to represent what is in the backend, and an empty string means that it is not set.

It seems that if there was a remove on the schema.ResourceData then before we returned this result if it does not exist in backend, we could remove it. 

From documentation of https://github.com/hashicorp/terraform-plugin-sdk:
`In the meantime, some providers have found success in using strings instead of numbers, which is a confusing user experience, but unlocks the ability to use "" to detect “unset” and "0" for 0. This is far from ideal, but again, is another option that can unblock you immediately.`

genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -156,15 +157,38 @@ func resourceProcessAutomationTrigger() *schema.Resource {
...
 			},
 			"event_ttl_seconds": {
-				Description: "How old an event can be to fire the trigger",
-				Type:        schema.TypeInt,
-				Optional:    true,
-				Required:    false,
+				Description:  "How old an event can be to fire the trigger. Must be an number greater than or equal to 10",
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validateEventTTLGreaterThanOrEqualTo10(),
```
carnellj, line 163: +1

carnellj (approved):

genesyscloud/resource_genesyscloud_processautomation_trigger.go
ronanwatkins: Looks like you can avoid it being set to 0 by updating the code in the read function to
```
		if trigger.EventTTLSeconds != nil && *trigger.EventTTLSeconds > 0 {
			d.Set("event_ttl_seconds", *trigger.EventTTLSeconds)
		}
```
and omitting the else. I hadn't known that setting an attribute to nil sets it to 0 if it's an int. We might have to set if that's causing issues in other resources.

Tyler-Bath: @ronanwatkins doing that then gives a response that looks like this in the CLI.

`the_trigger = {
  "enabled" = true
  "event_ttl_seconds" = tonumber(null)
  "id" = "a0e6766d-bc63-4ea0-a7ef-aa14bd50e054"
  "match_criteria" = toset([
    {
      "json_path" = "mediaType"
      "operator" = "In"
      "value" = ""
      "values" = tolist([
        "CHAT",
      ])
    },
  ])
  "name" = "Test Trigger - terraform creation - abcd-1234-values"
  "target" = toset([
    {
      "id" = "ae1e0cde-875d-4d13-a498-615e7a9fe956"
      "type" = "Workflow"
    },
  ])
  "topic_name" = "v2.detail.events.conversation.{id}.customer.end"
}
`

'tonumber(null)' just seems like nonsense, and i think i would prefer the empty string approach we have implemented

ronanwatkins: What are you doing to get this output from the CLI? I've never seen the CLI output any HCL like this

Tyler-Bath: I have this in my versions.tf:
```output "the_trigger" {
  description = "the test trigger"
  value       = resource.genesyscloud_processautomation_trigger.test-trigger
}```

Comment (ronanwatkins) [reply genesyscloud/resource_genesyscloud_processautomation_trigger.go:163]: Really not sure how it prints that stuff.
This is my config:
```
resource "genesyscloud_processautomation_trigger" "test-trigger" {
  name       = "Test Trigger"
  topic_name = "v2.detail.events.conversation.{id}.customer.end"
  enabled    = true
  target {
    id   = genesyscloud_flow.workflow.id
    type = "Workflow"
  }
  match_criteria {
    json_path = "mediaType"
    operator  = "Equal"
    value     = "CHAT"
  }
}
```
And I got this output after updating it to remove `event_ttl_seconds`:
```
......
Terraform will perform the following actions:

  # genesyscloud_processautomation_trigger.test-trigger will be updated in-place
  ~ resource "genesyscloud_processautomation_trigger" "test-trigger" {
      - event_ttl_seconds = 60 -> null
        id                = "dff89de7-ccac-4814-85f9-080557fd79bb"
        name              = "Test Trigger"
        # (2 unchanged attributes hidden)


      + target {
          + id   = "bcceac36-ffd3-4843-a3f9-5a867fa0f657"
          + type = "Workflow"
        }
      - target {
          - id   = "fb79b041-c332-458d-87c9-62e9913d7d1c" -> null
          - type = "Workflow" -> null
        }
        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
genesyscloud_processautomation_trigger.test-trigger: Modifying... [id=dff89de7-ccac-4814-85f9-080557fd79bb]
genesyscloud_processautomation_trigger.test-trigger: Modifications complete after 2s [id=dff89de7-ccac-4814-85f9-080557fd79bb]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

Tyler-Bath: if i remove my output section then i get the same thing as you. i was just concerned that if someone using this provider adds the terraform output section, then they will get strange looking responses for event ttl

ronanwatkins: Oh yeah, you used outputs. I wouldn't be too worried about what they see there.

Tyler-Bath:
Everything should be good to go for this now. I also fixed a bug i ran into with attempting to update the topic name. Our API does not allow updates to this field so i failed throw an error when someone tries to update it.

PR #261, opened 2022-05-18
adding support for access token

michaeljroddy:


genesyscloud/provider.go
```diff
@@ -34,6 +34,12 @@ func New(version string) func() *schema.Provider {
 	return func() *schema.Provider {
 		return &schema.Provider{
 			Schema: map[string]*schema.Schema{
+				"oauthaccess_token": {
```
ronanwatkins, line 37: `access_token` would be alright for the name.

ronanwatkins:
I think you'll also have to edit docs/index.md to add info about this also. Also, there's a section in the README that this could be added to.

genesyscloud/provider.go
```diff
@@ -34,6 +34,12 @@ func New(version string) func() *schema.Provider {
...
 			Schema: map[string]*schema.Schema{
+				"access_token": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					DefaultFunc: schema.EnvDefaultFunc("GENESYSCLOUD_ACCESS_TOKEN", nil),
+					Description: "A string that the OAuth client uses to make requests. Can be set with the `GENESYSCLOUD_ACCESS_TOKEN` environment variable.",
+				},
 				"oauthclient_id": {
 					Type:        schema.TypeString,
 					Required:    true,
```
ronanwatkins, line 45: I just spotted that `oauthclient_id`, `oauthclient_secret` and `aws_region` should be made optional now also. Index.md can be updated to reflect that then.

michaeljroddy: aws_region too?

PR #276, opened 2022-06-02
DEVENGAGE-1176: Upgraded to the latest version of the go sdk

carnellj:
Hey Ronan,

Here is my PR upgrading to v72 of the GO Sdk.  As you can see a number of things changed including a couple of key APIs.  Can you tell me if this looks good?  If this does not pass the smell test(like something changed with the new GO SDK), please reject the PR and open up a ticket with Michael.

Thanks,
    John


ronanwatkins:
Yeah, that's fine. I see one change on parameter ordering but that's an acceptable change with the openapi-generator.

PR #291, opened 2022-06-23
BLD-56635: add journey segment resource

gweszelovszki:


genesyscloud/resource_genesyscloud_journey_segment.go
```diff
@@ -0,0 +1,623 @@
...
+	"fmt"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/mypurecloud/platform-client-sdk-go/v72/platformclientv2"
```
ronanwatkins, line 14: I've just updated to v74 of the SDK, can you rebase and update this please.

genesyscloud/resource_genesyscloud_journey_segment.go
```diff
@@ -0,0 +1,623 @@
...
+				Type:        schema.TypeSet,
+				Required:    true,
+				Elem:        journeyPatternResource,
+			},
+		},
+	}
+
+	externalSegmentResource = &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"id": {
```
ronanwatkins, line 109: What is this the ID of? If it's another resource that we haven't already created in the provider then we'll have to create it also

genesyscloud/resource_genesyscloud_journey_segment.go
```diff
@@ -0,0 +1,623 @@
...
+	return readJourneySegment(ctx, d, meta)
+}
+
+func deleteJourneySegment(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	displayName := d.Get("display_name").(string)
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	journeyApi := platformclientv2.NewJourneyApiWithConfig(sdkConfig)
+
+	log.Printf("Deleting jounrey segment with display name %s", displayName)
```
ronanwatkins, line 350: jounrey -> journey

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func TestAccResourceJourneySegmentSession(t *testing.T) {
+	runTestCase(t, "basic_session_attributes")
+}
+
+func TestAccResourceJourneySegmentCustomer(t *testing.T) {
+	runTestCase(t, "basic_customer_attributes")
+}
+
+func TestAccResourceJourneySegmentContextOnly(t *testing.T) {
```
ronanwatkins, line 23: This fails for me with `API Error: 400 - Event type was not found with name: OtherEventName`

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/mypurecloud/platform-client-sdk-go/v72/platformclientv2"
+)
+
+func TestAccResourceJourneySegmentSession(t *testing.T) {
+	runTestCase(t, "basic_session_attributes")
```
ronanwatkins, line 16: This fails for me with `API Error: 400 - Event type was not found with name: EventName`

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/mypurecloud/platform-client-sdk-go/v72/platformclientv2"
+)
+
+func TestAccResourceJourneySegmentSession(t *testing.T) {
```
ronanwatkins, line 15: These tests only seem to be creating the resource without updating any attributes or using the `Check` field of the test config to check if the correct attributes are set.
They'll need to be updated to be more comprehensive. That's why we create the builder functions for the test resources instead of writing the config into a file.

genesyscloud/util_basesetting_properties.go
```diff
@@ -24,6 +24,33 @@ func buildBaseSettingsProperties(d *schema.ResourceData) *map[string]interface{}
 	return &returnValue
 }
 
+func setNullableValue[T any](d *schema.ResourceData, key string, value *T) {
```
ronanwatkins, line 27: Nice to see some generics in this project!
These could go in a new file though. This file was just created to add some helpers for a particular resource.

genesyscloud/resource_genesyscloud_journey_segment.go
```diff
@@ -0,0 +1,623 @@
...
+	})
+}
+
+func updateJourneySegment(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	journeyApi := platformclientv2.NewJourneyApiWithConfig(sdkConfig)
+	journeySegment := buildSdkPatchSegment(d)
+
+	log.Printf("Updating journey segment %s", d.Id())
+	if _, resp, err := journeyApi.PatchJourneySegment(d.Id(), *journeySegment); err != nil {
```
ronanwatkins, line 336: Have a search for `retryWhen(isVersionMismatch, func() (*platformclientv2.APIResponse, diag.Diagnostics)` in other resources. We wrap updates in that retry logic because sometimes there's a version mismatch when updating resources

ronanwatkins:
We also add data sources for each resource we create. They're pretty simple and fetch the resource by name.

genesyscloud/resource_genesyscloud_journey_segment.go
ronanwatkins: Let me know if you need to write more resources. I'm working on a code generator that will create a lot of the boiler plate for you.

genesyscloud/resource_genesyscloud_journey_segment.go
istvan-bihari-genesys: Main merged, updating new code too.

genesyscloud/resource_genesyscloud_journey_segment.go
istvan-bihari-genesys: Thanks, fixing.

genesyscloud/resource_genesyscloud_journey_segment_test.go
ronanwatkins: I'm running against a prod org in us-east-1

istvan-bihari-genesys: Interesting, why does it not fail at DCA?

Based on the API documentation it is not a required parameter and expected to accept any value.
```
journey
(Journey) The pattern of rules defining the segment.
patterns
(JourneyPattern[]) A list of zero or more patterns to match.
criteria
(Criteria[]) A list of one or more criteria to satisfy.
count
(integer) The number of times the pattern must match.
streamType
(string required) The stream type for which this pattern can be matched on.Valid values: Web, Custom, Conversation.
sessionType
(string required) The session type for which this pattern can be matched on.
eventName
(string) The name of the event for which this pattern can be matched on.
```
We have no extra information for this field and UI also does not make to set it. Maybe we should just remove from our tes cases.

istvan-bihari-genesys: I removed the event_name from test configs. Based on the code, if I provide one  than it should exist. Probably we had luck in DCA...
https://bitbucket.org/inindca/journey-rules-engine/src/master/src/main/java/com/genesys/journey/rulesengine/service/PatternValidator.java?at=master#lines-122

genesyscloud/resource_genesyscloud_journey_segment_test.go
ronanwatkins: Getting `API Error: 400 - Error validating JourneyPatterns. `attributes.bleki.value` is not a valid attribute for eventType: All Event Types (f3a0203c-a0ca-47e7-9c0d-63006b90b979)` now

genesyscloud/resource_genesyscloud_journey_segment_test.go
ronanwatkins: Getting `API Error: 400 - Error validating JourneyPatterns. `attributes.bleki.value` is not a valid attribute for eventType: All Event Types (01ed790c-db26-452c-9834-effb75599daf)`

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/mypurecloud/platform-client-sdk-go/v74/platformclientv2"
+)
+
+func TestAccResourceJourneySegmentSession(t *testing.T) {
+	runResourceJourneySegmentTestCase(t, "basic_session_attributes")
+}
+
+func TestAccResourceJourneySegmentCustomer(t *testing.T) {
```
ronanwatkins, line 19: The update function isn't being called during these tests

genesyscloud/resource_genesyscloud_journey_segment_test.go
istvan-bihari-genesys: Sad. :(
I do not understand why does it success on DCA and not in Live. Will try to get a Live account for testing.

genesyscloud/resource_genesyscloud_journey_segment_test.go
istvan-bihari-genesys: Why do you think that update isn't called during these tests? As I noted in other comment, in each test data folder we have multiple files where the first is the create step an than the following are the updates:

First test step is always the create step and all following steps are updates because the resource name is always the same. `generateTestSteps` generate test step array form these files for 1 test case, just as in already existing tests.
For example:
`test\data\resource\journey_segment\basic_customer_attributes\01_create_customer_scope.tf` has resource `"genesyscloud_journey_segment" "terraform_test_-TEST-CASE-" {`
This is the create step, than after that the update is:
`test\data\resource\journey_segment\basic_customer_attributes\02_external_segment_update_name.tf` with same resource name resource `"genesyscloud_journey_segment" "terraform_test_-TEST-CASE-" {`
and so on with `03_external_segment_force_new_creation.tf`

Here I have to note that meanwhile we developed this flow, we always debugged the `updateJourneySegment` and it was called at requested count depending on the update files count.

istvan-bihari-genesys: image
Just to show that update is executed.

ronanwatkins: It must be something to do with how you're creating the config. It looks like a new resource is created instead of updating it.

ronanwatkins (approved):

PR #303, opened 2022-07-08
[DEVENGAGE-1283] Outbound Attempt Limit resource & data action 

charliecon:


genesyscloud/resource_genesyscloud_outbound_attemptlimit.go
```diff
@@ -0,0 +1,334 @@
+package genesyscloud
```
ronanwatkins, line 1: You'll have to add the resource exporter stuff. I haven't added code generation for that yet

PR #307, opened 2022-07-11
adding response log hook

michaeljroddy:
updated the go sdk version to latest v75. added the response log hook.


ronanwatkins:
Take a look at the tests, looks like the GetRoutingQueues change has caused some issues

ronanwatkins (approved):

PR #312, opened 2022-07-18
updating trigger terraform to allow for the description field, also u…

Tyler-Bath:
updating how topic name changes on updates are handled, and adding in description field to process automation triggers.

We used to fail the updates to triggers when the topic name field was changed, now the terraform will explicitly say that a replacement is required, and then asks the user to confirm the changes. So the user will need to be aware that the ID of the trigger will change in this case:

Screen Shot 2022-07-18 at 3 17 31 PM

### DIFF
--- a/genesyscloud/resource_genesyscloud_processautomation_trigger.go
+++ b/genesyscloud/resource_genesyscloud_processautomation_trigger.go
@@ -27,15 +27,18 @@ type ProcessAutomationTrigger struct {
 	Enabled         *bool            `json:"enabled,omitempty"`
 	EventTTLSeconds *int             `json:"eventTTLSeconds,omitempty"`
 	Version         *int             `json:"version,omitempty"`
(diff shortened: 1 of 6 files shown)


genesyscloud/resource_genesyscloud_processautomation_trigger.go
```diff
@@ -285,17 +303,14 @@ func updateProcessAutomationTrigger(ctx context.Context, d *schema.ResourceData,
 			return resp, diag.Errorf("Failed to read process automation trigger %s: %s", d.Id(), getErr)
 		}
 
-		//make sure that topic name is not updated
-		if topic_name != *trigger.TopicName {
```
ronanwatkins, line 289: I totally missed this in the initial review. Yeah we should never do this. Always let `ForceNew` take care of it

ronanwatkins:
The tests don't work for me. Looks like the description isn't being set

PR #313, opened 2022-07-20
DEVENGAGE-1299: Changed the sanitize name code to produce a stable re…

carnellj-genesys:
…source name

### DIFF
--- a/genesyscloud/resource_exporter.go
+++ b/genesyscloud/resource_exporter.go
@@ -2,10 +2,10 @@ package genesyscloud
 
 import (
 	"context"
+	"crypto/sha1"
+	"encoding/hex"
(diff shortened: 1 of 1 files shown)


carnellj-genesys:
Turns out we are using a hash that generates a stable number.  Originally Jeremy and I believed we were not hashing a stable number because he thought we were generating a unique number each time.  Once I deep dived on the code, I realized:

1.  The code appends a hash based on an input string rather than a randomly generated number.  The generated hash is stable and not likely to conflict with other resource names.  
2.  The resource name is only mangled if we have spaces in the resource name.  I am going to close the ticket in DEVENGAGE-1299 to reflect this.

PR #314, opened 2022-07-21
DEVENGAGE-1307:  Fixed the from_email to make it required.

carnellj-genesys:


genesyscloud/resource_genesyscloud_routing_email_route.go
```diff
@@ -124,7 +124,7 @@ func resourceRoutingEmailRoute() *schema.Resource {
 			"from_email": {
 				Description: "The sender email to use for outgoing replies.",
 				Type:        schema.TypeString,
-				Optional:    true,
+				Required:    true,
```
carnellj-genesys, line 127: The problem here is that the API does not mark the from_email as required.  It actually is required as the underlying API will complain that a value must be between 1 and 256.

ronanwatkins: Typical.

ronanwatkins:
You'll need to run go generate and push the changes

PR #323, opened 2022-08-03
Feat/devengage 1298

dginty4:


genesyscloud/resource_genesyscloud_routing_settings.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v75/platformclientv2"
+	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/consistency_checker"
+	"log"
+)
+
+func resourceRoutingSettings() *schema.Resource {
+	return &schema.Resource{
+		Description: "An organizations routing settings",
```
ronanwatkins, line 16: organizations -> organization's

genesyscloud/resource_genesyscloud_routing_settings.go
```diff
@@ -0,0 +1,277 @@
...
+		CreateContext: createWithPooledClient(createRoutingSettings),
+		ReadContext:   readWithPooledClient(readRoutingSettings),
+		UpdateContext: updateWithPooledClient(updateRoutingSettings),
+		DeleteContext: deleteWithPooledClient(deleteRoutingSettings),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"reset_agent_on_presence_change": {
```
ronanwatkins, line 27: This may need to be computed. Try creating a test without providing this attribute. It would probably error out when it doesn't expect this to have a value

ronanwatkins:
Documentation sync check is failing so you'll need to update that and check it in

genesyscloud/data_source_genesyscloud_routing_settings_test.go
```diff
@@ -0,0 +1,58 @@
...
+					trueValue,
+					generateSettingsTranscription(transcription, confidence, trueValue, falseValue),
+				) + generateRoutingSettingsDataSource(
+					settingsDataSource,
+					trueValue,
+					"genesyscloud_routing_settings."+settingsResource,
+					generateSettingsTranscription(transcription, confidence, trueValue, falseValue),
+				),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("data.genesyscloud_routing_settings."+settingsDataSource, "transcription.0.transcription", "genesyscloud_routing_settings."+settingsResource, "transcription.0.transcription"),
```
ronanwatkins, line 35: Could add a few more checks here for the reset_agent_on_presence_change value and contactcenter values

genesyscloud/resource_genesyscloud_routing_settings.go
```diff
@@ -0,0 +1,297 @@
...
+			if err != nil {
+				return diag.Errorf("Failed to update Contact center for routing setting %s: %s\n", d.Id(), err)
+			}
+		}
+	}
+	return nil
+}
+
+func updateTranscription(d *schema.ResourceData, routingAPI *platformclientv2.RoutingApi) diag.Diagnostics {
+	var transcription string
```
ronanwatkins, line 262: Just spotted this. Usually, we want to give variables as little scope as possible so the best thing would be to declare and initialize these only where they are used. So on line 272 it would be `transcription := transcriptionMap["transcription"].(string)` for example.

ronanwatkins (approved):

ronanwatkins (approved):

PR #324, opened 2022-08-09
[DEVENGAGE-1285] Outbound Call Analysis Response Sets

charliecon:


genesyscloud/resource_genesyscloud_outbound_callanalysisresponseset.go
```diff
@@ -0,0 +1,386 @@
...
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			`name`: {
+				Description: `The name of the Response Set.`,
+				Required:    true,
+				Type:        schema.TypeString,
+			},
+			`responses`: {
+				Description: `Map of disposition identifiers to reactions. Required if beep_detection_enabled = true.`,
```
ronanwatkins, line 165: Description should be updated since it's not a map

ronanwatkins: Maybe it should be `List of maps of disposition ...`

charliecon: @ronanwatkins Done ✅

ronanwatkins (approved):

PR #345, opened 2022-09-08
DEVENGAGE-1049: Knowledge base resources

diemenschmaschine92:
Adds knowledge base, documents, and categories


genesyscloud/data_source_genesyscloud_knowledge_knowledgebase.go
```diff
@@ -0,0 +1,64 @@
...
+	return &schema.Resource{
+		Description: "Data source for Genesys Cloud Knowledge Base. Select a knowledge base by name.",
+		ReadContext: readWithPooledClient(dataSourceKnowledgeKnowledgebaseRead),
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "Knowledge base name",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"core_language": {
```
ronanwatkins, line 25: Usually we just use the name to search. Is there a reason for also using the core_language?

genesyscloud/resource_genesyscloud_knowledge_category.go
```diff
@@ -0,0 +1,281 @@
...
+
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	knowledgeAPI := platformclientv2.NewKnowledgeApiWithConfig(sdkConfig)
+
+	log.Printf("Reading knowledge category %s", knowledgeCategoryId)
+	return withRetriesForRead(ctx, d, func() *resource.RetryError {
+		knowledgeCategory, resp, getErr := knowledgeAPI.GetKnowledgeKnowledgebaseLanguageCategory(knowledgeCategoryId, knowledgeBaseId, languageCode)
+		if getErr != nil {
+			if isStatus404(resp) {
+				log.Println("404")
```
ronanwatkins, line 163: Best to remove this log

examples/resources/genesyscloud_knowledge_category/resource.tf
```diff
@@ -0,0 +1,9 @@
+resource "genesyscloud_knowledge_category" "example_category" {
+  knowledge_base_id = genesyscloud_knowledge.example_knowledgebase.id
```
ronanwatkins, line 2: genesyscloud_knowledge -> genesyscloud_knowledge_knowledgebase

genesyscloud/resource_genesyscloud_knowledge_category.go
```diff
@@ -0,0 +1,281 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"language_code": {
+				Description:  "language code of the category",
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"en-US", "en-UK", "en-AU", "de-DE", "es-US", "es-ES", "fr-FR", "pt-BR", "nl-NL", "it-IT", "fr-CA"}, false),
+			},
+			"knowledge_category": {
```
ronanwatkins, line 114: I don't think the name and description of the knowledgeCategory element would have any effect on the resource. It's probably best to name this something like `parent_knowledge_category_id` and just have it as a string value containing an ID.
Will need to include it in the tests too to make sure it works

genesyscloud/resource_genesyscloud_knowledge_document.go
```diff
@@ -0,0 +1,477 @@
...
+		CreateContext: createWithPooledClient(createKnowledgeDocument),
+		ReadContext:   readWithPooledClient(readKnowledgeDocument),
+		UpdateContext: updateWithPooledClient(updateKnowledgeDocument),
+		DeleteContext: deleteWithPooledClient(deleteKnowledgeDocument),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"knowledge_base_id": {
```
ronanwatkins, line 157: Needs to be added to RefAttrs

genesyscloud/resource_genesyscloud_knowledge_category_test.go
```diff
@@ -0,0 +1,116 @@
...
+		categoryName               = "Terraform Knowledge Category" + uuid.NewString()
+		categoryDescription        = "test-description1"
+	)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				// Create
```
ronanwatkins, line 30: Need an update in the tests also

genesyscloud/resource_genesyscloud_knowledge_category_test.go
```diff
@@ -0,0 +1,116 @@
...
+				) +
+					generateKnowledgeCategory(
+						knowledgeCategoryResource1,
+						knowledgeBaseResource1,
+						knowledgeBaseCoreLanguage1,
+						categoryName,
+						categoryDescription,
+					),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("genesyscloud_knowledge_category."+knowledgeCategoryResource1, "knowledge_category.0.name", categoryName),
```
ronanwatkins, line 45: Could do with more attribute checks

genesyscloud/resource_genesyscloud_knowledge_document_test.go
```diff
@@ -0,0 +1,212 @@
...
+		contentLocationUrl           = "http://example.com"
+		articleAlternatives          = []string{"alt1, alt2"}
+	)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				// Create
```
ronanwatkins, line 40: Need an update also

genesyscloud/resource_genesyscloud_knowledge_knowledgebase_test.go
```diff
@@ -0,0 +1,83 @@
...
+		knowledgeBaseDescription1  = "test-knowledgebase-description1"
+		knowledgeBaseCoreLanguage1 = "en-US"
+	)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				// Create
```
ronanwatkins, line 26: Need an update also

genesyscloud/resource_genesyscloud_knowledge_document_test.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	// Success. All knowledge bases destroyed
+	return nil
+}
+
+func generateKnowledgeDocumentLifeCycle() string {
+	return `
+	lifecycle {
+		ignore_changes = [
+			"knowledge_document[0].categories",
```
ronanwatkins, line 208: Why are we ignoring this change?

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
```diff
@@ -0,0 +1,202 @@
...
+				Description:  "Core language for knowledge base in which initial content must be created, language codes [en-US, en-UK, en-AU, de-DE] are supported currently, however the new DX knowledge will support all these language codes",
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"en-US", "en-UK", "en-AU", "de-DE", "es-US", "es-ES", "fr-FR", "pt-BR", "nl-NL", "it-IT", "fr-CA"}, false),
+			},
+			"published": {
+				Description: "Flag that indicates the knowledge base is published",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
```
charliecon, line 81: I got a plan not empty error when running the tests for this resource and I think it's because this field should be computed

genesyscloud/resource_genesyscloud_knowledge_category.go
```diff
@@ -0,0 +1,281 @@
...
+	sdkConfig := meta.(*providerMeta).ClientConfig
+	knowledgeAPI := platformclientv2.NewKnowledgeApiWithConfig(sdkConfig)
+
+	knowledgeCategoryRequest := buildKnowledgeCategory(knowledgeCategory)
+
+	log.Printf("Creating knowledge category %s", knowledgeCategory["name"].(string))
+	knowledgeCategoryResponse, _, err := knowledgeAPI.PostKnowledgeKnowledgebaseLanguageCategories(knowledgeBaseId, languageCode, knowledgeCategoryRequest)
+	if err != nil {
+		fmt.Errorf("Failed to create knowledge category %s: %s", *knowledgeCategoryResponse.Name, err)
+		return diag.Errorf("Failed to create knowledge category %s: %s", *knowledgeCategoryResponse.Name, err)
```
charliecon, line 139: If the post isn't successful, I don't think `*knowledgeCategoryResponse.Name` will have any value in it, so it's better to use `knowledgeCategory["name"].(string)`. Or maybe just leave it out since name is an optional field so there mightn't be any value to begin with

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
diemenschmaschine92: That's odd, the test for knowledgebase passes for me.  I'm wondering if it still fails if you do a git pull from this branch?

genesyscloud/resource_genesyscloud_knowledge_document_test.go
diemenschmaschine92: I can remove this actually. I thought it was needed at some point but the test is passing without it now.

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
diemenschmaschine92: Actually another comment is saying this test needs an update step, if you're running a different test for the resource that's probably why.  Will update

charliecon: Actually it passed for me on this attempt. I noticed when I create a knowledge base for the org we use in this repo's workflow, `published` comes back as true in the response, but with hollywoo it comes back false. Not sure if that means it should be computed or not. @ronanwatkins any thoughts?

genesyscloud/resource_genesyscloud_knowledge_category.go
diemenschmaschine92: knowledgeCategory is actually defining the name and description of the category being created, not those of the parent category.

examples/resources/genesyscloud_knowledge_category/resource.tf
diemenschmaschine92: thanks for catching this

genesyscloud/data_source_genesyscloud_knowledge_knowledgebase.go
diemenschmaschine92: I included both because the knowledge base creation resource `POST
/api/v2/knowledge/knowledgebases` lists "name" as optional but "core_language" as required.  Which doesn't really make sense to me, but I was thinking this should guarantee uniqueness.

diemenschmaschine92: This doesn't guarantee uniqueness, though.
Two solutions I can think of:
1) make "name" required for the resource even though it's optional in the API
2) just not have a data source for this resource

1) makes most sense to me, any ideas?

genesyscloud/resource_genesyscloud_knowledge_category.go
```diff
@@ -0,0 +1,281 @@
...
+		CreateContext: createWithPooledClient(createKnowledgeCategory),
+		ReadContext:   readWithPooledClient(readKnowledgeCategory),
+		UpdateContext: updateWithPooledClient(updateKnowledgeCategory),
+		DeleteContext: deleteWithPooledClient(deleteKnowledgeCategory),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"knowledge_base_id": {
```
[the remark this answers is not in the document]
diemenschmaschine92: I suppose it is.  I wondered if there was difference between a straight id reference and an id reference that gets expanded by the SDK into a full object when it comes to RefAttrs? This one doesn't get expanded.

genesyscloud/data_source_genesyscloud_knowledge_knowledgebase.go
ronanwatkins: No that's fine. Good to know the reasoning behind it

genesyscloud/resource_genesyscloud_knowledge_category.go
ronanwatkins: Oh sorry I misread that

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
ronanwatkins: Yeah probably safe to put it as computed. There generally isn't any negative consequences of marking an attribute as computed.

> Actually another comment is saying this test needs an update step, that's probably why I'm not seeing the failure. Update steps are just good to have in the tests because it provides for a more comprehensive test

genesyscloud/resource_genesyscloud_knowledge_category.go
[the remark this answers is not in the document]
ronanwatkins: Don't think there's a difference. It's just so when someone does an export of the resource you get a logical reference instead of a raw ID

carnellj-genesys (approved):

genesyscloud/resource_genesyscloud_knowledge_category.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+	}
+
+	return resources, nil
+}
+
+func knowledgeCategoryExporter() *ResourceExporter {
+	return &ResourceExporter{
+		GetResourcesFunc: getAllWithPooledClient(getAllKnowledgeCategories),
+		RefAttrs:         map[string]*RefAttrSettings{}, // No references
```
ronanwatkins: This will need to have the `knowledge_base_id` reference

jds98929: The ref is added there, not sure why this isn't showing as Outdated.

diemenschmaschine92: This ref is added, not sure why it's showing as Outdated.

diemenschmaschine92: It looks like there's a permission missing for the acceptance tests
knowledge-base-403

There are also a couple tests for existing resources failing.  Going to look into the media retention policy failure

carnellj-genesys: Fix the permission issue

diemenschmaschine92: @carnellj-genesys Fixed. Also the data source test is fixed now.  It searches for published and unpublished results and takes the published one if possible.

PR #355, opened 2022-09-28
[DEVENGAGE-1242] Outbound Campaigns

charliecon:


carnellj-genesys (commented):
looks good Charlie.  I would remove the one //TODO comment and instead create a JIRA ticket as a place holder.  Also why on the delete do you check for a 400 and then retry.  How long will you retry on a 400 type error.

charliecon:
@carnellj-genesys TODO are removed. I'm not sure why that is in the delete function. It comes from the code generator

charliecon:
> looks good Charlie. I would remove the one //TODO comment and instead create a JIRA ticket as a place holder. Also why on the delete do you check for a 400 and then retry. How long will you retry on a 400 type error.

I assume the 400 is for this error code: `Could not delete the resource because it is referenced by another entity.`. I can't recreate this locally, but I could see it being in an issue in the pipeline tests.

carnellj-genesys (approved):

PR #358, opened 2022-10-04
Removing references to queue media settings videoComm and socialExpre…

charliecon:
…ssion


docs/resources/routing_queue.md
```diff
@@ -172,26 +170,6 @@ Required:
 - `service_level_percentage` (Number) The desired Service Level. A float value between 0 and 1.
```
carnellj-genesys, line 170: Is the documentation generated off the comments inside the code.  Wouldnt you need to make the modifications there?

carnellj-genesys (approved):

PR #382, opened 2022-11-08
DEVENGAGE-1241 Outbound settings

dginty4:


genesyscloud/resource_genesyscloud_outbound_settings_test.go
```diff
@@ -0,0 +1,201 @@
...
+			},
+			{
+				// Import/Read
+				ResourceName:            "genesyscloud_outbound_settings." + resourceId,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"max_calls_per_agent", "max_line_utilization", "abandon_seconds", "compliance_abandon_rate_denominator", "automatic_time_zone_mapping"},
+			},
+		},
+	})
```
carnellj-genesys, line 122: Dont you need a test to make sure the resource is deleted?

genesyscloud/validators.go
```diff
@@ -57,6 +58,20 @@ func validateTime(time interface{}, _ cty.Path) diag.Diagnostics {
 	return diag.Errorf("Time %v is not a valid time", time)
 }
 
+// Validates a date string is in format hh:mm
```
carnellj-genesys, line 61: Nice

carnellj-genesys (changes requested):

genesyscloud/data_source_genesyscloud_outbound_settings.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v80/platformclientv2"
+)
+
+func dataSourceOutboundSettings() *schema.Resource {
+	return &schema.Resource{
+		Description:   "An organization's outbound settings",
+		ReadContext:   readWithPooledClient(dataSourceOutboundSettingsRead),
+		SchemaVersion: 1,
+		Schema:        resourceOutboundSettings().Schema,
```
charliecon, line 15: I'm not sure if there is any need for this data source if the resource is for defining default behaviours. @carnellj-genesys thoughts? 
Also I'm getting a panic error when I try to create this data source with no fields set, similar to how we do with the data source `genesyscloud_auth_division_home`

genesyscloud/resource_genesyscloud_outbound_settings_test.go
dginty4: Outbound settings only have two endpoints a get and a patch so I don't think I can delete the resource but I could be wrong?

genesyscloud/data_source_genesyscloud_outbound_settings.go
dginty4: When I was creating the routing settings, which I think is very similar to the outbound settings, Ronan thought I should create a data source but I can remove it if we don't think it's necessary

carnellj-genesys: I am ok with leaving the data source out.

PR #386, opened 2022-11-10
[DEVENGAGE-1487] Skill groups resource member_division_ids

charliecon:


genesyscloud/resource_genesyscloud_routing_skill_group.go
```diff
@@ -201,13 +208,123 @@ func createOrUpdateSkillGroups(ctx context.Context, d *schema.ResourceData, meta
 		log.Printf("Updated skill group %s", skillGroupsRequest.Name)
 	}
 
+	// Update member division IDs
+	memberDivisionIds := d.Get("member_division_ids").([]interface{})
```
carnellj-genesys, line 212: This overall function is getting pretty big.  Do we want to move the GroupMember logic to their own set of functions.

genesyscloud/resource_genesyscloud_routing_skill_group.go
```diff
@@ -201,13 +208,123 @@ func createOrUpdateSkillGroups(ctx context.Context, d *schema.ResourceData, meta
...
+
+	return allIds, nil
+}
+
+func createListsForSkillgroupsMembersDivisionsPost(schemaSkillGroupMemberDivisionIds []string, apiSkillGroupMemberDivisionIds []string,
+	create bool, divisionId string, meta interface{}) ([]string, []string, diag.Diagnostics) {
+	toAdd := make([]string, 0)
+	toRemove := make([]string, 0)
+	if stringInSlice("*", schemaSkillGroupMemberDivisionIds) {
+		if len(schemaSkillGroupMemberDivisionIds) > 1 {
```
carnellj-genesys, line 284: Should this code block be extracted into its own methods.  Also idiomatic go tries to avoid nested if..then functions

genesyscloud/resource_genesyscloud_routing_skill_group.go
```diff
@@ -309,6 +427,58 @@ func readSkillGroups(ctx context.Context, d *schema.ResourceData, meta interface
 			d.Set("skill_conditions", nil)
 		}
 
+		log.Printf("Reading skills group %s member divisions", d.Id())
```
carnellj-genesys, line 430: Again this function is getting big.  The 'readMemberDivision' logic is cohesive and should be extracted into its own functions

carnellj-genesys (changes requested):

carnellj-genesys (approved):

PR #403, opened 2022-12-08
DEVENGAGE-1534: Removing the ForceNew attribute on the data table.

carnellj-genesys:


genesyscloud/resource_genesyscloud_architect_datatable.go
```diff
@@ -27,7 +27,6 @@ var (
 				Description: "Name of the property.",
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
```
carnellj-genesys, line 30: This is the big change.  The issue here is that I am removing the ForceNew field.  I think what happened in Ronan added this for the situations where a user redefines the Schema and wants to deploy.  The API will only allow additive-type behavior on the schema with data.  If you try to update a property or delete a property from the schema definition the API will return an error.  I think Ronan added the ForceNew to work around this, but didn't realize that this is one of the few artifacts directly referenced in an Architect flow

carnellj-genesys (commented):
I went through each of the attributes in resources with a ForceNew and explicitly call out that if they change the object will be dropped and recreated.

charliecon (approved):

PR #408, opened 2022-12-19
Feat/devengage 1500 - routing sms addresses

dginty4:


charliecon:
@dginty4 I noticed as well the resource test isn't passing in the pipeline - https://github.com/MyPureCloud/terraform-provider-genesyscloud/actions/runs/3732667087/jobs/6332443150 
There must be some difference between hollywoo and this org that's not allowing them to pass

carnellj-genesys (approved):

PR #412, opened 2023-01-05
DEVENGAGE-1562 Updating contact list resource

dginty4:


genesyscloud/resource_genesyscloud_outbound_contactlist.go
```diff
@@ -100,12 +120,19 @@ func resourceOutboundContactList() *schema.Resource {
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
 			`phone_columns`: {
-				Description: `Indicates which columns are phone numbers. Changing the phone_columns attribute will cause the outboundcontact_list object to be dropped and recreated with a new ID`,
+				Description: `Indicates which columns are phone numbers. Changing the phone_columns attribute will cause the outboundcontact_list object to be dropped and recreated with a new ID. Required if email_columns is empty`,
 				Optional:    true,
 				ForceNew:    true,
```
carnellj-genesys, line 125: Is there a reason why we need a forcenew here.

dginty4: I'm not sure, it was like that when I started. @charliecon do you know why it's marked ForceNew? I can remove it if it's not necessary.

charliecon: Yes, I tried to update them there via the API explorer and got this back:
```
  "message": "An attempt was made to update a ContactList in an invalid way. Details: Phone number columns are different",
  "code": "cannot.update.phone.columns",
  "status": 400,
```
Same for `column_names`

carnellj-genesys: Ok.  I am good with that then

charliecon (approved):

PR #415, opened 2023-01-13
[feat/DEVENGAGE-1444] Response Management Response Assets

charliecon:


dginty4 (commented):
Why is there no update method? It looks like there is a PUT endpoint available.

dginty4 (approved):

PR #417, opened 2023-01-13
DEVENGAGE-1580: Removed the force new flag off of schedule.

carnellj-genesys:
I can't figure out why ForceNew was added here.  The put allows you to change the new without any problems.  The problem with schedules is that schedules can be referenced in an Architect flow.  If you change the name and force a new object to be created/delete, it breaks the dependency on architect flow.  I tried to rerun the test, but my OAuth client seems to be missing a permission that I cant figure out;

### DIFF
--- a/genesyscloud/resource_genesyscloud_architect_schedules.go
+++ b/genesyscloud/resource_genesyscloud_architect_schedules.go
@@ -62,10 +62,9 @@ func resourceArchitectSchedules() *schema.Resource {
 		SchemaVersion: 1,
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Description: "Name of the schedule. Note: If the name is changed, this will cause the schedule object to be dropped and recreated with a new ID.  This can cause an Architect Flow to become invalid.",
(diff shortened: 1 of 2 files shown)


dginty4 (approved):

genesyscloud/resource_genesyscloud_architect_schedules.go
```diff
@@ -62,10 +62,9 @@ func resourceArchitectSchedules() *schema.Resource {
 		SchemaVersion: 1,
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Description: "Name of the schedule. Note: If the name is changed, this will cause the schedule object to be dropped and recreated with a new ID.  This can cause an Architect Flow to become invalid.",
+				Description: "Name of the schedule.",
```
charliecon, line 65: `go generate`

carnellj-genesys: Fixed

charliecon (approved):

PR #418, opened 2023-01-13
Maes 583 update to include delaybyseconds

Tyler-Bath:
Update for process automations recently added delayBy field, Not yet enabled in the prods.


examples/resources/genesyscloud_processautomation_trigger/resource.tf
```diff
@@ -12,5 +12,6 @@ resource "genesyscloud_processautomation_trigger" "example-trigger" {
     value     = "CHAT"
   }
   event_ttl_seconds = 60
+  delay_by_seconds  = 30
```
charliecon, line 15: Having both `delay_by_seconds` and `event_ttl_seconds` in the example could send the wrong message. Ideally the example should be something that will work first time if a user decides to copy and paste it into their config.

charliecon: Great. Many thanks for the contribution! I'll add some other team members as reviewers as well for them to take a look.

charliecon (approved):

PR #430, opened 2023-02-01
DEVENGAGE-1445 Responsemanagement responses

dginty4:


genesyscloud/resource_genesyscloud_responsemanagement_response_test.go
```diff
@@ -0,0 +1,310 @@
...
+) string {
+	if interactionType != "" {
+		interactionType = fmt.Sprintf(`interaction_type = "%s"`, interactionType)
+	}
+	if schema != "" {
+		schema = fmt.Sprintf(`substitutions_schema_id = %s`, schema)
+	}
+	if responseType != "" {
+		responseType = fmt.Sprintf(`response_type = "%s"`, responseType)
+	}
```
charliecon, line 229: I found a better approach to this is leaving `response_type = %s` in the return string ,and then either passing `nullValue` or `strconv.Quote("<the value>")` into the function. It keeps things a lot more readable and manageable. (Same for the other fields ofc)

dginty4: You right Charlie, it doesn't look better, thanks

genesyscloud/data_source_genesyscloud_responsemanagement_response_test.go
```diff
@@ -0,0 +1,73 @@
...
+				Config: generateResponseManagementLibraryResource(
+					libraryResource,
+					libraryName,
+				) + generateResponseManagementResponseResource(
+					responseResource,
+					name,
+					[]string{"genesyscloud_responsemanagement_library." + libraryResource + ".id"},
+					"",
+					"",
+					"",
```
charliecon, line 38: These need to change to nullValue as well

charliecon (approved):

PR #431, opened 2023-02-07
[DEVENGAGE-983] Exporting user prompt audio

charliecon:


genesyscloud/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -714,6 +722,10 @@ func architectUserPromptExporter() *ResourceExporter {
 	return &ResourceExporter{
 		GetResourcesFunc: getAllWithPooledClient(getAllUserPrompts),
 		RefAttrs:         map[string]*RefAttrSettings{}, // No references
+		CustomFileWriter: CustomFileWriterSettings{
+			RetrieveAndWriteFilesFunc: ArchitectPromptAudioResolver,
+			SubDirectory:              "audio",
```
carnellj-genesys, line 727: Do we want to call this audio?  Audio is a little vague.  How about something like prompts or even audio_prompts to make it a little more specific.  

Will we build a similar model when we get to scripts and architect flows.   I think this is really cool.

Thanks,
    John

carnellj-genesys (approved):

genesyscloud/resource_genesyscloud_architect_user_prompt.go
charliecon: Done ✅

PR #443, opened 2023-02-16
DEVENGAGE-1638: Fixed bug in audio prompt custom export function

charliecon:


carnellj-genesys (commented):
Hey Charlie,

We didnt catch this with our tests.  Did we add a test now to make sure the filename and the file reference are the same

carnellj-genesys (approved):

PR #447, opened 2023-02-17
changed proficiency to integer

abejaranoh:
since Genesys Cloud only accepts integers, this must be changed

### DIFF
--- a/examples/resources/genesyscloud_user/resource.tf
+++ b/examples/resources/genesyscloud_user/resource.tf
@@ -23,7 +23,7 @@ resource "genesyscloud_user" "example_user" {
   }
   routing_skills {
     skill_id    = genesyscloud_routing_skill.test-skill.id
-    proficiency = 4.5
+    proficiency = 4
   }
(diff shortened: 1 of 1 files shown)


charliecon:
> since Genesys Cloud only accepts integers, this must be changed

Hi Alexander

After toying around with the resource, I noticed that float values work when the value is a factor of 0.5 , so 4.5 will work, but not 4.2
Not sure yet why that is, but I'll look into it and some add extra validation checking and context to the docs. 

As a sidenote; if you make some changes to the docs in the future, such as to an example or to a field description, be sure to run `go generate` from the terminal. This will generate the changes in `/docs`. 

Thanks for the heads up!
Charlie

abejaranoh:
Hallo Charlie,

would it then make sense t change the GUI instead? I doubled checked and there is no way to give "half a star" to a skill

charliecon:
@abejaranoh I agree it is odd behaviour. I can ask the team in charge about that.

PR #458, opened 2023-03-02
DEVENGAGE-1583 - Process automation triggers update to go sdk

dginty4:


dginty4:
@charliecon just in case you're confused whats happening here, I was updating the process automation triggers resource to use the Go SDK but we've realised it's going to be very difficult to update due to one of attributes being listed as a JsonNode which is basically just a placeholder value in the swagger docs. the actually type of the attribute is inherited depending on the values of other fields but this can't really be translated to Go.

charliecon (approved):

PR #460, opened 2023-03-06
[Do not merge]DEVENGAGE-1672: Fixed the web_deploy datasource test

carnellj-genesys:


genesyscloud/data_source_genesyscloud_webdeployments_deployment_test.go
```diff
@@ -31,10 +29,26 @@ func TestAccDataSourceWebDeploymentsDeployment(t *testing.T) {
...
+		allow_all_domains = true
+		configuration {
+			id = "${genesyscloud_webdeployments_configuration.minimal.id}"
+			version = "${genesyscloud_webdeployments_configuration.minimal.version}"
+		}
+	}
+	
 	data "genesyscloud_webdeployments_deployment" "basic-data" {
-		name = %s
+		depends_on=[genesyscloud_webdeployments_deployment.basic]
```
carnellj-genesys, line 50: This is the fix. We were creating the resources and the data source at the same time.  The data source executes before the resources are made because there is no implicit dependency.  So by declaring an explicit dependency we ensure that the data source does not get executed until all the resources are complete.

PR #474, opened 2023-03-14
DEVENGAGE-1685: Fixing issue with missing package for go generate

charliecon:


charliecon:
@carnellj-genesys Not the ideal solution but it highlights what the issue is. At the moment, running `go mod tidy` removes the indirect reference to `github.com/mypurecloud/terraform-provider-genesyscloud v1.11.1`, which causes the command `go generate` to fail with error message `no required module provides package github.com/mypurecloud/terraform-provider-genesyscloud/apidocs;`
It seems this is coming from the import statement `import provider "terraform-provider-genesyscloud/genesyscloud"` not functioning the same way as what was there previously; `import "github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud"` 

I tried importing `terraform-provider-genesyscloud/apidocs` in different ways but it is not doable, as this is a program and not an importable package.

For now, the fix I found is to add `import _ "github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud"` to main.go just so that `go mod tidy` does not remove `github.com/mypurecloud/terraform-provider-genesyscloud v1.11.1` from go.mod

main.go
```diff
@@ -7,6 +7,8 @@ import (
 
 	_ "terraform-provider-genesyscloud/genesyscloud/tfexporter"
 
+	_ "github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud"
```
carnellj-genesys, line 10: I would add a comment to not remove this line and the reason why.

carnellj-genesys (approved):

main.go
charliecon: Will do. 

I should mention as well, I just tried changing the package name in `apidocs/main.go` from `main` to `apidocs` so that it could be imported into `{root}/main.go` and `{root}/genesyscloud/provider.go` but no luck with that approach either.

PR #475, opened 2023-03-15
Add functionality to enable a campaign from creation

bbbco:
We need to be able to enable a campaign from the onset of creation. Considering the Genesys API requires the campaign to be created in an "off" state, we can create the campaign initially off, and then update the campaign status to on immediately afterwards to simulate building an enabled campaign from the onset.


bbbco:
Closing out in deference to https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/476

PR #476, opened 2023-03-16
Feat/enable campaign after creation

bbbco:
We need to be able to enable a campaign from the onset of creation. Considering the Genesys API requires the campaign to be created in an "off" state, we can create the campaign initially off, and then update the campaign status to on immediately afterwards to simulate building an enabled campaign from the onset.

There's still some weird behavior that occurred before this PR where a campaign that is in a COMPLETE state cannot be set to OFF. We should probably add better handling for this edge case in a future fix.


charliecon:
@bbbco Thanks for the contribution! 

When trying to set the campaign status from `complete` to `off`, the API returns the error message: 
> "An attempt was made to update a Campaign in an invalid way. Details: Invalid transition from \<COMPLETE\> to \<OFF\>"

This is a tricky one because, as a workaround, we could set `ForceNew: true` on the campaign_status field, which would cause the campaign to be destroyed and recreated, so that the above edge case would be possible. However, that would mean valid transitions such as `off` -> `on` would also trigger the same behaviour. Is there a way to update `complete` to `off` that the API allows?

bbbco:
Yes, I came across this edge case. However, I believe the edge case occurred before this PR change.

I did reach out to the Outbound Genesys Cloud group and was able to confirm the following info:

- You can update any fields on the campaign at once EXCEPT the `campaignStatus` field
- If you want to update `campaignStatus`, it must be the only change in the update
- If you update any other fields, the `campaignStatus` must remain the same value in the update request
- Regarding status field values, `COMPLETE` is not `ON`, but rather means the campaign is `OFF` and has finished processing all contacts in the contact list
- Going from `COMPLETE` to `OFF` is not considered a valid transition; instead `COMPLETE` ~= `OFF` + 100% progress of completing the campaign. 
- It appears that a campaign can either be `ON` or in any other state. It can transition from any other state to `ON` or `ON` to `OFF`, but the other states are set by the server and not user defined (`STOPPING`, `COMPLETE`, `INVALID`, `FORCED_OFF`, `FORCED_STOPPING`)

Considering this new info, I think I can update the logic and tests to properly account for this behavior.

bbbco:
@charliecon I revamped things a bit to more closely align the behavioral expectations to that of the Genesys Cloud API. I also added tests that confirm that the provider can go OFF -> ON -> OFF and ON -> OFF status wise.

I had to use an unconventional means to do so, as the campaign wouldn't stay running long enough because the campaigns had no contacts to attempt to call. Considering we don't have the ability to add contacts to a `contact_list` in the provider yet, I had to setup the initial supporting resources in the first step and use a check function (since it passes in the state) to figure out the `contact_list` ID so we could use the SDK to add a few contacts before building the rest of the outbound campaign. I imagine that we should eventually add support to add contacts to a contact list in the future, but that felt out of scope for this PR. I'm glad I got this all working, and it passes all the tests!

.vscode/launch.json
```diff
@@ -1,36 +1,35 @@
 {
-    // Use IntelliSense to learn about possible attributes.
```
carnellj-genesys, line 2: This shouldnt be checked into source control.   I would add this to the .gitignore.

carnellj-genesys (approved):

charliecon (approved):

carnellj-genesys (approved):

PR #481, opened 2023-03-22
Updating go SDK versions, Adding timeout to integration action config…

Tyler-Bath:
Updating go SDK, adding timeout to integration Action


Tyler-Bath:
There are a few places that needed to be updated due to the update in sdk. Ive put temporary empty values as the inputs to the new arguments. From what i can tell in the public api: The wrapupcodes - they are not used unless a feature toggle is turned on which it is only on for some orgs in lowers. The SMS - does not seem to use the new arg at all in public api.
Not sure if we should be reaching out to those teams to figure out what to do.

carnellj-genesys (approved):

charliecon (approved):

PR #487, opened 2023-03-29
DEVENGAGE-1719: Update the GO SDK so that I could use the new role lo…

carnellj-genesys:
…ok up endpoint with the usecount disabled.

### DIFF
--- a/genesyscloud/data_source_genesyscloud_routing_email_domain.go
+++ b/genesyscloud/data_source_genesyscloud_routing_email_domain.go
@@ -8,7 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/mypurecloud/platform-client-sdk-go/v94/platformclientv2"
+	"github.com/mypurecloud/platform-client-sdk-go/v95/platformclientv2"
 )
 
 // Returns the schema for the routing email domain
@@ -37,7 +37,7 @@ func dataSourceRoutingEmailDomainRead(ctx context.Context, d *schema.ResourceDat
 		for pageNum := 1; ; pageNum++ {
 			const pageSize = 100
 
-			domains, _, getErr := routingAPI.GetRoutingEmailDomains(pageSize, pageNum, false)
+			domains, _, getErr := routingAPI.GetRoutingEmailDomains(pageSize, pageNum, false, "")
 
 			if getErr != nil {
 				return resource.NonRetryableError(fmt.Errorf("Error requesting email domain %s: %s", name, getErr))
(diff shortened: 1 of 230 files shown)


genesyscloud/data_source_genesyscloud_routing_email_domain.go
```diff
@@ -37,7 +37,7 @@ func dataSourceRoutingEmailDomainRead(ctx context.Context, d *schema.ResourceDat
 		for pageNum := 1; ; pageNum++ {
 			const pageSize = 100
 
-			domains, _, getErr := routingAPI.GetRoutingEmailDomains(pageSize, pageNum, false)
+			domains, _, getErr := routingAPI.GetRoutingEmailDomains(pageSize, pageNum, false, "")
```
carnellj-genesys, line 40: Email API added a new query parameter so had to adjust the API call.

genesyscloud/resource_genesyscloud_routing_queue.go
```diff
@@ -595,22 +595,25 @@ func readQueue(ctx context.Context, d *schema.ResourceData, meta interface{}) di
...
-				d.Set("media_settings_chat", flattenMediaSetting(chatSettings))
-			}
-			if emailSettings, ok := (*currentQueue.MediaSettings)[mediaSettingsKeyEmail]; ok {
-				d.Set("media_settings_email", flattenMediaSetting(emailSettings))
-			}
-			if messageSettings, ok := (*currentQueue.MediaSettings)[mediaSettingsKeyMessage]; ok {
-				d.Set("media_settings_message", flattenMediaSetting(messageSettings))
-			}
+
+		if currentQueue.MediaSettings != nil && currentQueue.MediaSettings.Call != nil {
```
carnellj-genesys, line 599: This was a bit of screwery.  The JSON body did not change, but the API owners moved the swagger def from being a map to a specific object with values for each object.  So I had to go through and change some of the code for the routing queues to move away from using a map to using individual objects.

genesyscloud/resource_genesyscloud_auth_role.go
```diff
@@ -234,7 +234,7 @@ func readAuthRole(ctx context.Context, d *schema.ResourceData, meta interface{})
 	log.Printf("Reading role %s", d.Id())
 
 	return withRetriesForRead(ctx, d, func() *resource.RetryError {
-		role, resp, getErr := authAPI.GetAuthorizationRole(d.Id(), nil)
+		role, resp, getErr := authAPI.GetAuthorizationRole(d.Id(), false, nil)
```
carnellj-genesys, line 237: This is the new field call that will disable usercount.  This is an extremely expensive operation and we don't use if for CX as Code

go.mod
```diff
@@ -61,6 +61,7 @@ require (
 	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
+	github.com/mypurecloud/platform-client-sdk-go/v94 v94.0.1 // indirect
```
charliecon, line 64: Need to run `go mod tidy` to get rid of the dependency. GitHub Actions can't seem to handle having two versions of our platform sdk at the same time

charliecon (approved):

PR #488, opened 2023-04-03
Devengage-1670: Chunking dnis array upload to ivr configs 

charliecon:


genesyscloud/resource_genesyscloud_architect_ivr.go
```diff
@@ -327,3 +370,27 @@ func flattenIvrDnis(dnisArr *[]string) *schema.Set {
 	}
 	return nil
 }
+
+func uploadChunk(d *schema.ResourceData, chunk []string, architectApi *platformclientv2.ArchitectApi) diag.Diagnostics {
+	var dnis []string
+
+	ivr, _, getErr := architectApi.GetArchitectIvr(d.Id())
```
carnellj-genesys, line 377: It would be cool if we could proxy out the GetArchitectIvr so that we could test this one function without any other dependencies.

carnellj-genesys (approved):

genesyscloud/proxies/architect_api/architect_ivr.go
```diff
@@ -0,0 +1,98 @@
...
+		log.Printf("Uploading block %v of DID numbers to ivr config %s", i+1, id)
+		putIvr, resp, err := a.uploadDnisChunk(ivr, chunk)
+		if err != nil {
+			return putIvr, resp, err
+		}
+		ivr = putIvr
+	}
+
+	return ivr, nil, nil
+}
```
charliecon, line 79: @carnellj-genesys
What do you think of placing these chunkUpload functions inside the proxy folder, and testing them in there? It makes the resource files look cleaner but maybe the proxy package is taking on more responsibility than it should?

PR #491, opened 2023-04-07
Changing emergency number to fix cicd tests and refactoring

charliecon:
The only change here for fixing the test was changing the phone number used in the location to avoid clashing with other tests that use the same number. Everything else is refactoring to move away from the function `generateOutboundCampaignBasic`


genesyscloud/data_source_genesyscloud_outbound_campaignrule_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func TestAccDataSourceCampaignRule(t *testing.T) {
-	t.Parallel()
+	//t.Parallel()
```
carnellj-genesys, line 13: Remove this commented out code

carnellj-genesys (approved):

dginty4 (approved):

HemanthDogiparthi12:
looks like the consistency_checker failing ??

charliecon:
> looks like the consistency_checker failing ??

@HemanthDogiparthi12 That message is appearing on the main branch as well. I assume if we get all tests passing in a build, that it wouldn't appear.

PR #493, opened 2023-04-11
Devengage: 1761 cleanup err handling

carnellj-genesys:


genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go
```diff
@@ -680,10 +680,10 @@ func TestAccResourceTfExportUserPromptExportAudioFile(t *testing.T) {
 					falseValue,
 					falseValue,
 				) + gcloud.GenerateUserPromptResource(&gcloud.UserPromptStruct{
-					userPromptResourceId,
-					userPromptName,
-					strconv.Quote(userPromptDescription),
-					userPromptResources,
+					ResourceID:  userPromptResourceId,
```
carnellj-genesys, line 683: Refactored this because Go best practice is to explicitly have key names in your struct definition.  Caught this via go vet.

genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go
```diff
@@ -697,10 +697,10 @@ func TestAccResourceTfExportUserPromptExportAudioFile(t *testing.T) {
 			// Update to two resources with separate audio files
 			{
 				Config: gcloud.GenerateUserPromptResource(&gcloud.UserPromptStruct{
-					userPromptResourceId,
-					userPromptName,
-					strconv.Quote(userPromptDescription),
-					userPromptResources2,
+					ResourceID:  userPromptResourceId,
```
carnellj-genesys, line 700: Refactored this because Go best practice is to explicitly have key names in your struct definition.  Caught this via go vet.

genesyscloud/util_files.go
```diff
@@ -2,119 +2,113 @@ package genesyscloud
...
-	} else {
-		file, err = os.Open(path)
-		if err != nil {
-			return nil, nil, err
-		}
-		reader = file
-	}
-
-	return reader, file, nil
+type S3Uploader struct {
```
carnellj-genesys, line 15: Create a new struct called S3Uploader.

genesyscloud/util_files.go
```diff
@@ -2,119 +2,113 @@ package genesyscloud
...
-	if file != nil {
-		defer file.Close()
-	}
-
-	_, err = io.Copy(bodyBuf, reader)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to copy file content to the handler. Error: %s ", err)
-	}
-
+func (s *S3Uploader) substituteValues(bodyBuf *bytes.Buffer) {
```
carnellj-genesys, line 37: Broke this out of the upload method and made this its own method.  This code can be a standalone method and also makes it easier to unit test.

genesyscloud/util_files.go
```diff
@@ -2,119 +2,113 @@ package genesyscloud
...
 
 		bodyBuf.Reset()
 		bodyBuf.WriteString(fileContents)
 	}
+}
 
-	req, _ := http.NewRequest("PUT", presignedUrl, bodyBuf)
-	for key, value := range headers {
-		req.Header.Set(key, value)
+func (s *S3Uploader) Upload() ([]byte, error) {
```
carnellj-genesys, line 50: Move the prepareandUpload code to any Upload function associated with a struct.

genesyscloud/util_files.go
```diff
@@ -2,119 +2,113 @@ package genesyscloud
...
+	bodyBuf := &bytes.Buffer{}
+
+	_, err := io.Copy(bodyBuf, s.reader)
```
carnellj-genesys, line 53: Notice the reader.  Originally we were passing in the filename and then looking up the file inside of the method.  Using a reader allows us to stub out the test case without having to have a file written out.

genesyscloud/util_files_test.go
```diff
@@ -0,0 +1,125 @@
...
+						en-us:
+							defaultLanguageSkill:
+							noValue: true
+							textToSpeech:
+							defaultEngine:
+								voice: Jill`
+	fileReader := strings.NewReader(yamlFile)
+
+	// Create a mock HTTP server
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
carnellj-genesys, line 36: This is the money.  Go has a built in mock server where we can register handlers to mock out calls.  This allows us to write test cases where an API call is being made without having to do anything special to the actual underlying code (e.g. like exposing the httpClient).

Theoritically we could mock out the entire set of calls to our public API and never have a resource test actually touch Genesys Cloud.  This would significantly simplify our dependency on having a Genesys Cloud resource available.

genesyscloud/util_files_test.go
```diff
@@ -0,0 +1,125 @@
...
+	_, err := s3Uploader.Upload()
+
+	expectedResult := fmt.Sprintf("failed to upload file to S3 bucket with an HTTP status code of %d", http.StatusBadRequest)
+	if err != nil {
+		assert.Equal(t, err.Error(), expectedResult)
+	}
+}
+
+// TestSubstitutions will test the substitution replacement in the S3Upload
+func TestSubstitutions(t *testing.T) {
```
carnellj-genesys, line 105: Again look for tight cohesion within a method.  Small methods become very easy to unit test and will allow us to drive our code coverage numbers up.  We should have a large number of test cases and a small number of acceptance tests.

genesyscloud/util_retries.go
```diff
@@ -19,7 +19,8 @@ import (
 func withRetries(ctx context.Context, timeout time.Duration, method func() *resource.RetryError) diag.Diagnostics {
 	err := diag.FromErr(resource.RetryContext(ctx, timeout, method))
 	if err != nil && strings.Contains(fmt.Sprintf("%v", err), "timeout while waiting for state to become") {
-		ctx, _ := context.WithTimeout(context.Background(), timeout)
```
carnellj-genesys, line 22: We should never ignore the cancel object and should explicitly do a defer cancel() on the returned function.  This was caught by go vet.

genesyscloud/util_retries.go
```diff
@@ -39,7 +40,8 @@ func withRetriesForReadCustomTimeout(ctx context.Context, timeout time.Duration,
 		errStringLower := strings.ToLower(fmt.Sprintf("%v", err))
 		if strings.Contains(errStringLower, "timeout while waiting for state to become") ||
 			strings.Contains(errStringLower, "context deadline exceeded") {
-			ctx, _ := context.WithTimeout(context.Background(), timeout)
+			ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
carnellj-genesys, line 43: We should never ignore the cancel object and should explicitly do a defer cancel() on the returned function.  This was caught by go vet.

scripts/hooks/pre-commit
```diff
@@ -4,9 +4,10 @@ STAGED_GO_FILES=$(git diff --cached --name-only -- '*.go')
 if [[ $STAGED_GO_FILES == "" ]]; then
     echo "no go files updated"
 else
-    echo "Formatting go files"
+    echo "Formatting and vetting code go files"
```
carnellj-genesys, line 7: Adding vetting to the checking process.  Note:  You need to do a make install-hooks to get this installed locally

charliecon (approved):

HemanthDogiparthi12 (approved):

PR #499, opened 2023-04-18
Feat/devengage 1614: Script resource

charliecon:


examples/resources/genesyscloud_script/apis.md
```diff
@@ -0,0 +1,5 @@
+* [POST /api/v2/scripts/{scriptId}/export](https://developer.genesys.cloud/devapps/api-explorer#post-api-v2-scripts--scriptId--export)
+* [GET /api/v2/scripts](https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-scripts)
```
charliecon, line 2: A private API was used to create the script
I'm not sure if I should include something like `POST https://apps.{domains}/uploads/v2/scripter`. I would just put `/uploads/v2/scripter` but I noticed it's `apps`, and not `api`

examples/resources/genesyscloud_script/apis.md
```diff
@@ -0,0 +1,5 @@
...
+* [GET /api/v2/scripts/{scriptId}](https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-scripts--scriptId-)
+* [GET /api/v2/scripts/uploads/{uploadId}/status](https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-scripts-uploads--uploadId--status)
+* DELETE /api/v2/scripts/{scriptId}
```
charliecon, line 5: Would it be better to omit this line? Since this endpoint isn't publicly listed yet

examples/resources/genesyscloud_script/apis.md
charliecon: @dginty4 @carnellj

examples/resources/genesyscloud_script/apis.md
charliecon: @dginty4 @carnellj

carnellj-genesys: Omit the line if it is not public

carnellj-genesys (approved):

PR #507, opened 2023-04-28
Fixing knowledgebase export process

charliecon:
Because the exporter function was trying to use the pageNumber system when it should have been using cursor logic, it was getting stuck in an infinite loop. Trusting the terraform code generator too much probably caused this one.

It also wasn't exporting both published and unpublished entities.

### DIFF
--- a/genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
+++ b/genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
@@ -22,7 +22,7 @@ func getAllKnowledgeKnowledgebases(_ context.Context, clientConfig *platformclie
 
 	publishedEntities, err := getAllKnowledgebaseEntities(*knowledgeAPI, true)
 	if err != nil {
-		return nil, diag.Errorf("%v", err)
+		return nil, err
 	}
 
 	for _, knowledgeBase := range *publishedEntities {
@@ -31,7 +31,7 @@ func getAllKnowledgeKnowledgebases(_ context.Context, clientConfig *platformclie
 
 	unpublishedEntities, err := getAllKnowledgebaseEntities(*knowledgeAPI, false)
 	if err != nil {
-		return nil, diag.Errorf("%v", err)
+		return nil, err
 	}
 
 	for _, knowledgeBase := range *unpublishedEntities {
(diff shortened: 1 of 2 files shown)


genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
```diff
@@ -20,23 +20,64 @@ func getAllKnowledgeKnowledgebases(_ context.Context, clientConfig *platformclie
...
+		after    string
+		entities []platformclientv2.Knowledgebase
+	)
+
+	const pageSize = 100
+	for i := 0; ; i++ {
+		knowledgeBases, _, getErr := knowledgeApi.GetKnowledgeKnowledgebases("", after, "", fmt.Sprintf("%v", pageSize), "", "", published, "", "")
 		if getErr != nil {
-			return nil, diag.Errorf("Failed to get page of knowledge bases: %v", getErr)
+			return nil, fmt.Errorf("Failed to get page of knowledge bases: %v", getErr)
```
carnellj-genesys, line 54: Hey Charlie why did you use a diag error vs a a go error?  Not questioning the decision just seeking to understand the issue.

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
```diff
@@ -20,23 +20,64 @@ func getAllKnowledgeKnowledgebases(_ context.Context, clientConfig *platformclie
...
 
 		if knowledgeBases.Entities == nil || len(*knowledgeBases.Entities) == 0 {
 			break
 		}
 
-		for _, knowledgeBase := range *knowledgeBases.Entities {
-			resources[*knowledgeBase.Id] = &ResourceMeta{Name: *knowledgeBase.Name}
+		entities = append(entities, *knowledgeBases.Entities...)
+
+		if knowledgeBases.NextUri == nil || *knowledgeBases.NextUri == "" {
```
carnellj-genesys, line 63: The nil check is unnecessary.  In Go if NextURI it will automatically initialize the empty string.  An item will only if be nil if it is a struct, map, slice or array.

carnellj-genesys (approved):

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
carnellj-genesys: A few small comments.

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
charliecon: No reason. Just changed it to diag, definitely makes more sense

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
charliecon: Done ✅

genesyscloud/resource_genesyscloud_knowledge_knowledgebase.go
```diff
@@ -31,12 +58,26 @@ func getAllKnowledgeKnowledgebases(_ context.Context, clientConfig *platformclie
...
```
charliecon, line 63: @carnellj-genesys Turns out this check is needed because the api response json doesn't include any field `NextUri` once it reaches the last page. If anything the check to see if the string is empty is probably superfluous , but I think I'll just leave it in to be safe

PR #509, opened 2023-05-02
Refactoring architect ivr proxy code

charliecon:


genesyscloud/proxies/architect_api/architect_ivr.go
```diff
@@ -29,36 +28,90 @@ type ArchitectIvrProxy struct {
...
+		dnisChunks = utillists.ChunkStringSlice(*ivr.Dnis, a.MaxDnisPerRequest)
+		if len(dnisChunks) != 1 {
+			ivr.Dnis = nil
+		}
+	}
+
+	respIvr, resp, err := a.putArchitectIvrBasic(a, id, ivr)
+
+	if len(dnisChunks) > 1 {
+		respIvr, _, err = uploadIvrDnisChunks(a, dnisChunks, id)
```
carnellj-genesys, line 94: Just curious as to why you did not make the uploadIvrDnsChunks as a private method on the structure.  My main concern with the previous rework is having the uploadIvrDnsChunks being called from outside the struct when the code that relied on it was in a struct.

carnellj-genesys (approved):

genesyscloud/proxies/architect_api/architect_ivr.go
charliecon: I was thinking if this function is only needed internally within the package, it wouldn't belong on the struct. And I assumed the method name starting with a lower case letter would make it private
Are you suggesting defining the method as `func (a  *ArchitectIvrProxy) uploadIvrDnisChunks(...` rather than pasing the struct object in as a param. Thinking about it now, that would look better here

charliecon: @carnellj-genesys Ah, I think I needed more time to sit with this. I should have kept the uploadChunk methods as part of the struct so I can mock them out and test the outer function (attached to the public field) that calls them

PR #515, opened 2023-05-16
Adding support for new contact list field + minor refactoring of tests

charliecon:


genesyscloud/resource_genesyscloud_outbound_contactlist.go
```diff
@@ -451,3 +493,65 @@ func flattenSdkOutboundContactListContactEmailAddressColumnSlice(contactEmailAdd
...
+				sdkColumnDataTypeSpecification.MaxLength = &maxLengthInt
+			}
+			sdkColumnDataTypeSpecificationsSlice = append(sdkColumnDataTypeSpecificationsSlice, sdkColumnDataTypeSpecification)
+		}
+	}
+
+	return &sdkColumnDataTypeSpecificationsSlice
+}
+
+func flattenSdkOutboundContactListColumnDataTypeSpecifications(columnDataTypeSpecifications []platformclientv2.Columndatatypespecification) []interface{} {
```
HemanthDogiparthi12, line 529: where is this function used ..

genesyscloud/resource_genesyscloud_outbound_messagingcampaign_test.go
```diff
@@ -30,18 +30,17 @@ func TestAccResourceOutboundMessagingCampaign(t *testing.T) {
 		contactListResource   = generateOutboundContactList(
 			contactListResourceId,
 			contactListName,
-			"",
-			"",
+			nullValue,
```
HemanthDogiparthi12, line 33: for understanding purpose. why is this changed from empty string to nullValue ?

genesyscloud/resource_genesyscloud_outbound_ruleset_test.go
```diff
@@ -240,19 +240,20 @@ func TestAccResourceOutboundRuleset(t *testing.T) {
...
 					automaticTimeZoneMapping,
-					"",
-					"",
+					nullValue,
+					nullValue,
 					generatePhoneColumnsBlock(
 						"Cell",
 						"cell",
-						"Cell",
+						strconv.Quote("Cell"),
```
HemanthDogiparthi12, line 253: is there any test impact with this change from previous to now.

genesyscloud/resource_genesyscloud_outbound_messagingcampaign_test.go
charliecon: Previously, the generate function was creating the resource as a string, and to exclude certain fields from the resource, the function contains an if statement to check if the string was empty. If it was empty, then the line would be omitted. It looked something like this:
```golang
func generateResource(name string) string {
    if name != "" {
        name = fmt.Sprintf(`name = "%s"`, name)
    }
    return fmt.Sprintf(`
    resource "example" "resource" {
        %s
    }
`, name)
}
```

Now, instead, I'm passing in `nullValue` (which is "null"). Setting the field to null does the exact same thing as omitting the field altogether, and this approach also makes the generate func more readable.

```golang
func generateResource(name string) string {  
    return fmt.Sprintf(`
    resource "example" "resource" {
        name = %s
    }
`, name)
}
```

Passing in `nullValue` evaluates to `name = null`, and to add a value you need to wrap the data in quotation marks when passing it to the function - this is what the `strconv.Quote()` func does

genesyscloud/resource_genesyscloud_outbound_ruleset_test.go
charliecon: No, I've been keeping an eye on the matrix tests. They are still passing. I'll make sure again before merging

carnellj-genesys (approved):

PR #520, opened 2023-05-18
Update maximum_capacity range to include zero

bbbco:
This is a partial fix to https://inindca.atlassian.net/browse/DEVENGAGE-1868

This change allows a `maximum_capacity` value to be set to 0-25, not just 1-25.

The API allows this, and we can see that it is confirmed in the API errors:

```
PUT /api/v2/routing/users/3b19297d-64f5-4136-b2be-baeda46e23ea/utilization HTTP/1.1
Content-Type: application/json
Authorization: Bearer XXXXX
Host: api.mypurecloud.com
Content-Length: 925

{
	"utilization": {
		"call": {
			"maximumCapacity": 26,
			"interruptableMediaTypes": [
				"chat",
				"email",
				"message"
			],
			"includeNonAcd": false
		},
		"chat": {
			"maximumCapacity": 1,
			"interruptableMediaTypes": [
				"email",
				"message"
			],
			"includeNonAcd": false
		},
		"email": {
			"maximumCapacity": 2,
			"interruptableMediaTypes": [
				"call",
				"callback",
				"chat"
			],
			"includeNonAcd": false
		},
		"message": {
			"maximumCapacity": 1,
			"interruptableMediaTypes": [
				"chat",
				"email"
			],
			"includeNonAcd": false
		},
		"callback": {
			"maximumCapacity": 0,
			"interruptableMediaTypes": [
				"chat",
				"email",
				"message"
			],
			"includeNonAcd": false
		},
		"workitem": {
			"maximumCapacity": 4,
			"interruptableMediaTypes": [
				"call",
				"callback",
				"chat",
				"email",
				"message"
			],
			"includeNonAcd": true
		}
	},
	"level": "Agent"
}
---
{
	"message": "Maximum capacity must be between 0 and 25.",
	"code": "invalid.media.capacity",
	"status": 400,
	"messageWithParams": "Maximum capacity must be between 0 and 25.",
	"messageParams": {},
	"contextId": "260efd02-4360-4c94-991c-6d691a93148d",
	"details": [],
	"errors": []
}
```

---

This is only a partial fix for  DEVENGAGE-1868, as this does not address the issue with the exporter rendering the `maximum_capacity` attribute if the value is zero.

### DIFF
--- a/docs/resources/routing_utilization.md
+++ b/docs/resources/routing_utilization.md
@@ -67,7 +67,7 @@ resource "genesyscloud_routing_utilization" "org-utililzation" {
 
 Required:
 
-- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 1 and 25.
+- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 0 and 25.
 
 Optional:
 
@@ -80,7 +80,7 @@ Optional:
 
 Required:
 
-- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 1 and 25.
+- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 0 and 25.
 
 Optional:
 
@@ -93,7 +93,7 @@ Optional:
 
 Required:
 
-- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 1 and 25.
+- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 0 and 25.
 
 Optional:
 
@@ -106,7 +106,7 @@ Optional:
 
 Required:
 
-- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 1 and 25.
+- `maximum_capacity` (Number) Maximum capacity of conversations of this media type. Value must be between 0 and 25.
 
 Optional:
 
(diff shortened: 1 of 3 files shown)


carnellj-genesys (approved):

charliecon (approved):

charliecon (commented):
This is the error I'm seeing. I'm guessing it comes down to some difference between the orgs? @carnellj-genesys 
```json
{
  "message": "Maximum capacity must be between 1 and 50",
  "code": "invalid.media.capacity",
  "status": 400,
  "messageWithParams": "Maximum capacity must be between 1 and 50",
  "messageParams": {},
  "contextId": "382acc9f-34fb-468c-af6a-f1f17116c855",
  "details": [],
  "errors": []
}
```

bbbco:
I reached out to the Genesys Cloud ACD Routing channel on Genesys Cloud chat and got this response:
```
Seth McCarus-9:35 AM
the org-level utilization range is 1 to 50.
the agent-level utilization range is 0 to 25.

I think that when we implemented agent-level utilization, the existing org-level max of 50 seemed to high, so we set a lower agent-level max value.  and of course it makes sense to allow a max of 0 at an agent level but not at an org level.
```

Given this response, it seems that there are two different endpoints: one for org wide (1 to 5) and one for users (0 to 25). I believe that the range in this PR is the correct range that should be used.

PR #533, opened 2023-06-07
Fix DEVENGAGE-1923 - Enable campaign rules / sequences from the point of creation

bbbco:
This PR fixes https://inindca.atlassian.net/browse/DEVENGAGE-1923

### DIFF
--- a/genesyscloud/resource_genesyscloud_outbound_sequence.go
+++ b/genesyscloud/resource_genesyscloud_outbound_sequence.go
@@ -40,13 +40,13 @@ func resourceOutboundSequence() *schema.Resource {
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
 			`status`: {
-				Description:  `The current status of the CampaignSequence. A CampaignSequence can be turned 'on' or 'off'.`,
+				Description:  `The current status of the CampaignSequence. A CampaignSequence can be turned 'on' or 'off' (default).`,
 				Optional:     true,
 				Computed:     true,
 				Type:         schema.TypeString,
 				ValidateFunc: validation.StringInSlice([]string{`on`, `off`}, false),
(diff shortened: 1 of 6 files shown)


genesyscloud/resource_genesyscloud_outbound_sequence.go
```diff
@@ -120,6 +124,15 @@ func createOutboundSequence(ctx context.Context, d *schema.ResourceData, meta in
 
 	d.SetId(*outboundSequence.Id)
 
+	// Campaigns sequences can be enabled after creation
+	if status != "" && status == "on" {
```
charliecon, line 128: I guess this can just be `if status == "on"` ?

charliecon (approved):

carnellj-genesys (approved):

genesyscloud/resource_genesyscloud_outbound_sequence.go
bbbco: Sure.

charliecon:
Hey @bbbco 

I just reverted the changes this PR made after I noticed one of the test cases was failing about 3/5 times (locally and in GitHub actions)

[https://github.com/MyPureCloud/terraform-provider-genesyscloud/actions/runs/5241688173/jobs/9464173640#step:6:526](https://github.com/MyPureCloud/terraform-provider-genesyscloud/actions/runs/5241688173/jobs/9464173640#step:6:526)

It looks like the update function called from the create is updating `status` to "on" , and the next read at the end of this update returns `status: "on"` (good), but the final read at the end of the create function returns `status: "off"`. 

It seems to be a bug coming from the API, but I'm not certain. Is it possible that, when unsuccessfully switching on the sequence, status is just set back to "off" after being "on" for a moment? 

This test failure doesn't necessarily mean the resource doesn't behave how it should, but we're trying to clean up our test cases so that our workflow is as accurate as possible when indicating the app's health.

PR #536, opened 2023-06-09
Campaigns, campaign rules, and sequences can be turned off at point of destroy

bbbco:
In conjunction with [this PR](https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/533) I added functionality and tests to ensure that campaigns, campaign rules, and sequences can be turned off at point of destroy. 

Otherwise, we see the following errors when running destroy:

```
│ Error: Exhausted retries. Last error: [{0 Failed to delete Outbound Campaign: API Error: 400 - Cannot delete a campaign when it is running (f0143ffa-3ba8-4ef2-a1de-17e046f1fe14)  []}]
```

```
│ Error: Exhausted retries. Last error: [{0 Failed to delete Outbound Campaign Rule: API Error: 400 - Cannot delete an enabled campaign rule (ede90978-2d89-4b7c-8ed8-7072415ee8c4)  []}]
```

Since I had also added enable functionality to the outbound sequences resource in the prior PR, I went ahead and added a test to ensure we could still delete with an enabled sequence (the API doesn't require you to disable it before deleting).

### DIFF
--- a/genesyscloud/resource_genesyscloud_outbound_campaign.go
+++ b/genesyscloud/resource_genesyscloud_outbound_campaign.go
@@ -266,6 +266,9 @@ func outboundCampaignExporter() *ResourceExporter {
 			`callable_time_set_id`: {
 				RefType: "genesyscloud_outbound_callabletimeset",
 			},
+			`script_id`: {
+				RefType: "genesyscloud_script",
+			},
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


charliecon (approved):

charliecon:
@bbbco  Just some merge conflicts from the 'enabled on creation' PR 

Thanks for all the contributions!

carnellj-genesys (approved):

PR #537, opened 2023-06-10
added CGR config to queue

igorgordienko:


igorgordienko:
added cgr queues config

PR #542, opened 2023-06-13
DEVENGAGE-1925: Added primary and secondary sites

carnellj-genesys:
Hey guys,

I have added primary_sites and secondary_sites attributes to the site resource.  Please note there were no enhancements to the test because these fields are only applicable when working with physical edges.  I had to borrow a physical edge in test and manually tested the code.  Let me know if you have any questions or concerns


charliecon:
@carnellj-genesys It looks like the primary and secondary arrays are populated with the id of site "PureCloud Voice - AWS" by default. I created a basic site with version 1.16.0 and saw the same thing when I inspected it on the api explorer - both arrays containing the id for the purecloud aws site. The only reason there was no plan not empty error there is that the fields aren't defined in the schema, so the values aren't read in. 

I think you just need to set `Computed: true` for both primary and seconary_sites.

carnellj-genesys:
> @carnellj-genesys It looks like the primary and secondary arrays are populated with the id of site "PureCloud Voice - AWS" by default. I created a basic site with version 1.16.0 and saw the same thing when I inspected it on the api explorer - both arrays containing the id for the purecloud aws site. The only reason there was no plan not empty error there is that the fields aren't defined in the schema, so the values aren't read in.
> 
> I think you just need to set `Computed: true` for both primary and seconary_sites.

Yeah.  I came to the same conclusion last night that the values were probably being pre-populated.  I will try the computed: true and give it a try.  Thanks for the help.

HemanthDogiparthi12 (approved):

PR #547, opened 2023-06-16
Updating to v102 and fixing breaks caused by this

charliecon:


genesyscloud/resource_genesyscloud_knowledge_document_variation.go
```diff
@@ -582,7 +600,7 @@ func buildDocumentListBlocks(blocksIn map[string]interface{}) *[]platformclientv
 			varType := blockMap["type"].(string)
 			blockOut := platformclientv2.Documentbodylistblock{
 				VarType: &varType,
-				Blocks:  buildDocumentContentBlocks(blockMap),
+				Blocks:  buildDocumentContentListBlocks(blockMap),
```
charliecon, line 603: @diemenschmaschine92 requesting your review on this because the new version of the sdk reflects some knowledgebase api changes for document variations. 
Looks like this `Blocks` field changed from type `Documentcontentblock` to the new `Documentlistcontentblock` in some places and not others. 
I think the changes I made suffice. The tests are passing still anyway.

diemenschmaschine92 (approved):

genesyscloud/resource_genesyscloud_knowledge_document_variation.go
diemenschmaschine92: Are you planning on merging this before this one?
https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/545
So I know if I need to pull in changes and resolve any conflicts.
That's the PR that will fix the export bug in production.

charliecon: > @charliecon Are you planning on merging this before this one? #545 So I know if I need to pull in changes and resolve any conflicts. That's the PR that will fix the export bug in production.

I think it's best to merge this one first, and then fix any conflicts in your PR, if that's okay. I noticed you added a few calls to knowledgebase get all functions, and they might have changed slightly with this new version

PR #560, opened 2023-06-26
Outbound Ruleset Package

HemanthDogiparthi12:
This PR have two main goals
One is to resolve bug 1877 where the skills attributes are not being resolved properly by the exporter. These changes are in the commit [changes for 1877 exporter for skills](https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/560/commits/3e633e5ff255f980a3132e10b2c1bdf0c6fdd94b)

and the other is to properly isolate and package outboundruleset. The remaining commits are part of attaining this.

Changes for Resource Injection to Provider:

We have taken an approach to inject the required resources, datasources and exporters via default init for the newly decoupled individual resource packages. One problem that is identified for this is during the test case execution if there are dependencies with other individual packages, the resources are not identified , since the test cases run in their respective package test-binaries and there is no visibility for the outside packages unless  you have them imported.
The import again here is a problem since there are unavoidable cyclic dependencies which cannot be solved even with interface injection. 

This PR tries to address the resource injection problem moving out the responsibility from provider to the Main class. 

1. The Main class will inject the Resources,Exporters, and DataSources via Interface , thereby decoupling the import cycle dependency.
2. For the Test cases , each individual package will inject resources required for the package in a separate init_test class confined to each package.
3. The resource_exporter is moved to package of its own for the import cycle problems and it makes sense to have it in an individual package of it own and the dependency tree is liner.
4. Separate packages for outbound , since there are lot of interdependency between these.  and decoupled outbound_contact_list and outbound_attempt_limit, since the outbound_ruleset package depends on contact_list resource for its test cases and contact list depends on attempt_limit.
5. Some more util functions are moved to their own package.
6. Used TestMain for each package for initialising the test resources and running the Test Suite for that package.
7. Update to Latest platform-client sdk version.


carnellj-genesys (approved):

charliecon (approved):
Just some merge conflicts that need fixing

genesyscloud/outbound/resource_genesyscloud_outbound_contactlistfilter.go
```diff
@@ -104,8 +106,8 @@ var (
 	}
 )
 
-func getAllOutboundContactListFilters(_ context.Context, clientConfig *platformclientv2.Configuration) (ResourceIDMetaMap, diag.Diagnostics) {
-	resources := make(ResourceIDMetaMap)
+func getAllOutboundContactListFilters(_ context.Context, clientConfig *platformclientv2.Configuration) (resource_exporter.ResourceIDMetaMap, diag.Diagnostics) {
+	resources := make(resource_exporter.ResourceIDMetaMap)
```
carnellj-genesys, line 110: No snake_case

carnellj-genesys (changes requested):

genesyscloud/outbound_ruleset/outbound_ruleset_resource_init_test.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (r *registerTestInstance) registerTestResources() {
+	providerResources["genesyscloud_outbound_ruleset"] = ResourceOutboundRuleset()
+	providerResources["genesyscloud_routing_queue"] =  gcloud.ResourceRoutingQueue()
+}
+
+func (r *registerTestInstance) registerTestDataSources() {
+	providerDataSources["genesyscloud_outbound_ruleset"] =  DataSourceOutboundRuleset()
+	providerResources["genesyscloud_outbound_contact_list"] = outbound..ResourceOutboundContactList()
```
charliecon, line 36: Need to remove the `outbound..`

carnellj-genesys (approved):

PR #570, opened 2023-07-18
Devengage 1992 add scripts publishing

carnellj-genesys:
Hi @charliecon @dginty4 @HemanthDogiparthi12 

This work makes sure that we publish a script after we create it.  Previously the scripts team did not make their publish endpoint available.  The endpoint is available and I implemented the publication work.  I also added code to filter out any scripts from the getAll call that were not published.

Thanks,
    John

### DIFF
--- a/genesyscloud/resource_genesyscloud_script.go
+++ b/genesyscloud/resource_genesyscloud_script.go
@@ -14,7 +14,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/mypurecloud/platform-client-sdk-go/v103/platformclientv2"
+	"github.com/mypurecloud/platform-client-sdk-go/v105/platformclientv2"
 )
 
 func getAllScripts(ctx context.Context, clientConfig *platformclientv2.Configuration) (ResourceIDMetaMap, diag.Diagnostics) {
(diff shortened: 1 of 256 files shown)


genesyscloud/resource_genesyscloud_script.go
```diff
@@ -31,6 +31,26 @@ func getAllScripts(ctx context.Context, clientConfig *platformclientv2.Configura
 			break
 		}
 		for _, script := range *scripts.Entities {
+			//We only want scripts that are published.
+			_, resp, err := scriptsAPI.GetScriptsPublishedScriptId(*script.Id, "")
```
carnellj-genesys, line 35: Added code here to check whether a script is published.  We only want to return the published scripts and not unpublished scripts

genesyscloud/resource_genesyscloud_script.go
```diff
@@ -146,6 +166,15 @@ func createScript(ctx context.Context, d *schema.ResourceData, meta interface{})
 		return diag.Errorf("Script %s not found after creation.", scriptName)
 	}
 
+	//Publish the 	script
+	publishScriptBody := &platformclientv2.Publishscriptrequestdata{
```
carnellj-genesys, line 170: When we create a script we publish it

genesyscloud/resource_genesyscloud_script.go
```diff
@@ -146,6 +166,15 @@ func createScript(ctx context.Context, d *schema.ResourceData, meta interface{})
...
+		ScriptId: sdkScripts[0].Id,
+	}
+
+	if _, _, err := scriptsAPI.PostScriptsPublished("0", *publishScriptBody); err != nil {
```
carnellj-genesys, line 174: @charliecon How come there is not update script object.  Does the API always just replace the object

genesyscloud/resource_genesyscloud_script_test.go
```diff
@@ -103,3 +105,39 @@ func testVerifyScriptDestroyed(state *terraform.State) error {
 	// Success. All Scripts destroyed
 	return nil
 }
+
+func validateScriptPublished(scriptResourceName string) resource.TestCheckFunc {
```
carnellj-genesys, line 109: helper function to validate whether the script published after it was done

test/data/resource/genesyscloud_script/test_script.json
```diff
@@ -209,7 +209,8 @@
                                 "value": {}
                             },
                             "value": {
-                                "typeName": "variable"
+                                "typeName": "variable",
+                                "value": "MyCheckBox"
```
carnellj-genesys, line 213: Had to fix the script.  Once I started publishing the script failed with a validation error because the checkbox in the script was missing a value.

dginty4:
Lots of merge conflicts, probably due to Hemanths PR getting merged

PR #575, opened 2023-07-24
DEVENGAGE-2024: Fixed failing test cases

carnellj-genesys:


genesyscloud/outbound/resource_genesyscloud_outbound_init.go
```diff
@@ -25,7 +25,7 @@ func SetRegistrar(l registrar.Registrar) {
 	l.RegisterResource("genesyscloud_outbound_messagingcampaign", ResourceOutboundMessagingCampaign())
 	l.RegisterResource("genesyscloud_outbound_sequence", ResourceOutboundSequence())
 	l.RegisterResource("genesyscloud_outbound_settings", ResourceOutboundSettings())
-	l.RegisterResource("genesyscloud_outbound_wrapupcodemappings", resourceOutboundWrapUpCodeMappings())
+	l.RegisterResource("genesyscloud_outbound_wrapupcodemappings", ResourceOutboundWrapUpCodeMappings())
```
carnellj-genesys, line 28: This was causing one of the issues where the genesyscloud_outbound_wrapupcodemappings resource was not registered so when the export ran, the exporter complained about not being able to find the resource.

genesyscloud/outbound/resource_genesyscloud_outbound_wrapupcodemappings.go
```diff
@@ -45,7 +46,7 @@ func OutboundWrapupCodeMappingsExporter() *resourceExporter.ResourceExporter {
 	}
 }
 
-func resourceOutboundWrapUpCodeMappings() *schema.Resource {
+func ResourceOutboundWrapUpCodeMappings() *schema.Resource {
```
carnellj-genesys, line 49: Had to make this function public

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -8,32 +8,23 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
-	registrar "terraform-provider-genesyscloud/genesyscloud/resource_register"
 )
 
 func SetRegistrar(l registrar.Registrar) {
 	l.RegisterResource("genesyscloud_tf_export", ResourceTfExport())
-	
-}
-
-// func GetRegistrarresources() {
```
carnellj-genesys, line 29: Removed this commented out code.  We should leave commented code in the codebase

genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go
```diff
@@ -43,52 +43,52 @@ type QueueExport struct {
...
+// TestAccResourceTfExport does a basic test check to make sure the export file is created.
+func TestAccResourceTfExport(t *testing.T) {
+	var (
+		exportTestDir   = "../../.terraform" + uuid.NewString()
+		exportResource1 = "test-export1"
+		configPath      = filepath.Join(exportTestDir, defaultTfJSONFile)
+		statePath       = filepath.Join(exportTestDir, defaultTfStateFile)
+	)
 
-// func TestAccResourceTfExport(t *testing.T) {
```
carnellj-genesys, line 47: This test case was commented out.  I saw no reason to comment it out and added it back in.

genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go
```diff
@@ -380,6 +393,12 @@ func TestAccResourceTfExportIncludeFilterResourcesByRegEx(t *testing.T) {
 			[]string{
 				strconv.Quote("genesyscloud_routing_queue::-prod"),
 			},
+			[]string{
```
carnellj-genesys, line 396: This was the cause of one of the bugs where I was getting a nil pointer exception.  The problem is that essentially we were creating the resources we were using in the test at the same time we were executing the export.  This is a race condition because the resources that the exporter test is executing might not be complete before the test cases complete.  

This is why sometimes we do not see the test failing.  So I added an explicit depends_on to each test case to ensure that resources that are needed by the exporter test case are created before the exporter starts.  This code creates a list of resources that needs to be passed into resource creation.

genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go
```diff
@@ -425,7 +444,14 @@ func TestAccResourceTfExportExcludeFilterResourcesByRegEx(t *testing.T) {
 			exportTestDir,
 			trueValue,
 			[]string{
-				strconv.Quote("genesyscloud_routing_queue::-prod"),
```
carnellj-genesys, line 428: This was a bug.  We wanted to test that dev|test items were excluded and prod resources were included.  So I needed to make sure the resources created and the resources being checked matched.

genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go
```diff
@@ -1256,45 +1292,51 @@ func generateTfExportByName(
...
+	logErrors string,
+	dependencies []string) string {
 	return fmt.Sprintf(`resource "genesyscloud_tf_export" "%s" {
 		directory = "%s"
 		include_state_file = %s
 		resource_types = [%s]
 		exclude_attributes = [%s]
 		export_as_hcl = %s
 		log_permission_errors = %s
+		depends_on=[%s]
```
carnellj-genesys, line 1304: Adding the depends_on

genesyscloud/resource_genesyscloud_journey_outcome.go
```diff
@@ -235,12 +236,13 @@ func flattenJourneyOutcome(d *schema.ResourceData, journeyOutcome *platformclien
...
 	resourcedata.SetNillableValue(d, "context", lists.FlattenAsList(journeyOutcome.Context, flattenContext))
 	resourcedata.SetNillableValue(d, "journey", lists.FlattenAsList(journeyOutcome.Journey, flattenJourney))
+
 	resourcedata.SetNillableValue(d, "associated_value_field", lists.FlattenAsList(journeyOutcome.AssociatedValueField, flattenAssociatedValueField))
 }
 
 func flattenAssociatedValueField(associatedValueField *platformclientv2.Associatedvaluefield) map[string]interface{} {
 	associatedValueFieldMap := make(map[string]interface{})
-	associatedValueFieldMap["dataType"] = associatedValueField.DataType
+	associatedValueFieldMap["data_type"] = associatedValueField.DataType
```
carnellj-genesys, line 245: This was another bug that only showed itself when the data condition was setup.  When an associated value field is set the schema value being set (dataType) is not defined in the schema.  It is data_type.  Terraform throws an exception if you try to set a schema value that is not defined in the schema.

dginty4 (approved):

charliecon (approved):

PR #576, opened 2023-07-25
DEVENGAGE-1990 Campaign Rule Sets export issue

dginty4:


genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
```diff
@@ -56,7 +56,7 @@ var (
 			},
 			`actions`: {
 				Description: `The list of actions to be taken if the conditions are true.`,
-				Optional:    true,
+				Required:    true,
```
dginty4, line 59: Although the api accepts rulesets without actions it leads to a bug, the outbound team is fixing the api to make it required but I made it required here to avoid issues now

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_utils.go
```diff
@@ -0,0 +1,269 @@
...
+	return platformclientv2.Ruleset{
+		Name:        &name,
+		ContactList: gcloud.BuildSdkDomainEntityRef(d, "contact_list_id"),
+		Queue:       gcloud.BuildSdkDomainEntityRef(d, "queue_id"),
+		Rules:       buildDialerules(d.Get("rules").([]interface{})),
+	}
+}
+
+// buildDialerules maps a []interface{} into a Genesys Cloud *[]platformclientv2.Dialerrule
+func buildDialerules(rules []interface{}) *[]platformclientv2.Dialerrule {
```
dginty4, line 23: I was thinking would this be a better naming convention for the flatten and build functions. This function used to be called `buildsdkOutboundRulesetDialerruleSlice`. These names get very long and get even longer depending how many nested resources there are. I also think they're misleading since the function is building a general `Dialerrule` it's not specific to the ruleset. Might save us from creating multiple functions to build the same objects too

carnellj-genesys (changes requested):
A couple of other things:

Please include a short description in each file of what it does using:

/*
*/ format

Also you missed calling the registrar method in the main class for the outbound_ruleset.  Without calling this field, the component wont actually register.  Your tests cases will pass because you are registering rulesets in each test case.

carnellj-genesys (changes requested):
A couple of other things:

Please include a short description in each file of what it does using:

/*
*/ format

Also you missed calling the registrar method in the main class for the outbound_ruleset.  Without calling this field, the component wont actually register.  Your tests cases will pass because you are registering rulesets in each test case.

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_utils.go
```diff
@@ -0,0 +1,269 @@
...
+	rulesSlice := make([]platformclientv2.Dialerrule, 0)
+	for _, rule := range rules {
+		var sdkRule platformclientv2.Dialerrule
+		ruleMap := rule.(map[string]interface{})
+
+		resourcedata.BuildSDKStringValueIfNotNil(&sdkRule.Name, ruleMap, "name")
```
carnellj-genesys, line 29: Nice

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_utils.go
carnellj-genesys: I am good with that

genesyscloud/util/resourcedata/resourcedata.go
```diff
@@ -14,12 +15,70 @@ const (
 	DateParseFormat = "2006-01-02"
 )
 
+// Use these functions to read properties from the schema and set it on in a map in build function
+
+// This function will read a map and set a property on an object if the value exists
+func BuildSDKStringValueIfNotNil(field **string, targetMap map[string]interface{}, key string){
```
carnellj-genesys, line 21: Is this correct I have never a double pointer to a string

carnellj-genesys (changes requested):

genesyscloud/util/resourcedata/resourcedata.go
dginty4: It is, I don't know why but it wasn't setting the value when I was trying with one but I could be wrong

carnellj-genesys (approved):

charliecon (approved):

genesyscloud/util/resourcedata/resourcedata.go
```diff
@@ -14,12 +15,97 @@ const (
...
+}
+
+// This function will read the value of a map property and set it in a map
+func SetMapStringMapValueIfNotNil(targetMap map[string]interface{}, key string, valueList *map[string]string) {
+	if valueList != nil {
+		results := make(map[string]interface{})
+		for k, v := range *valueList {
+			results[k] = v
+		}
+		targetMap["properties"] = results
```
charliecon, line 72: Should this be a hardcoded key?

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_utils.go
```diff
@@ -0,0 +1,255 @@
...
+		Rules:       buildDialerules(d.Get("rules").([]interface{})),
+	}
+}
+
+// buildDialerules maps a []interface{} into a Genesys Cloud *[]platformclientv2.Dialerrule
+func buildDialerules(rules []interface{}) *[]platformclientv2.Dialerrule {
+	rulesSlice := make([]platformclientv2.Dialerrule, 0)
+	for _, rule := range rules {
+		var sdkRule platformclientv2.Dialerrule
+		ruleMap := rule.(map[string]interface{})
```
charliecon, line 33: It's better to do a little `ok` check here. If the person using the resource defines an empty rule map like this `rule{}` they'll get a panic error here 
```golang
ruleMap, ok := rule.(map[string]interface{})
if !ok {
  continue
}
```

charliecon: just continuing on in the loop is grand

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_utils.go
```diff
@@ -0,0 +1,255 @@
...
+
+	return &rulesSlice
+}
+
+// buildConditions maps a []interface{} into a Genesys Cloud *[]platformclientv2.Condition
+func buildConditions(conditions []interface{}) *[]platformclientv2.Condition {
+	conditionSlice := make([]platformclientv2.Condition, 0)
+	for _, conditions := range conditions {
+		var sdkCondition platformclientv2.Condition
+		conditionMap := conditions.(map[string]interface{})
```
charliecon, line 52: same here

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_utils.go
```diff
@@ -0,0 +1,255 @@
...
+
+	return &conditionSlice
+}
+
+// buildDialeractions maps a []interface{} into a Genesys Cloud *[]platformclientv2.Dialeraction
+func buildDialeractions(actions []interface{}) *[]platformclientv2.Dialeraction {
+	actionsSlice := make([]platformclientv2.Dialeraction, 0)
+	for _, action := range actions {
+		var sdkAction platformclientv2.Dialeraction
+		actionMap := action.(map[string]interface{})
```
charliecon, line 82: same here

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_utils.go
```diff
@@ -0,0 +1,255 @@
...
+
+	return &fieldmappingsSlice
+}
+
+// buildDataactionconditionpredicates maps a []interface{} into a Genesys Cloud *[]platformclientv2.Dataactionconditionpredicate
+func buildDataactionconditionpredicates(predicates []interface{}) *[]platformclientv2.Dataactionconditionpredicate {
+	predicatesSlice := make([]platformclientv2.Dataactionconditionpredicate, 0)
+	for _, predicate := range predicates {
+		var sdkPredicate platformclientv2.Dataactionconditionpredicate
+		predicateMap := predicate.(map[string]interface{})
```
charliecon, line 121: same here

PR #578, opened 2023-07-26
[no-jira] Small refactoring towards generics

charliecon:
This change doesn't actually reduce the lines of code *yet*. Just getting familiar with generics

### DIFF
--- a/genesyscloud/util/lists/util_lists.go
+++ b/genesyscloud/util/lists/util_lists.go
@@ -6,7 +6,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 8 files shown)


genesyscloud/util/lists/util_lists.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func StringInSlice(a string, list []string) bool {
+func ItemInSlice[T comparable](a T, list []T) bool {
```
carnellj-genesys, line 9: Nice.  Generics are very cool for solving boilerplate code.  

You can even use them in interfaces

carnellj-genesys (approved):

PR #582, opened 2023-07-27
Refactoring removeZeroValues func and adding unit test

charliecon:


charliecon:
@carnellj This is the work I did with reflection, just if you want to have a look. This solution works 

In the end it probably wasn't worth the time spent 😄

(FYI The behaviour for for lists and maps is that when they're initialised with the `make` function, `IsZero() == false`, so the empty array/map is left in the configMap. I don't see this really mattering one way or the other)

PR #584, opened 2023-07-31
No jira fix Missing Exporters Register

HemanthDogiparthi12:


carnellj-genesys (approved):

charliecon:
@HemanthDogiparthi12 could you just run `go fmt` against the files that have some inconsistencies with the formatting? After that, if the matrix tests are passing and the exporters are getting registered okay, we can go ahead and merge

HemanthDogiparthi12:
> @HemanthDogiparthi12 could you just run `go fmt` against the files that have some inconsistencies with the formatting? After that, if the matrix tests are passing and the exporters are getting registered okay, we can go ahead and merge

We can proceed to merge this. The exports tests passed on the run.

PR #587, opened 2023-08-02
Fixing the export of web deployment configurations

charliecon:


genesyscloud/resource_genesyscloud_webdeployments_configuration.go
```diff
@@ -760,11 +761,11 @@ func determineLatestVersion(ctx context.Context, api *platformclientv2.WebDeploy
 		}
 
 		if maxVersion == 0 {
-			version = "draft"
+			version = draft
+		} else {
+			version = strconv.Itoa(maxVersion)
```
charliecon, line 766: This is the main change. This line wasn't wrapped in the `else` before. So version was being set to `strconv.Itoa(maxVersion)` whether it was 0 or not. This resulted in a 404 because there probably isn't a version "0" of this configuration

dginty4 (approved):

PR #589, opened 2023-08-03
No jira added logging

carnellj-genesys:


genesyscloud/provider.go
```diff
@@ -282,7 +280,7 @@ func initClientConfig(data *schema.ResourceData, version string, config *platfor
 		},
 		ResponseLogHook: func(response *http.Response) {
 			if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
-				log.Printf("Response %s", response.Status)
+				log.Printf("Response %s for request:%s %s", response.Status, response.Request.Method, response.Request.URL)
```
carnellj-genesys, line 283: Added the url in the response log so if we get a non-200 response we no longer get a little more info.

genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go
```diff
@@ -445,6 +445,10 @@ func TestAccResourceTfExportExcludeFilterResourcesByRegEx(t *testing.T) {
 			trueValue,
 			[]string{
 				strconv.Quote("genesyscloud_routing_queue::-(dev|test)$"),
+				strconv.Quote("genesyscloud_outbound_ruleset"),
```
carnellj-genesys, line 448: Excluding these items because I realized they added significantly more time to the export because in all are other tests we tell the platform what to include.  Here we tell the platform what to exclude.  We have a lot of data in the inintestingorg and exporting all this data just to prove out the test case is unnecessary

HemanthDogiparthi12 (approved):

PR #595, opened 2023-08-10
DEVENGAGE-2040 Change get all external contatcs to use scan endpoint

dginty4:


genesyscloud/external_contacts/genesyscloud_externalcontacts_contact_proxy.go
```diff
@@ -111,37 +111,14 @@ func (p *externalContactsContactsProxy) updateExternalContact(ctx context.Contex
...
-
-		if externalContacts.Entities == nil || len(*externalContacts.Entities) == 0 {
-			break
-		}
-
-		for _, externalContact := range *externalContacts.Entities {
-			log.Printf("Dealing with external contact id : %s", *externalContact.Id)
-			allExternalContacts = append(allExternalContacts, externalContact)
-		}
+	externalContacts, _, err := p.externalContactsApi.GetExternalcontactsScanContacts(100, "")
```
charliecon, line 114: Would this only be getting the first 100 contacts ?

dginty4: I don't think so. This endpoint is a little different and I believe this means it will retrieve the contacts in batches of 100. I could have misinterpreted so feel free to check. https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-externalcontacts-scan-contacts

dginty4: This is working properly now

genesyscloud/external_contacts/genesyscloud_externalcontacts_contact_proxy.go
```diff
@@ -142,6 +126,12 @@ func getAllExternalContactsFn(ctx context.Context, p *externalContactsContactsPr
 			log.Printf("Dealing with external contact id : %s", *externalContact.Id)
 			allExternalContacts = append(allExternalContacts, externalContact)
 		}
+
+		log.Println(*externalContacts)
+		if externalContacts.Cursors.After == nil {
```
charliecon, line 131: I would say `if externalContacts.Cursors == nil || externalContacts.Cursors.After == nil {` 

I think if you jump to the nested field when the outer one doesn't exist in the first place, you can get an error

genesyscloud/external_contacts/genesyscloud_externalcontacts_contact_proxy.go
```diff
@@ -142,6 +126,12 @@ func getAllExternalContactsFn(ctx context.Context, p *externalContactsContactsPr
 			log.Printf("Dealing with external contact id : %s", *externalContact.Id)
 			allExternalContacts = append(allExternalContacts, externalContact)
 		}
+
+		log.Println(*externalContacts)
```
charliecon, line 130: You can remove this log statement. I could see it clogging things up a bit

genesyscloud/external_contacts/genesyscloud_externalcontacts_contact_proxy.go
```diff
@@ -142,6 +126,12 @@ func getAllExternalContactsFn(ctx context.Context, p *externalContactsContactsPr
 			log.Printf("Dealing with external contact id : %s", *externalContact.Id)
```
charliecon, line 126: Same with this one

charliecon (approved):

PR #598, opened 2023-08-15
DEVENGAGE-2073 add home screen to messenger settings

MateuszKikmunter:
Added home screen to messenger settings.
This will allow to create webdeployments with voice channel selected only (for co-browse) - othwerwise it fails on validation. (Squonk)


MateuszKikmunter:
> Looks good. You just need to run `go generate` from the root of the project

I ran the command and it made changes to the docs file, should I include it in the PR too?

charliecon:
> > Looks good. You just need to run `go generate` from the root of the project
> 
> I ran the command and it made changes to the docs file, should I include it in the PR too?

Yep!

charliecon (approved):

carnellj-genesys (approved):

PR #600, opened 2023-08-17
Avoiding UpdateContext for flows resource

charliecon:
There was an issue in the flows resource where a user would:
- Successfully deploy a flow
- Change something in the yaml file and make it invalid
- Re-run `terraform apply` , the apply would fail
- Re-run `terraform apply` again, no changes would be detected in the file (because there were none) and `No changes. Apply complete.` would display in green

I noticed this didn't happen with the scripts resource because all fields were `ForceNew` and no UpdateContext was used, only Create. This means the existing script would be destroyed before trying to re-upload with an invalid script. Running `terraform apply` after this would mean an attempt at a create, not an update. 

On the other hand, the flows resource had a create and update function, but the create always just ran right into the update, so it wasn't necessary.

This PR changes the flows resource to behave like the scripts resource.

### DIFF
--- a/genesyscloud/resource_genesyscloud_flow.go
+++ b/genesyscloud/resource_genesyscloud_flow.go
@@ -12,7 +12,7 @@ import (
 	"time"
 
 	resourceExporter "terraform-provider-genesyscloud/genesyscloud/resource_exporter"
-	files "terraform-provider-genesyscloud/genesyscloud/util/files"
+	"terraform-provider-genesyscloud/genesyscloud/util/files"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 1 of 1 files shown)


genesyscloud/resource_genesyscloud_flow.go
```diff
@@ -74,32 +73,31 @@ func ResourceFlow() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ValidateFunc: ValidatePath,
+				ForceNew:     true,
 			},
 			"file_content_hash": {
 				Description: "Hash value of the YAML file content. Used to detect changes.",
 				Type:        schema.TypeString,
 				Required:    true,
+				ForceNew:    true,
```
carnellj-genesys, line 82: So trying to understand this.  We added force new to everything.  By adding a forceNew does this cause the file_content_hash to always be reset even if the deploy fails?

carnellj-genesys (approved):

genesyscloud/resource_genesyscloud_flow.go
charliecon: Yes. Because it's all ForceNew, when the user tries to update something terraform will tear down the existing flow, wiping the state file, then fail a create. If the user re-runs `terraform apply` it will try again at a create.

PR #603, opened 2023-08-23
[no-jira] Change to flow resource docs

charliecon:


genesyscloud/resource_genesyscloud_flow.go
```diff
@@ -58,7 +58,7 @@ func FlowExporter() *resourceExporter.ResourceExporter {
 
 func ResourceFlow() *schema.Resource {
 	return &schema.Resource{
-		Description: `Genesys Cloud Flow`,
+		Description: `Genesys Cloud Architect Flow. Note: Changing any field in this resource will cause the current flow to be dropped and recreated with a new ID.`,
```
dginty4, line 61: Is the old flow deleted? If not is that worth mentioning?

charliecon: @dginty4 Yeah it is deleted. That's what "dropped" is supposed to mean in this case

dginty4: Ah ok, sorry I thought it was said in standup it wasn't deleted, my bad

dginty4 (approved):

genesyscloud/resource_genesyscloud_flow.go
charliecon: Oh yeah that bug was fixed in another PR that's been merged already

PR #617, opened 2023-09-06
DEVENGAGE-2106 Fix regex applying to all resource exporters

PrinceMerluza:
Add a check to filter functions so that the regex will only be applied to the correct resource.

EDIT: Also added test cases for the scenarios

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -25,11 +25,11 @@ unittests:
 
 
 coverage:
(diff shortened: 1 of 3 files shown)


charliecon (approved):

carnellj-genesys (approved):
Nice work @PrinceMerluza   Merging

PR #625, opened 2023-09-13
DEVENGAGE-2022 Allow exported resources to be split into multiple files

PrinceMerluza:
Added `split_files_by_resource` property for resource `genesys_cloud_tf_export`.
Default is `false` and does the current behavior of exporting a giant file. 

If set to `true`, files will be split by resource type eg. genesyscloud_routing_queue.tf(HCL export) / genesyscloud_routing_queue.tf.json (JSON export)
Terraform provider will also be in its own file: provider.tf / provider.tf.json
If there are unresolved attribute, variables will also be declared in separate file: variables.tf / variables.tf.json

Test cases added.

Change on state: Instead of checking for the default export file in the resource read, it will check the entire directory for any files. destroy will also mean deleting every file(and subdirs) inside the export directory.

### DIFF
--- a/genesyscloud/tfexporter/export_common.go
+++ b/genesyscloud/tfexporter/export_common.go
@@ -2,6 +2,7 @@ package tfexporter
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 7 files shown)


carnellj-genesys (approved):
This looks really good

PR #632, opened 2023-09-19
DEVENGAGE-2131: Refactor integration and data actions

PrinceMerluza:
Refactor `integration` resource
Refactor `integration_credential` resource
Refactor `integration_action` resource

Also refactored `recording_media_retention_policy` resource. The reason is that its test case requires an `integration` resource.
Not sure if there's a better way of getting it without circular dependency between 'gcloud' and the new packages so I went ahead and refactored the `recording_media_retention_policy` resource as well.

Also had to make some utility functions from different resources `public` so the new packages can use them. 
Methods from util_json and some resource specific test resource builders. 

Test cases pass for the integration packages and modified gcloud resources.

Failing test case on `recording_media_retention_policy` due to user roles. This test case is also not passing in current version of terraform provider so the reason is not the refactoring. Ignoring for this one, since that may need a different issue to fix.


PrinceMerluza:
> Looks awesome prince

Thanks John :)

PrinceMerluza:
+ Applied the fix from this PR as well https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/634

carnellj-genesys:
@charliecon Can you please do a quick review and then merge if everything looks good.

PrinceMerluza:
Resolved merge conflicts

genesyscloud/integration_action/resource_genesyscloud_integration_action.go
```diff
@@ -0,0 +1,275 @@
...
+			d.Set("secure", *action.Secure)
+		} else {
+			d.Set("secure", nil)
+		}
+
+		if action.Config.TimeoutSeconds != nil {
+			d.Set("config_timeout_seconds", *action.Config.TimeoutSeconds)
+		} else {
+			d.Set("config_timeout_seconds", nil)
+		}
```
charliecon, line 168: For these operations, I recommend using the function SetNillableValue in the resourcedata package. It uses generics to perform all the `if T is nil { d.Set(key, *T) } else { d.Set(key, nil) }` logic in one function. 

Here is an example implementation: [https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/routing_sms_addresses/resource_genesyscloud_routing_sms_addresses.go#L109](https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/routing_sms_addresses/resource_genesyscloud_routing_sms_addresses.go#L109)

genesyscloud/integration_action/resource_genesyscloud_integration_action_utils.go
```diff
@@ -0,0 +1,186 @@
...
+	responseMap := make(map[string]interface{})
+	if sdkResponse.TranslationMap != nil {
+		responseMap["translation_map"] = *sdkResponse.TranslationMap
+	}
+	if sdkResponse.TranslationMapDefaults != nil {
+		responseMap["translation_map_defaults"] = *sdkResponse.TranslationMapDefaults
+	}
+	if sdkResponse.SuccessTemplate != nil {
+		responseMap["success_template"] = *sdkResponse.SuccessTemplate
+	}
```
charliecon, line 184: There is also a function in the resourcedata package for this logic: `SetMapValueIfNotNil`

genesyscloud/recording_media_retention_policy/resource_genesyscloud_recording_media_retention_policy.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+		if retentionPolicy.Conditions != nil {
+			d.Set("conditions", flattenConditions(retentionPolicy.Conditions))
+		}
+		if retentionPolicy.Actions != nil {
+			d.Set("actions", flattenPolicyActions(retentionPolicy.Actions, pp, ctx))
+		}
+		if retentionPolicy.PolicyErrors != nil {
+			d.Set("policy_errors", flattenPolicyErrors(retentionPolicy.PolicyErrors))
+		}
```
charliecon, line 139: Same here with the resourcedata functions. To set a key with an interface array value returned from a function, you can use `SetNillableValueWithInterfaceArrayWithFunc`

PrinceMerluza: Used resourcedata util functions where applicable

genesyscloud/integration_action/resource_genesyscloud_integration_action_utils.go
PrinceMerluza: Used resourcedata util functions where applicable

genesyscloud/integration_action/resource_genesyscloud_integration_action.go
PrinceMerluza: Used resourcedata util functions where applicable

charliecon (approved):

PR #633, opened 2023-09-20
bug/DEVENGAGE-2138 Edge Phones export failing

dginty4:


genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -530,7 +530,7 @@ func getAllPhones(_ context.Context, sdkConfig *platformclientv2.Configuration)
 
 	edgesAPI := platformclientv2.NewTelephonyProvidersEdgeApiWithConfig(sdkConfig)
 
-	for pageNum := 1; ; pageNum++ {
+	for pageNum := 1; pageNum <= 100; pageNum++ {
```
carnellj-genesys, line 533: This is not going to work because you will only get the first 100 pages.  If there are more than 100 pages  you will prevent the original error but introduce a new bug (you can have up to 10000 records or 1000 pages).

You need to call the API first and then get the pageCount from the first call.  You then make all of the succeeding calls up to the pageCount. That is the only way you can know for sure exactly what the issue.

My name is John Carnell and I disapprove of this PR.

carnellj-genesys (changes requested):

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
dginty4: Since the page size is 100 and there can be up 10,000 records doesn't that mean there can't be more 100 pages? 1,000 pages with a page size of 100 would mean 100,000 records?

carnellj-genesys: We are future-proofing :).  By calling the API first, getting the page count, and then cycling on the page count we isolate ourselves from having to make a change if the upper limit ever goes up.  As a general rule.  We should be paginating based on the page count.  I am looking through the code-base and while our iterations technically work, they leave us up to un an unbounded loop if one of our assumptions is off.

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -530,9 +530,18 @@ func getAllPhones(_ context.Context, sdkConfig *platformclientv2.Configuration)
...
-	for pageNum := 1; ; pageNum++ {
+	phones, _, _ := edgesAPI.GetTelephonyProvidersEdgesPhones(1, 100, "", "", "", "", "", "", "", "", "", "", "", "", "", nil, nil)
+
+	for pageNum := 1; pageNum <= *phones.PageCount; pageNum++ {
 		const pageSize = 100
-		phones, _, getErr := edgesAPI.GetTelephonyProvidersEdgesPhones(pageNum, pageSize, "", "", "", "", "", "", "", "", "", "", "", "", "", nil, nil)
+		phones, resp, getErr := edgesAPI.GetTelephonyProvidersEdgesPhones(pageNum, pageSize, "", "", "", "", "", "", "", "", "", "", "", "", "", nil, nil)
+
+		// This endpoint will throw a 400 if the page number limit is reached. The limit can be less that the total number of pages
+		if resp != nil && resp.StatusCode == 400 {
```
dginty4, line 540: @carnellj-genesys I've had to check for a 400 response since the `PageCount` can be higher than 100 which causes the bug

dginty4: @carnellj-genesys this should work now

carnellj-genesys (approved):

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -530,11 +530,18 @@ func getAllPhones(_ context.Context, sdkConfig *platformclientv2.Configuration)
...
+	for _, phone := range *phones.Entities {
+		if phone.State != nil && *phone.State != "deleted" {
+			resources[*phone.Id] = &resourceExporter.ResourceMeta{Name: *phone.Name}
+		}
+	}
+
+	for pageNum := 2; pageNum <= *phones.PageCount; pageNum++ {
 		const pageSize = 100
-		// unlocks result limit
 		const sortBy = "id"
```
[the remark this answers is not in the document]
carnellj-genesys: Ship it

PR #634, opened 2023-09-22
DEVENGAGE-2148 Fix export issue with media retention policies with deleteDaysThreshold <= 365

PrinceMerluza:
Fix issue by replacing the API call to use deleteDaysThreshold  = 0 instead of 365.

### DIFF
--- a/genesyscloud/resource_genesyscloud_recording_media_retention_policy.go
+++ b/genesyscloud/resource_genesyscloud_recording_media_retention_policy.go
@@ -972,7 +972,7 @@ func getAllMediaRetentionPolicies(_ context.Context, clientConfig *platformclien
 
 	for pageNum := 1; ; pageNum++ {
 		const pageSize = 100
-		retentionPolicies, _, getErr := recordingAPI.GetRecordingMediaretentionpolicies(pageSize, pageNum, "", []string{}, "", "", "", true, false, false, 365)
+		retentionPolicies, _, getErr := recordingAPI.GetRecordingMediaretentionpolicies(pageSize, pageNum, "", []string{}, "", "", "", true, false, false, 0)
 
 		if getErr != nil {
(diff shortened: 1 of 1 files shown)


genesyscloud/resource_genesyscloud_recording_media_retention_policy.go
```diff
@@ -972,7 +972,7 @@ func getAllMediaRetentionPolicies(_ context.Context, clientConfig *platformclien
 
 	for pageNum := 1; ; pageNum++ {
 		const pageSize = 100
-		retentionPolicies, _, getErr := recordingAPI.GetRecordingMediaretentionpolicies(pageSize, pageNum, "", []string{}, "", "", "", true, false, false, 365)
+		retentionPolicies, _, getErr := recordingAPI.GetRecordingMediaretentionpolicies(pageSize, pageNum, "", []string{}, "", "", "", true, false, false, 0)
```
carnellj-genesys, line 975: Hey @PrinceMerluza ,

Why did you chose a value of 0 here.  I could not find any documentation for this API about setting this value to zero so I just wanted to see you reasoning behind setting this to zero.

carnellj-genesys (commented):
I only had a quick comment.  If you can give me some feedback on it, I would appreciate it.

genesyscloud/resource_genesyscloud_recording_media_retention_policy.go
PrinceMerluza: Hi John, so technically the correct API call really would be setting this to 'null/nil' and not including `deleteDaysThreshold` as a query parameter in the actual request. However this API method only allows `int` and so I tested if the behavior would be the same.

And upon testing setting it to `0` is the same as ignoring it since it will include all policies that have `deleteDaysThreshold  > 0` ie. include all policies since `deleteDaysThreshold` is required to be a positive integer.

carnellj-genesys (approved):

PR #636, opened 2023-09-26
Feat/devengage 1991 - Architect Grammar Resource

dginty4:


genesyscloud/architect_grammar/genesyscloud_architect_grammar_proxy.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	var grammar platformclientv2.Grammar
+	for _, grammarSdk := range *grammars.Entities {
+		if *grammarSdk.Name == name {
+			log.Printf("Retrieved the grammar id %s by name %s", *grammarSdk.Id, name)
+			grammar = grammarSdk
+			return *grammar.Id, false, nil
+		}
+	}
```
charliecon, line 194: You could reuse the `getAllArchitectGrammarFn` function here to get all the grammars. You'd just have to add name as a parameter to that function. Then for the exporter `name` can just be `""`

genesyscloud/architect_grammar/genesyscloud_architect_grammar_proxy.go
```diff
@@ -0,0 +1,283 @@
...
+	_, _, err := p.architectApi.GetArchitectGrammarLanguage(grammarId, languageCode)
+	if err != nil {
+		log.Printf("Laguage %s not found, creating language", languageCode)
+		language, err := p.createArchitectGrammarLanguage(ctx, grammarId, language)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to create grammar language: %s", err)
+		}
+
+		return language, nil
+	}
```
charliecon, line 221: Is there a specific scenario when this would be needed? Seems like terraform would always be calling the create function for you before an update

genesyscloud/architect_grammar/resource_genesyscloud_architect_grammar.go
```diff
@@ -0,0 +1,189 @@
...
+	proxy := getArchitectGrammarProxy(clientConfig)
+	resources := make(resourceExporter.ResourceIDMetaMap)
+
+	grammars, err := proxy.getAllArchitectGrammar(ctx)
+	if err != nil {
+		return nil, diag.Errorf("Failed to get grammars: %v", err)
+	}
+
+	for _, grammar := range *grammars {
+		log.Printf("Dealing with grammar id : %s", *grammar.Id)
```
charliecon, line 32: I wouldn't bother with this log statement. It can complicate things a bit when you're exporting a lot of objects

genesyscloud/architect_grammar/resource_genesyscloud_architect_grammar.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		//cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, ResourceArchitectGrammar())
+
+		resourcedata.SetNillableValue(d, "name", grammar.Name)
+		resourcedata.SetNillableValue(d, "description", grammar.Description)
+		resourcedata.SetNillableValueWithInterfaceArrayWithFuncWithState(d, "languages", grammar.Languages, flattenGrammarLanguages)
+		
+		log.Printf("Read Architect Grammar %s %s", d.Id(), *grammar.Name)
+		//return cc.CheckState()
```
charliecon, line 112: Consistency checker commented out

genesyscloud/architect_grammar/resource_genesyscloud_architect_grammar.go
```diff
@@ -0,0 +1,189 @@
...
+
+	// Update grammar
+	log.Printf("Updating Architect Grammar %s", *architectGrammar.Name)
+	grammar, err := proxy.updateArchitectGrammar(ctx, d.Id(), &architectGrammar)
+	if err != nil {
+		return diag.Errorf("Failed to update grammar: %s", err)
+	}
+
+	// Update each language associated with the grammar
+	for _, language := range *architectGrammar.Languages {
```
charliecon, line 132: Might be better to move all this logic inside the `for` statement over to the proxy file. So you would just call the update function here, and the proxy would take care of everything else. Same for the create

genesyscloud/architect_grammar/resource_genesyscloud_architect_grammar_utils.go
```diff
@@ -0,0 +1,231 @@
...
+		if language.DtmfFileMetadata != nil && language.DtmfFileUrl != nil {
+			fileType := ""
+			if language.DtmfFileMetadata.FileType != nil {
+				fileType = strings.ToLower(*language.DtmfFileMetadata.FileType)
+			}
+			dtmfFileName := fmt.Sprintf("%s-dtmf-%s.%s", *language.Language, grammarId, fileType)
+			if err := files.DownloadExportFile(fullPath, dtmfFileName, *language.DtmfFileUrl); err != nil {
+				return err
+			}
+		}
```
charliecon, line 184: These two if blocks are repetitive. Could make a function out of it. Generics might come in handy

genesyscloud/architect_grammar/genesyscloud_architect_grammar_proxy.go
dginty4: Ya without this I was getting an error when I try to add a new language to a grammar after it is created. Since grammar languages use their own endpoints the language has to exist to update it

charliecon: I wonder if it would be better to have grammar languages as another resource? I see there's a Create, Read, Update and Delete function in the SDK anyway.

dginty4: I did consider this at the start but thought it would be unnecessary. It could be needed though because it's not actually possible to delete a language once it's added since it is it's own object. User prompts behaves the same

dginty4:
I am refactoring this into two resources

genesyscloud/architect_grammar_language/resource_genesyscloud_architect_grammar_language.go
```diff
@@ -0,0 +1,144 @@
...
+	proxy := getArchitectGrammarLanguageProxy(clientConfig)
+	resources := make(resourceExporter.ResourceIDMetaMap)
+
+	languages, err := proxy.getAllArchitectGrammarLanguage(ctx)
+	if err != nil {
+		return nil, diag.Errorf("Failed to get grammar languages: %v", err)
+	}
+
+	for _, language := range *languages {
+		languageId := *language.GrammarId + ":" + *language.Language
```
carnellj-genesys, line 34: Just out of curiosity why are we appending the two strings with a :

genesyscloud/architect_grammar_language/resource_genesyscloud_architect_grammar_language_utils.go
```diff
@@ -0,0 +1,163 @@
...
+	if dtmfFileData, ok := configMap["dtmf_file_data"].([]interface{}); ok {
+		setExporterFileData(dtmfFileData, grammarId, language, subDir, "dtmf")
+	}
+}
+
+func setExporterFileData(fileDataMap []interface{}, grammarId string, language platformclientv2.Grammarlanguage, subDir string, fileType string) {
+	//Set file name and content hash in the exporter map
+	if fileData, ok := fileDataMap[0].(map[string]interface{}); ok {
+		fileExtension := ""
+		if fileType == "voice" && language.VoiceFileMetadata.FileType != nil {
```
carnellj-genesys, line 150: I see magic strings for voice and dtmf in a couple of different spots.  I would extract these out into constant values and reference the constants instead of the direct strings.

genesyscloud/architect_grammar_language/resource_genesyscloud_architect_grammar_language.go
dginty4: The language id is in the format `<grammarId>:<languageCode>` and should be returned by the api but there's a bug where the language id is never being returned by the apis so I have to manually create the id by combining the `grammarId` and `languageCode` properties from the api responses. The architect team have are aware of the bug and when it is fixed we can remove these line and just use the `id` property from the api responses.

docs/guides/custom_auth_action.md
carnellj-genesys, line 1: You might want to merge with master.  Prince Merzula just wrote this guide and it should be included

genesyscloud/architect_grammar/genesyscloud_architect_grammar_proxy.go
carnellj-genesys: 👍

carnellj-genesys (approved):

PR #642, opened 2023-09-28
DEVENGAGE-2020: Moving IVRs, DID pools, DIDs into their own package

charliecon:
Creating packages `telephony_providers_edges_did_pool`, `telephony_providers_edges_did` and `architect_ivr`


genesyscloud/outbound/data_source_genesyscloud_outbound_messagingcampaign_test.go
```diff
@@ -84,10 +83,9 @@ func TestAccDataSourceOutboundMessagingCampaign(t *testing.T) {
 		)
 	)
 
-	config := platformclientv2.GetDefaultConfiguration()
-	err := config.AuthorizeClientCredentials(os.Getenv("GENESYSCLOUD_OAUTHCLIENT_ID"), os.Getenv("GENESYSCLOUD_OAUTHCLIENT_SECRET"))
+	config, err := gcloud.AuthorizeSdk()
```
charliecon, line 86: I made this change in a few different files. In a lot of places this `AuthorizeClientCredentials` function was being called instead of using the function `AuthorizeSdk`, which handles all the logic of grabbing the client id/secret env variables.

genesyscloud/resource_genesyscloud_architect_emergencygroup_test.go
```diff
@@ -22,27 +22,30 @@ func TestAccResourceArchitectEmergencyGroups(t *testing.T) {
...
+	if err != nil {
+		t.Skip("failed to authorize client credentials")
+	}
+
+	// TODO: Create the IVR inside the test config once emergency group has been moved to its own package.
+	// Currently, the ivr resource cannot be registered for these tests because of a cyclic dependency issue.
+	ivrId := "f94e084e-40eb-470b-80d6-0f99cf22d102"
+	if !ivrExists(config, ivrId) {
+		t.Skip("Skipping because IVR does not exists in the target org.")
+	}
```
charliecon, line 41: The architect IVR resource was being registered for this emergency group test. That cannot be done now because a cyclic dependency would exist between the `genesyscloud` and `architect_ivr` package. As a temporary workaround, I left an IVR in the inintesting org and hardcoded the ID here. I could create the IVR using the go sdk but I would also have to create an associated flow, which would be more complex.

genesyscloud/resource_genesyscloud_telephony_providers_edges_phone_test.go
```diff
@@ -182,44 +181,24 @@ func TestAccResourcePhoneBasic(t *testing.T) {
...
+	// TODO: Use did pool resource inside config once cyclic dependency issue is resolved between genesyscloud and did_pools package
+	didPoolId, err := createDidPoolForEdgesPhoneTest(platformConfig, number)
+	if err != nil {
 		t.Fatal(err)
 	}
+	defer func() {
+		if err := deleteDidPool(platformConfig, didPoolId); err != nil {
+			t.Logf("failed to delete did pool '%s': %v", didPoolId, err)
+		}
+	}()
```
charliecon, line 200: Same situation here as with the emergency group tests. DID Pool resource can't be registered for these tests, so these functions create and delete the DID pool

genesyscloud/architect_ivr/genesyscloud_architect_ivr_proxy.go
```diff
@@ -0,0 +1,290 @@
...
+
+// deleteArchitectIvrFn is an implementation function for deleting a Genesys Cloud Architect IVR
+func deleteArchitectIvrFn(_ context.Context, a *architectIvrProxy, id string) (*platformclientv2.APIResponse, error) {
+	return a.api.DeleteArchitectIvr(id)
+}
+
+// getAllArchitectIvrsFn is an implementation function for retrieving all Genesys Cloud Architect IVRs
+func getAllArchitectIvrsFn(_ context.Context, a *architectIvrProxy, name string) (*[]platformclientv2.Ivr, error) {
+	var allIvrs []platformclientv2.Ivr
+	for pageNum := 1; ; pageNum++ {
```
carnellj-genesys, line 168: We need to change this pattern.  While almost all APIs return a 200 if you try to call something out of bounds some do not.  We should call the api first, store the first page results and then get the pageCount and call up to the pageCount

carnellj-genesys (approved):
Two small comments.  Otherwises looks good.  Let me know when you make the changes and I will approve.  The refactor is moving along!!!!

carnellj-genesys (approved):
Per my earlier comment.  Please make the two small changes.  let me know and then I will approve.

genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
```diff
@@ -0,0 +1,240 @@
...
+			return nil
+		}
+
+		cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, ResourceArchitectIvrConfig())
+		d.Set("name", *ivrConfig.Name)
+		d.Set("dnis", lists.StringListToSetOrNil(ivrConfig.Dnis))
+
+		resourcedata.SetNillableValue(d, "description", ivrConfig.Description)
+
+		if ivrConfig.OpenHoursFlow != nil {
```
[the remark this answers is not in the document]
charliecon: @carnellj-genesys They are different in that we're checkign if the outer struct `OpenHoursFlow` is not nil, and then setting `OpenHoursFlow.Id` as the value. I could create a new function in the resourcedata package for this scenario?

genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
```diff
@@ -0,0 +1,240 @@
...
+// createIvrConfig is used by the resource to create a Genesys Cloud Architect IVR
+func createIvrConfig(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	name := d.Get("name").(string)
+	description := d.Get("description").(string)
+	openHoursFlowId := gcloud.BuildSdkDomainEntityRef(d, "open_hours_flow_id")
+	closedHoursFlowId := gcloud.BuildSdkDomainEntityRef(d, "closed_hours_flow_id")
+	holidayHoursFlowId := gcloud.BuildSdkDomainEntityRef(d, "holiday_hours_flow_id")
+	scheduleGroupId := gcloud.BuildSdkDomainEntityRef(d, "schedule_group_id")
+	divisionId := d.Get("division_id").(string)
+	dnis := lists.BuildSdkStringList(d, "dnis")
```
dginty4, line 49: A `buildIvrConfigFromResourceData` function might help duplicating a lot of this in the update function

genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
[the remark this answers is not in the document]
charliecon: My bad. Just realised I should be using `SetNillableReference. Updating now

dginty4 (approved):

charliecon:
> Two small comments. Otherwises looks good. Let me know when you make the changes and I will approve. The refactor is moving along!!!!

@carnellj-genesys All changes have been made ✅

PR #652, opened 2023-10-04
Feat/cleanup direct routing queue settings

jonesb-genesys:
There has been a change to some objects on the Queue surrounding Direct Routing. The public-api PR merged https://bitbucket.org/inindca/public-api-v2/pull-requests/11702 but will probably take some time to make it to the SDK. I am going to be OOO from 10/6 to 10/15. I wanted to make sure these fixes/test fixes were ready in a PR in the event the changes make it to the SDK and tests fail while I am OOO.

Thanks!
Brendan


carnellj-genesys (approved):

charliecon (approved):

charliecon:
Hi @jonesb-genesys 

Don't forget you'll have to run `go generate` from the root of the project to update the docs. Before doing that, you might have to update the example resource inside `examples/resources/genesyscloud_routing_queue`, if the example is outdated.

Also, you'll need to put up a separate PR to update the platform go sdk version in the project to the latest version that includes these new fields. As of right now, the latest version is v113 and it does not contain the changes. v112 is the version currently on main. (Before doing that, I recommend checking that no open PRs are updating the version already, just to potentially save time. Someone might already be doing it)

The test `TestAccResourceRoutingQueueDirectRouting` must be failing as it stands with these changes. After the platform sdk version has been updated and the changes are in, you can run `go test -v -run TestAccResourceRoutingQueueDirectRouting` to confirm it's passing and nothing has broken. 

When all is done, let us know here and we'll remove the "do not merge" tag from the title. 

Thanks for the contribution!

-Charlie

jonesb-genesys:
Updated to v115, though am getting this error when trying to run tests
`go test -v -run TestAccResourceRoutingQueueDirectRouting               
2023/10/17 10:17:55 POST https://login./oauth/token giving up after 1 attempt(s): Post "https://login./oauth/token": dial tcp: lookup login.: no such host
exit status 1
FAIL    terraform-provider-genesyscloud/genesyscloud    0.406s`

Not sure if something is wrong locally or was merged in from master but cannot confirm the test passes at this time since I am not exactly sure the origin of this error

jonesb-genesys:
Ok John was really helpful and we got my local testing working and the test is confirmed passing. Thanks!

PR #653, opened 2023-10-05
[DEVENGAGE-2189] Only add a random suffix on export to resources that have been sanitized with the same matching name pattern

bbbco:
As I mentioned in the https://inindca.atlassian.net/browse/DEVENGAGE-2189 JIRA, adding a random string of digits as a suffix to _all_ resource names that have been sanitized feels very heavy handed. We certainly need to handle the edge case to prevent a name collision on export, but this should only happen if there is the risk of an actual name collision. 

This PR adds support for detecting such a name collision and using the same approach to add the string of numbers based on the hash of the original name, or otherwise leaving the name alone after being sanitized.

I also added a test case to check for these various cases.

### DIFF
--- a/genesyscloud/resource_exporter/resource_exporter.go
+++ b/genesyscloud/resource_exporter/resource_exporter.go
@@ -246,18 +246,30 @@ var unsafeNameChars = regexp.MustCompile(`[^0-9A-Za-z_-]`)
 var unsafeNameStartingChars = regexp.MustCompile(`[^A-Za-z_]`)
 
 func sanitizeResourceNames(idMetaMap ResourceIDMetaMap) {
-	idMetaMapCopy := idMetaMap
+	// Pull out all the original names of the resources for reference later
+	originalResourceNames := make(map[string]string)
+	for k, v := range idMetaMap {
+		originalResourceNames[k] = v.Name
+	}
+
+	// Iterate over the idMetaMap and sanitize the names of each resource
 	for _, meta := range idMetaMap {
 		sanitizedName := SanitizeResourceName(meta.Name)
-		// Append a hash of the original name to ensure uniqueness for names that were sanitized
-		// with a matching pattern
+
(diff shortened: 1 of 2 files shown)


bbbco:
Here's an example of what it looks like before:
```
genesyscloud_user_roles.ashley_rapone_genesys_com_3226576763
genesyscloud_user_roles.brian_goad_genesys_com_2300943017
genesyscloud_user_roles.chris_woods_genesys_com_1015040417
genesyscloud_user_roles.fa_advisor1_pjfinancial_com_1944987575
genesyscloud_user_roles.faadvisor2_pjfinancial_com_3819324546
genesyscloud_user_roles.faadvisor3_pjfinancial_com_1654418959
genesyscloud_user_roles.george_ganahl_genesys_com_4023066032
genesyscloud_user_roles.monty_du_pont_com_1458860085
genesyscloud_user_roles.monty_du_pont_com_2192183309
genesyscloud_user_roles.patrick_johnson_genesys_com_2916959053
genesyscloud_user_roles.test_agent_genesys_com_3502235651
```
and after this change:
```
genesyscloud_user_roles.ashley_rapone_genesys_com
genesyscloud_user_roles.brian_goad_genesys_com
genesyscloud_user_roles.chris_woods_genesys_com
genesyscloud_user_roles.fa_advisor1_pjfinancial_com
genesyscloud_user_roles.faadvisor2_pjfinancial_com
genesyscloud_user_roles.faadvisor3_pjfinancial_com
genesyscloud_user_roles.george_ganahl_genesys_com
genesyscloud_user_roles.monty_du_pont_com_1458860085
genesyscloud_user_roles.monty_du_pont_com_2192183309
genesyscloud_user_roles.patrick_johnson_genesys_com
genesyscloud_user_roles.test_agent_genesys_com
```
You can see that it handles the edge case where two resource names could collide and appends the hash suffix to both (each) resource names.

charliecon (approved):

PR #658, opened 2023-10-09
Outbound Ruleset data source

dginty4:


genesyscloud/outbound_ruleset/genesyscloud_outbound_ruleset_proxy.go
```diff
@@ -157,12 +157,20 @@ func getOutboundRulesetIdByNameFn(ctx context.Context, p *outboundRulesetProxy,
...
+		for _, rulesetSdk := range *rulesets.Entities {
+			if *rulesetSdk.Name == name {
+				log.Printf("Retrieved the ruleset id %s by name %s", *rulesetSdk.Id, name)
+				ruleset = rulesetSdk
+			}
+		}
 		return "", false, fmt.Errorf("Too many values returned in look for outbound rulesets.  Unable to choose 1 ruleset.  Please refine search and continue.")
+	} else {
+		log.Printf("Retrieved the ruleset id %s by name %s", *(*rulesets.Entities)[0].Id, name)
+		ruleset = (*rulesets.Entities)[0]
```
charliecon, line 171: I don't think this `if/else` is a good idea. If the user provides the name "foobar" and the only item in the entities array is a ruleset called "foobar1", the ID for that ruleset will be returned, when really the user should see an error because they got the name wrong.

carnellj-genesys: +1

PR #660, opened 2023-10-13
No jira: Fixing media retention policy tests, user_roles tests, and refactoring function inside util/lists package

charliecon:


genesyscloud/user_group_role_common.go
```diff
@@ -54,102 +54,142 @@ func getAssignedGrants(subjectID string, authAPI *platformclientv2.Authorization
...
+		}
+		roleList = roleListReordered
+	}
+
+	return roleList, resp, nil
 }
 
-func updateSubjectRoles(ctx context.Context, d *schema.ResourceData, authAPI *platformclientv2.AuthorizationApi, subjectType string) diag.Diagnostics {
-	if d.HasChange("roles") {
-		rolesConfig := d.Get("roles")
```
charliecon, line 84: The only real changes to this function was moving everything out of the if statements to reduce nesting

Instead of it being:

```golang
if d.HasChange("roles") {
  // all the code
}
return nil
```

it is:

```golang
if !d.HasChange("roles") {
  return nil
}
// all the code ...
return nil
```

carnellj-genesys (approved):

PR #665, opened 2023-10-18
Revert "[DEVENGAGE-2189] Only add a random suffix on export to resources that have been sanitized with the same matching name pattern"

carnellj-genesys:
Reverts MyPureCloud/terraform-provider-genesyscloud#653


carnellj-genesys (commented):
@dginty4 Hey can you review and approve.  We need to back this change out because it breaks some of the PS work.

@bbbco Backing out and then lets chat further.

dginty4 (approved):

PR #670, opened 2023-10-24
bug/devtooling 31 - Adding logging to queue member export

dginty4:
Adding logging to get queue member call to help track down problems if members are left out. I also refactored some of the file to make it shorter


genesyscloud/resource_genesyscloud_routing_queue.go
```diff
@@ -1635,15 +1496,28 @@ func updateQueueUserRingNum(queueID string, userID string, ringNum int, api *pla
 }
 
 func getRoutingQueueMembers(queueID string, memberBy string, api *platformclientv2.RoutingApi) ([]platformclientv2.Queuemember, diag.Diagnostics) {
```
dginty4, line 1498: Added the logging here

genesyscloud/resource_genesyscloud_routing_queue.go
```diff
@@ -1184,21 +1091,10 @@ func buildSdkConditionalGroupRouting(d *schema.ResourceData) (*platformclientv2.
...
-			}
-
-			if conditionValue, ok := ruleSettings["condition_value"].(float64); ok {
-				sdkCGRRule.ConditionValue = &conditionValue
-			}
-
-			if metric, ok := ruleSettings["metric"].(string); ok {
-				sdkCGRRule.Metric = &metric
-			}
+			sdkCGRRule.WaitSeconds = platformclientv2.Int(ruleSettings["wait_seconds"].(int))
```
charliecon, line 1094: There was logic before to check if `wait_seconds` existed in the map and to set it on the object if it did. It doesn't look like the Int function in the sdk provides that. It'd be grand if the field was required because that would guarantee the value is set, but in this case a panic could occur.

I tested it out with the code below. If `c` isn't added to the map, you get a panic. If it is, everything runs fine.

```golang
package main

import (
	"fmt"

	"github.com/mypurecloud/platform-client-sdk-go/v115/platformclientv2"
)

type Foo struct {
	Bar *int
}

func main() {
	var x = make(map[string]interface{})

	x["c"] = 3

	var f Foo

	f.Bar = platformclientv2.Int(x["c"].(int))
        if f.Bar != nil {
		fmt.Println(*f.Bar)
        }
}

```

genesyscloud/resource_genesyscloud_routing_queue.go
```diff
@@ -1635,15 +1496,28 @@ func updateQueueUserRingNum(queueID string, userID string, ringNum int, api *pla
...
-
 	var members []platformclientv2.Queuemember
+	const pageSize = 100
+
+	// Need to call this method to find the member count for a queue. GetRoutingQueueMembers does not return a `total` property for us to use.
+	queue, _, err := api.GetRoutingQueue(queueID)
+	queueMembers := *queue.MemberCount
+	if err != nil {
+		return nil, diag.Errorf("Can't find queue %s", queueID)
+	}
```
charliecon, line 1507: I would move this error check up a line, before trying to access a value on the object `queue`

genesyscloud/resource_genesyscloud_routing_queue.go
```diff
@@ -513,19 +525,10 @@ func ResourceRoutingQueue() *schema.Resource {
 }
 
 func createQueue(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	name := d.Get("name").(string)
-	divisionID := d.Get("division_id").(string)
-	description := d.Get("description").(string)
-	skillEvaluationMethod := d.Get("skill_evaluation_method").(string)
-	autoAnswerOnly := d.Get("auto_answer_only").(bool)
-	enableTranscription := d.Get("enable_transcription").(bool)
-	suppressInQueueCallRecording := d.Get("suppress_in_queue_call_recording").(bool)
```
charliecon, line 522: It looks like this isn't replaced with anything below. That's probably why you had to add this field to ImportStateVerifyIgnore in the tests

carnellj-genesys (approved):

charliecon (approved):

PR #673, opened 2023-10-24
add new region support

justin-k-genesys:


carnellj-genesys (approved):
These should go out with the release on Friday

PR #678, opened 2023-10-30
DEVTOOLING-209: Added code to prevent the first record from ever expo…

carnellj-genesys:
…rting the queue id for the first conditional routing rule.


charliecon (approved):
Looks good. I wrote the original code and I didn't consider that another rule could reference the queue that it is contained in

PR #680, opened 2023-11-01
no jira: Using after cursor in workbins getAll function

charliecon:
I noticed that the getAll function wasn't using the after cursor in its requests. By setting `pageSize` to 2 and creating 4 workbins, I was able to confirm that it did get stuck in an infinite loop.


dginty4:
No sure if it's related but on of the workbin tests failed

charliecon:
> No sure if it's related but one of the workbin tests failed

It isn't related to this change. The workbin permissions are division aware and those tests are updating the division_id field to newly created ones.

carnellj-genesys (approved):

dginty4 (approved):

PR #682, opened 2023-11-01
DEVTOOLING-38: Updating flow description instead of name in tests

charliecon:
When you invoke the archy import service, it looks up the flow by name and type. If it exists, it updates the current flow. If not, it creates a new flow. That means if a cx as code user tries to change the name on an update, a new flow with a new guid is created and the previous is not cleaned up. We already noted this in the docs.

The reason why there are so many flows not getting cleaned up is because we were updating the name and type in a few tests and verifying the value after. I changed it so that we update the flow description instead.


carnellj-genesys:
Dohhh!!!!!!

carnellj-genesys (approved):

PR #684, opened 2023-11-03
bug/DEVTOOLING-134: Process automation trigger does not support values attribute

dginty4:


genesyscloud/process_automation_trigger/process_automations_triggers_struct.go
```diff
@@ -58,6 +58,7 @@ func (p *ProcessAutomationTrigger) toJSONString() (string, error) {
 		value["jsonPath"] = obj["jsonPath"]
 		value["operator"] = obj["operator"]
 		value["value"] = obj["value"]
+		value["values"] = obj["values"]
```
carnellj-genesys, line 61: That is one of easiest bug fixes I have ever seen.  We should update the test cases with an example using the values attribute

charliecon (approved):

carnellj-genesys (approved):

PR #695, opened 2023-11-09
DEVTOOLING-235 Refactored outbound sequence

dginty4:
This PR will fix the cyclic dependency error in the outbound package that was caused when I refactored outbound_campaigns. I refactored outbound sequences and moved outbound campaignrule into its own package. With the release tomorrow I wasn't able to refactor outbound campaignrules fully so we'll finish it in the next sprint. I also had to clean up some of the tests in the outbound package, I missed them while refactoring outbound_campaigns since the tests weren't running


charliecon (commented):
I noticed some test failures. One says "`Error: Failed to delete outbound sequence ecced297-3412-41ae-b998-dc5a5d37dbd2: Failed to delete outbound sequence: API Error: 400 - Sequence cannot be deleted while it's running (256fefb4-b8c0-42fb-9f3a-ba3421286fc3)`" , and another mentions an invalid campaign. Do these fail when you run them locally as well?

dginty4:
> I noticed some test failures. One says "`Error: Failed to delete outbound sequence ecced297-3412-41ae-b998-dc5a5d37dbd2: Failed to delete outbound sequence: API Error: 400 - Sequence cannot be deleted while it's running (256fefb4-b8c0-42fb-9f3a-ba3421286fc3)`" , and another mentions an invalid campaign. Do these fail when you run them locally as well?

They didn't when I ran them but they are probably intermittent. I will try to replicate them

dginty4:
@charliecon I've addressed the failing outbound sequence test. I'm not sure what to do about the others, they don't happen for me locally. It might have something to do with everything running in parallel

charliecon (approved):
Do they not work with 10 seconds sleep? Twenty feels like a lot

dginty4:
> Do they not work with 10 seconds sleep? Twenty feels like a lot

I thought ten was enough too but I saw on the matix tests earlier it failed to delete a campaign because it hadn't stopped yet so I increased it

PR #709, opened 2023-11-15
DEVTOOLING-58 Add task management schema resource

PrinceMerluza:
Add task management schema resource

Some notes:
- Created manually (not with resource-generator) since the schema object is of type `Dataschema` and is not considered as a main object in the swagger file. 
image

- there's a 'manual' API call done in `getTaskManagementWorkitemSchemaDeletedStatusFn` this is because the API returns an object with `deleted` property but this is not part of the `Dataschema` struct in the SDK.
- 'getAll' API returns everything in bulk including deleted schemas (no pagination).


dginty4:
@PrinceMerluza the cx as code generator has an official open source repo now, https://github.com/MyPureCloud/cxascode-resource-generator. This one is also updated to update the swagger docs everyday and use the preview swagger so you shouldn't encounter the problem above anymore. I would also recommend doing a git pull every now and again to make sure everything is up to date as the project in the old repo also has the functionality I just mentioned.

PrinceMerluza:
> @PrinceMerluza the cx as code generator has an official open source repo now, https://github.com/MyPureCloud/cxascode-resource-generator. This one is also updated to update the swagger docs everyday and use the preview swagger so you shouldn't encounter the problem above anymore. I would also recommend doing a git pull every now and again to make sure everything is up to date as the project in the old repo also has the functionality I just mentioned.

Oh that's cool. Congrats on getting it on the official repo. I already missed the tool when I started this resource, so looking forward to using it for the task management `worktype` after this one.

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_proxy.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// getTaskManagementWorkitemSchemaDeletedStatus gets the deleted status of a Genesys Cloud task management workitem schema
+func (p *taskManagementProxy) getTaskManagementWorkitemSchemaDeletedStatus(ctx context.Context, schemaId string) (isDeleted bool, statusCode int, err error) {
+	return p.getTaskManagementWorkitemSchemaDeletedStatusAttr(ctx, p, schemaId)
+}
+
+// createTaskManagementWorkitemSchemaFn is an implementation function for creating a Genesys Cloud task management workitem schema
+func createTaskManagementWorkitemSchemaFn(ctx context.Context, p *taskManagementProxy, schema *platformclientv2.Dataschema) (*platformclientv2.Dataschema, error) {
+	createdSchema, _, err := p.taskManagementApi.PostTaskmanagementWorkitemsSchemas(*schema)
```
carnellj-genesys, line 107: Lets put a log message before the create and after the create that logs by name the schema being created.  Immediately after the call log by name, resp status code, error and correlation id that the call was created successfully.  The reason why I want to have this is that often times when I am working with the PS team where they are exporting mass numbers of items they might have a failure and they only know the failed item by name.  I want to log the correlation id to because if there is a failure they can quickly see the correlation id rather then digging through the SDK debug log and trying to find the specific all.

genesyscloud/task_management_workitem_schema/resource_genesyscloud_task_management_workitem_schema_unit_test.go
```diff
@@ -0,0 +1,319 @@
+package task_management_workitem_schema
```
carnellj-genesys, line 1: Nice mix of unit and acceptance tests

carnellj-genesys:
> Add task management schema resource
> 
> Some notes:
> 
> * Created manually (not with resource-generator) since the schema object is of type `Dataschema` and is not considered as a main object in the swagger file.
>   image
> * there's a 'manual' API call done in `getTaskManagementWorkitemSchemaDeletedStatusFn` this is because the API returns an object with `deleted` property but this is not part of the `Dataschema` struct in the SDK.
> * 'getAll' API returns everything in bulk including deleted schemas (no pagination).

genesyscloud/task_management_workitem_schema/data_source_genesyscloud_task_management_workitem_schema.go
```diff
@@ -0,0 +1,48 @@
...
+		schemas, retryable, err := proxy.getTaskManagementWorkitemSchemasByName(ctx, name)
+		if err != nil && !retryable {
+			return retry.NonRetryableError(fmt.Errorf("error getting workitem schema %s: %v", name, err))
+		}
+
+		if retryable {
+			return retry.RetryableError(fmt.Errorf("no workitem schema found with name %s", name))
+		}
+
+		if len(*schemas) > 1 {
```
carnellj-genesys, line 40: In the actual function call to retrieve the records we match on name.  We should probably check with the API owners and see if it is possible to have multiple schemas with the same name.  If it, then this error makes sense otherwise it seems a bit redundant.

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_proxy.go
```diff
@@ -0,0 +1,226 @@
...
+func deleteTaskManagementWorkitemSchemaFn(ctx context.Context, p *taskManagementProxy, id string) (statusCode int, err error) {
+	resp, err := p.taskManagementApi.DeleteTaskmanagementWorkitemsSchema(id)
+	if err != nil {
+		return resp.StatusCode, fmt.Errorf("failed to delete task management workitem schema: %s", err)
+	}
+
+	return resp.StatusCode, nil
+}
+
+// getTaskManagementWorkitemSchemaDeletedStatusFn is an implementation function to get the 'deleted' status of a Genesys Cloud task management workitem schema
```
carnellj-genesys, line 183: I am confused about the method below. 

1.   Why do we need this code?  What is happening with these deleted items? 
2.  Why did we have to call this directly?  Is this not available in the SDK?

Usually, CX as Code does not care about items marked as deleted unless, like the user, the item never actually gets deleted and instead gets re-activated if it was deleted.  We also tend to care about deleted items on export.  In the getAll() function we use for export, we don't want to include items that have been deleted.  We only want the active items. (usually)

carnellj-genesys (commented):
Overall great job.  Just need a few questions answered.  Love the unit and acceptance tests.

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_init_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+// initTestresources initializes all test resources and data sources.
+func initTestresources() {
+	providerDataSources = make(map[string]*schema.Resource)
+	providerResources = make(map[string]*schema.Resource)
+
+	reg_instance := &registerTestInstance{}
+
+	reg_instance.registerTestResources()
+	reg_instance.registerTestDataSources()
```
charliecon, line 50: reg_instance -> regInstance
initTestresources -> initTestResources

I'll open a pr myself to refactor the other package that uses these var/func names

genesyscloud/task_management_workitem_schema/resource_genesyscloud_task_management_workitem_schema.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return diag.Errorf("failed to create task management workitem schema: %s", err)
+	}
+
+	d.SetId(*schema.Id)
+
+	// If enabled is set to 'false' do an update call to the schema
+	if enabled, ok := d.Get("enabled").(bool); ok && !enabled {
+		dataSchema.Version = platformclientv2.Int(1)
+		_, err := proxy.updateTaskManagementWorkitemSchema(ctx, *schema.Id, dataSchema)
```
charliecon, line 67: Some log statements could be useful here to help track what's going on. Something like "updating task management [...] because xyz" , and then "Updated task management..."

charliecon (approved):

genesyscloud/task_management_workitem_schema/data_source_genesyscloud_task_management_workitem_schema.go
PrinceMerluza: Hi John, yes it is possible to have multiple schemas with the same name and so I added that code:
image

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_proxy.go
PrinceMerluza: Hi John, so what happens when a schema is deleted is that the record itself can still be queried by `GET`, it would just have a `deleted` property toggled to true (I suppose similar to user). I don't think it can be reactivated but it's different from other objects which would just give us 404 after the object deletion:

image

So the method `getTaskManagementWorkitemSchemaDeletedStatus` is to get whether the the `deleted` property is true or false. Now the reason this is called directly and not on the usual `GET` is because workitem schemas are of type [Dataschema](https://github.com/MyPureCloud/platform-client-sdk-go/blob/master/platformclientv2/dataschema.go) and `Dataschema` does not have a `deleted` property:

image

The API response body(which has `deleted`) does not match the Dataschema struct and so the `deleted` information is lost when using the usual SDK method. I think the reason the JSON and the struct does not match is because `Dataschema` is not used exclusively for workitem schemas but is also used in some fields of external contacts and external organization. Anyway, that's the reason I created a separate method for getting deleted status. I kept the proxy `getTaskManagementWorkitemSchema` method as is using the SDK method to keep it simple.

> In the getAll() function we use for export, we don't want to include items that have been deleted. We only want the active items

The getAll API unfortunately does not have any parameters to customize the query but by default it just includes the non-`deleted` schemas so we don't have to check its results if they are deleted or not.

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_proxy.go
PrinceMerluza: Added logging for create and also noted this for future resources. I guess my question is do we want the detailed logging(with correlation id) on the other crud methods as well?

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_init_test.go
PrinceMerluza: Oops, did not notice the casings after the copypaste. Fixed.

genesyscloud/task_management_workitem_schema/resource_genesyscloud_task_management_workitem_schema.go
PrinceMerluza: Added logging on conditional update

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_proxy.go
carnellj-genesys: Right now I am only putting it on the create because that's where the problems tend to be.  Declan is reviewing the new golang slog capability so we might be revamping a lot of our coding work anyways

genesyscloud/task_management_workitem_schema/genesyscloud_task_management_workitem_schema_proxy.go
carnellj-genesys: Cool.  Thanks for the writeup

carnellj-genesys (approved):

PR #712, opened 2023-11-16
Devtooling 233 - Refactor outbound campaignrule

dginty4:


genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_unit_test.go
```diff
@@ -0,0 +1,376 @@
...
+		assert.Equal(t, *(*testCampaignRule.CampaignRuleActions)[0].CampaignRuleActionEntities.UseTriggeringEntity, *(*campaignRule.CampaignRuleActions)[0].CampaignRuleActionEntities.UseTriggeringEntity, "campaignRule.CampaignRuleActions[0].CampaignRuleActionEntities.UseTriggeringEntity check failed in update updateOutboundCampaignruleAttr")
+		assert.EqualValues(t, *(*testCampaignRule.CampaignRuleActions)[0].CampaignRuleActionEntities.Sequences, *(*campaignRule.CampaignRuleActions)[0].CampaignRuleActionEntities.Sequences, "campaignRule.CampaignRuleActions[0].CampaignRuleActionEntities.Sequences check failed in update updateOutboundCampaignruleAttr")
+		assert.EqualValues(t, *(*testCampaignRule.CampaignRuleActions)[0].CampaignRuleActionEntities.Campaigns, *(*campaignRule.CampaignRuleActions)[0].CampaignRuleActionEntities.Campaigns, "campaignRule.CampaignRuleActions[0].CampaignRuleActionEntities.Campaigns check failed in update updateOutboundCampaignruleAttr")
+
```
charliecon, line 169: There must be a cleaner way to accomplish this. This is tough to read and maintain. Maybe translating both objects to json and comparing to see they're equal? Pretty sure there's some package that lets you do that

charliecon (commented):
looks like there's an issue in `genesyscloud/outbound/resource_genesyscloud_outbound_init_test.go` as well

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_unit_test.go
dginty4: I found this package https://github.com/google/go-cmp that can check to see if the properties of two objects are the same. it will output an empty string if they are, if not it outputs an string showing the differences between the two. I tried it with two callable timesets and this was the output:
```
  platformclientv2.Callabletimeset{
        SetFieldNames: nil,
        Id:            nil,
-       Name:          &"Name 1",
+       Name:          &"Name 2",
        DateCreated:   nil,
        DateModified:  nil,
        Version:       nil,
        CallableTimes: &[]platformclientv2.Callabletime{
                {
                        SetFieldNames: nil,
                        TimeSlots: &[]platformclientv2.Campaigntimeslot{
                                {
                                        SetFieldNames: nil,
-                                       StartTime:     &"12.00.00",
+                                       StartTime:     &"9.00.00",
-                                       StopTime:      &"18.00.00",
+                                       StopTime:      &"10.00.00",
                                        Day:           nil,
                                },
                        },
-                       TimeZoneId: &"Ireland",
+                       TimeZoneId: &"England",
                },
        },
        SelfUri: nil,
  }

```

charliecon: Yeah that looks good 👍 that output string when they're not equal is an ideal error message as well

carnellj-genesys (approved):

charliecon (approved):

PR #717, opened 2023-11-20
Jenkinstest

kavinbala96:


Jenkinsfile
```diff
@@ -1,27 +1,78 @@
+@Library('pipeline-library') _
 pipeline {
+    agent {
+        node{
+            label "dev_mesos_large_v2"
+        }
+    }
+
+    environment {
+        CREDENTIALS_ID  = "GENESYSCLOUD_OAUTHCLIENT_ID_AND_SECERET"
```
carnellj-genesys, line 10: SECERT=>SECRET

kavinbala96: Will check with DK.

kavinbala96: fixed

Jenkinsfile
```diff
@@ -1,27 +1,78 @@
...
+        GOPATH = "$HOME/go"
+		TF_ACC = "1"
+        TF_LOG = "DEBUG"
+        TF_LOG_PATH = "../test.log"
+		GENESYSCLOUD_REGION = "us-east-1"
+        GENESYSCLOUD_SDK_DEBUG =  "true"
+        GENESYSCLOUD_TOKEN_POOL_SIZE =  20
+    }
+    tools {
+        go 'Go 1.20'
```
dginty4, line 20: We are using go 1.21 in the provider, should we use the same here?

Jenkinsfile
```diff
@@ -1,27 +1,78 @@
...
-    }
-  }
+
+        stage('Upload artifacts') {
+            steps {
+                archiveArtifacts artifacts: 'coverage.html', allowEmptyArchive: true
+
+            }
+        }
+
```
charliecon, line 75: Is the release stage supposed to be gone? I think it is missing from the file now

Jenkinsfile
kavinbala96: 1.21 isn't available as of yet in Jenkins.

HemanthDogiparthi12 (approved):

PR #723, opened 2023-11-24
Add Auth Product DataSource

HemanthDogiparthi12:


genesyscloud/authorization_product/data_source_genesyscloud_authorization_product.go
```diff
@@ -0,0 +1,65 @@
...
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: `Authorization Product name.`,
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+		},
+	}
+}
```
charliecon, line 35: Could this be moved to a `_schema.go` file like with other packages? Just to keep with the format

charliecon (approved):

PR #737, opened 2023-12-05
Devtooling 242 add field

kavinbala96:


charliecon (commented):
You'll need to run `go generate` at the root of the project. This will update the docs based on the schema data you have added. Before running that command, I would also edit the example resource. You can find that here - examples/resources/genesyscloud_responsemanagement_response/resource.tf
You'll see there is a comment about valid response_type values, you just need to add "Footer". You can also add a footer object, if you want.

carnellj-genesys (approved):

genesyscloud/resource_genesyscloud_responsemanagement_response.go
```diff
@@ -464,6 +493,36 @@ func buildSdkresponsemanagementresponseWhatsappdefinition(whatsappdefinition *sc
...
+
+			applicableResourcesList := lists.InterfaceListToStrings(applicableResources)
+
+			sdkFootertemplate.ApplicableResources = &applicableResourcesList
+		}
+	}
+
+	return &sdkFootertemplate
+
+}
```
dginty4, line 524: All the whitespace in this function seems unnecessary

dginty4 (approved):

HemanthDogiparthi12 (approved):

PR #738, opened 2023-12-07
DEVTOOLING-9 Wrapupcodes mapping export bug

PrinceMerluza:
Solution(s):

- Refactoring the attributes. Currently, the resource requires `mappings` attribute while `default_set` is optional. Reversed it to make `mappings` optional and `default_set` required. The reason is if we want to manage mappings, we want user to set the explicit default (even if it's the same default no flags) and wuc mapings (which are overrides) themselves are optional. ie, this allows a use case where a user just wants to change the `default_set` so it will apply to all wucs. The current implementation does not allow it. Also fixes an ambiguity of if omitting an array type field is the same as sending an empty array. 
- Fix issue where incorrect wuc mappings are exported. The API wrapupcode actually allows fake/invalid wuc uuids to be set. This means the records will stay in Genesys Cloud even if they don't correspond to an existing wuc. Also, even if you delete a wuc, the record of the mapping itself will stay and can be retrieved. This fix makes it so only valid records (one where the corresponding wuc actually exist) will be exported.

**Additional feature**: 
- Add an `AllowEmptyArrays` field in the `ResourceExporter`. This makes it so that arrays that are empty will not be sanitized from existence and appear in the exported configuration files.
- Rationale: Currently, empty arrays are sanitized by the exporter because most of the times it's the same as omitting it from the API. However in this case, we want the empty arrays to be explicit because omitting them instead has a different behavior.

Example Problem Use Case: 
If an org wants to export their wuc mappings configuration where `default_set` is [] and they want to apply it to a different org but the 2nd org already has a different default_set ( for example with some flags set). In the current implementation, since the `default_set` is not required and since it won't be exported due to empty array sanitization, we lose the data of what we want the `default_set` to be. Since omission of the field in the API just means "do not change this property" even if what we really want is "this is empty because we want the actual array to be empty". This ambiguity causes the 2nd org to not have a matching config from the 1st. Explicit declarations fixes this.

Another Problem with Current Exporter:
- The `mappings.flags` property is required in the schema however since it can be an empty array, it's also omitted from export. This can result in exports such as:

image

We want to retain these empty arrays because they're required attributes and terraform will complain about their absence.
The `AllowEmptyArrays` feature fixes this so we can get the expected export:

image

+ unit test for exporter AllowEmptyArrays functionality


carnellj-genesys (approved):

charliecon (commented):
Thanks for all the detail! We were waiting for the API team to sort out the issue with the invalid wrap up code ids coming back, but if it hasn't happened by now I think adding the check to our provider is a good idea

charliecon (approved):

PR #755, opened 2023-12-14
DEVTOOLING-282

willjeuniaux-genesys:
* Exporting two functions of the integration package: GetIntegrationsProxy & GetIntegrationById
* These functions are used within the integration_credential package to make sure we're not exporting an integration credential for an integration that no longer exists in GC (HTTP 404)
* In case there is an error other than 404 while trying to get the interaction, we trace the message but still export the integration credential.

### DIFF
--- a/genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
+++ b/genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
@@ -59,6 +59,7 @@ func getAllCredentials(ctx context.Context, clientConfig *platformclientv2.Confi
 			_, resp, err := integrationsProxy.GetIntegrationById(ctx, integrationId)
 			if err != nil {
 				if gcloud.IsStatus404(resp) {
+					log.Printf("Integration id %s no longer exist, we are therefore not exporting the associated integration credential id %s", integrationId, *cred.Id)
 					continue
 				} else {
 					log.Printf("Integration id %s exists but we got an unexpected error retrieving it: %v", integrationId, err)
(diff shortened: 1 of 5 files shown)


genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
```diff
@@ -41,16 +43,27 @@ utils function in the package.  This will keep the code manageable and easy to w
...
 	}
 
 	for _, cred := range *credentials {
 		log.Printf("Dealing with credential id : %s", *cred.Id)
 		if cred.Name != nil { // Credential is possible to have no name
+			// Verify that the integration entity itself exist before exporting the integration credentials associated to it: DEVTOOLING-282
+			integrationId := strings.Split(*cred.Name, "Integration-")[1]
+			_, resp, err := integrationsProxy.GetIntegrationById(ctx, integrationId)
+			if err != nil {
+				if gcloud.IsStatus404(resp) {
```
carnellj-genesys, line 61: We should put a log message here indicating that we hit this part of the code

carnellj-genesys (approved):

genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
willjeuniaux-genesys: Agreed, done!

charliecon (approved):

charliecon:
Hi @willjeuniaux-genesys 

I had to revert this PR because it contained cyclic dependencies; the integration_credential package imported the integration package, which imported the integration_credential package......

Could you open another PR off of this branch and fix these issues? Running `go vet ./genesyscloud/...` from the root of the project will let you know if it is fixed.

You could just duplicate the getIntegrationById function in each package. It's not the prettiest solution, but it is the simplest.

-Charlie

PR #762, opened 2023-12-15
Devtooling 293: Fixing bug + adding detail to error message in user prompts resource

charliecon:


genesyscloud/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -886,47 +884,43 @@ func readUserPrompt(ctx context.Context, d *schema.ResourceData, meta interface{
...
-				resourceFilename := resourceMap["filename"]
-				if resourceFilename.(string) == "" {
+				resourceMap, ok := promptResource.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				resourceFilename, ok := resourceMap["filename"].(string)
+				if !ok || resourceFilename == "" {
 					continue
 				}
```
charliecon, line 901: This is part of the bug fix. This condition:
```golang
resourceFilename := resourceMap["filename"]
if resourceFilename.(string) == "" {
continue
}
```

was not working as expected because, if filename didn't exist in the map, <nil> would be returned, not an empty string. So the for loop would not "continue"

dginty4 (approved):

PR #765, opened 2023-12-15
Devtooling 193 refactor

kavinbalagen:
Ready to merge


genesyscloud/resource_genesyscloud_init.go
```diff
@@ -54,7 +54,7 @@ func registerDataSources(l registrar.Registrar) {
 	l.RegisterResource("genesyscloud_routing_wrapupcode", ResourceRoutingWrapupCode())
 
 	l.RegisterDataSource("genesyscloud_user", DataSourceUser())
-	l.RegisterDataSource("genesyscloud_telephony_providers_edges_extension_pool", dataSourceExtensionPool())
+	//l.RegisterDataSource("genesyscloud_telephony_providers_edges_extension_pool", dataSourceExtensionPool())
```
dginty4, line 57: If these are not needed you can remove them

genesyscloud/resource_genesyscloud_init_test.go
```diff
@@ -78,7 +78,7 @@ func (r *registerTestInstance) registerTestResources() {
 	providerResources["genesyscloud_routing_settings"] = ResourceRoutingSettings()
 	providerResources["genesyscloud_routing_utilization"] = ResourceRoutingUtilization()
 	providerResources["genesyscloud_routing_wrapupcode"] = ResourceRoutingWrapupCode()
-	providerResources["genesyscloud_telephony_providers_edges_extension_pool"] = ResourceTelephonyExtensionPool()
+	//providerResources["genesyscloud_telephony_providers_edges_extension_pool"] = ResourceTelephonyExtensionPool()
```
dginty4, line 81: Same here

dginty4 (approved):

genesyscloud/resource_genesyscloud_init.go
kavinbalagen: addressed

genesyscloud/resource_genesyscloud_init_test.go
kavinbalagen: addressed

carnellj-genesys (approved):

genesyscloud/telephony_providers_edges_extension_pool/resource_genesyscloud_telephony_providers_edges_extension_pool_test.go
```diff
@@ -76,6 +76,10 @@ func TestAccResourceExtensionPoolBasic(t *testing.T) {
 }
 
 func deleteExtensionPoolWithNumber(startNumber string) error {
+	sdkConfig, err := gcloud.AuthorizeSdk()
+	if err != nil {
+		log.Fatal(err)
+	}
```
charliecon, line 82: I would return this error and log it where the function is called from, because the rest of this func won't work if the sdk wasn't authorized correctly

PR #769, opened 2023-12-19
Devtooling 282

willjeuniaux-genesys:
[Previous PR](https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/755) had circular dependencies, code reworked as recommended by @charliecon 

GetIntegrationById is added to integration_credential package directly, which no longer imports integration package. 

The integration_credential checks we're not exporting an integration_credential resource for an integration that no longer exists in GC (HTTP 404)

In case there is an error other than 404 while trying to get the interaction, we trace the message but still export the integration credential.

**go vet ./genesyscloud/...** returns the same output as before the change:
terraform-provider-genesyscloud/genesyscloud/tfexporter
genesyscloud\tfexporter\resource_genesyscloud_tf_export_test.go:35:2: struct field tag `-` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair
genesyscloud\tfexporter\resource_genesyscloud_tf_export_test.go:42:2: struct field tag `-` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair
genesyscloud\tfexporter\resource_genesyscloud_tf_export_test.go:47:2: struct field tag `-` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair

### DIFF
--- a/genesyscloud/integration_credential/genesyscloud_integration_credential_proxy.go
+++ b/genesyscloud/integration_credential/genesyscloud_integration_credential_proxy.go
@@ -37,6 +37,7 @@ type getIntegrationCredByIdFunc func(ctx context.Context, p *integrationCredsPro
 type getIntegrationCredByNameFunc func(ctx context.Context, p *integrationCredsProxy, credentialName string) (credential *platformclientv2.Credentialinfo, retryable bool, err error)
 type updateIntegrationCredFunc func(ctx context.Context, p *integrationCredsProxy, credentialId string, credential *platformclientv2.Credential) (*platformclientv2.Credentialinfo, error)
 type deleteIntegrationCredFunc func(ctx context.Context, p *integrationCredsProxy, credentialId string) (responseCode int, err error)
+type getIntegrationByIdFunc func(ctx context.Context, p *integrationCredsProxy, integrationId string) (integration *platformclientv2.Integration, response *platformclientv2.APIResponse, err error)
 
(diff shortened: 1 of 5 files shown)


genesyscloud/integration_credential/genesyscloud_integration_credential_proxy.go
```diff
@@ -48,6 +49,7 @@ type integrationCredsProxy struct {
 	getIntegrationCredByNameAttr getIntegrationCredByNameFunc
 	updateIntegrationCredAttr    updateIntegrationCredFunc
 	deleteIntegrationCredAttr    deleteIntegrationCredFunc
+	getIntegrationByIdAttr       getIntegrationByIdFunc
```
charliecon, line 52: I think you forgot to give this a value? As in the actual implementation function that calls the API `getIntegrationByIdFn`

willjeuniaux-genesys: Good catch, thank you. I've updated the code & tested successfully. I didn't see the error on my earlier test because i was not testing against the last version of my build. It looks like I am having a problem with the value resolved by var PLUGINS_DIR in the makefile, which points to a folder that is not the one terraform is looking at. I initially caught this error when I started setting things up and forgot I recently commented out my local fix. I will create a ticket for this to investigate further.

charliecon (approved):

PR #770, opened 2023-12-19
Feat/update utilization contract

tsantos-genesys:
Bumps platform-client-sdk-go from 116 to 119, I need the new version before I can put up new PRs with utilization labels.

This version bump causes resource_genesyscloud_routing_utilization and resource_genesyscloud_user to break, as there was a contract change. A class was renamed from Utilization to UtilizationResponse, and that change is being made here to keep those resources working. I ran a test in each resource and they passed.

However, **this bump is causing an issue when running "go generate".**

genesyscloud_recording_media_retention_policy_proxy.go is breaking, because getFormsEvaluationFn is supposed to return Evaluationform, but the api is returning Evaluationformresponse. I assume it's a similar change to my utilization one, but I don't know for sure, so I didn't change it, as I don't know the whole process.

If I should try to fix the above, please let me know. I just wanted to ask for guidance before changing things I'm unfamiliar with.

This is my first PR to this repo, so apologies in advance for mistakes or if I'm jumping the gun on this, and should wait for you guys to update the version. Please feel free to decline this PR.


tsantos-genesys:
Hello, @charliecon, I've updated the Evaluationform in the 3 places the compiler was complaining at.

charliecon (approved):

PR #771, opened 2023-12-19
Devtooling 144 datatable refactor

carnellj-genesys:


GNUmakefile
```diff
@@ -18,7 +18,7 @@ copy-hooks:
 
 # Run acceptance tests
 testacc:
-	TF_ACC=1 go test ./... -v $(TESTARGS) -run TestAcc -timeout 120m -parallel 20  -coverprofile=coverage.out
```
carnellj-genesys, line 21: @kavinbalagen  Did you originally add the "-run TestAcc" this will cause all of the acceptance tests to run even when individual tests are fired off.  I think you put this in here for the test pipeline in Jenkins.  Let me know as I think we can work around this by setting the TESTARGs environment variable in the Jenkins pipeline to be TestAcc.

Thanks.  You didn't do anything wrong I think its not quite working the way it was intended locally

genesyscloud/architect_datatable/resource_genesyscloud_architect_datatable_test.go
```diff
@@ -104,6 +108,29 @@ func TestAccResourceArchitectDatatable(t *testing.T) {
...
 }
 
+func testVerifyDatatablesDestroyed(state *terraform.State) error {
+	archAPI := platformclientv2.NewArchitectApi()
+	for _, rs := range state.RootModule().Resources {
+		if rs.Type != "genesyscloud_architect_datatable" {
+			continue
+		}
+
+		//TODO FIX THIS !!!!
```
charliecon, line 118: is this TODO something that should be done before merge, or when the SDK can be used ?

charliecon (commented):
looks like `go generate` needs to be ran

charliecon (approved):

PR #774, opened 2024-01-02
DEVTOOLING-309: Fixing Architect IVR chunking logic 

charliecon:
This PR addresses 3 issues that existed with the IVR dnis chunking logic:

1. **Speed** - Previously, on update, we were clearing the dnis array on the first PUT, then uploading each chunk again on subsequent requests. So 600 numbers exist in the dnis array, the user adds 20, and we're doing 13 requests (because we're uploading 50 new dnis numbers at a time.) We only need to do one requests in this scenario because the API can handle all 600 dnis that are already associated with the IVR. It's only new numbers that need to be chunked.
    * *Solution* - Slice the difference of the dnis array to establish which numbers are new. Set ivr.Dnis to the numbers that already are associated with the ivr (removing the new ones) + the first chunk, and then add on any new chunks incrementally in subsequent PUT requests.
2. **IVR not being deleted on a failed CREATE** - If there are more than `maxDnisPerRequest` (currently set to 50) in the dnis array on an IVR creation, and the chunking logic fails, then a dangling IVR would be left out there. This is because the IVR is created, but a subsequent update failed. As far as terraform is aware it failed on create so there is no need to call the delete function.
    * *Solution* - Manually call the delete IVR function if a chunk upload fails on IVR creation, before returning the error message.
3. **Inaccurate tf plans outputs** - If we're adding 150 new numbers, the tf plan will state that. However, if the third chunk upload fails, that means we have uploaded 100 numbers successfully before failure. Now when the user runs `terraform apply` again, they will see "50 to add". 
    * *Solution* - Save a copy of the IVR before attempting the update with chunking logic, and if the process fails, reset the IVR to it's original state before returning the error message. This way the user will see "150 to add" again, as expected.

### DIFF
--- a/genesyscloud/architect_ivr/genesyscloud_architect_ivr_proxy.go
+++ b/genesyscloud/architect_ivr/genesyscloud_architect_ivr_proxy.go
@@ -258,6 +258,13 @@ func (a *architectIvrProxy) uploadArchitectIvrWithChunkingLogic(ctx context.Cont
 	if len(dnisChunks) > 1 {
 		respIvr, resp, err = a.uploadIvrDnisChunks(ctx, dnisChunks, id)
 		if err != nil {
+			// if the chunking logic failed on create - delete the IVR that was created above
+			// Otherwise the CreateContext func will fail and terraform will assume the IVR was never created
+			if post {
+				if _, deleteErr := a.deleteArchitectIvr(ctx, id); deleteErr != nil {
+					log.Printf("failed to delete ivr '%s' after dnis chunking logic failed: %v", id, deleteErr)
+				}
+			}
 			return respIvr, resp, err
 		}
 	}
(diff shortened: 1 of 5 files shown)


genesyscloud/architect_ivr/genesyscloud_architect_ivr_proxy.go
```diff
@@ -306,3 +308,81 @@ func (a *architectIvrProxy) uploadDnisChunk(ctx context.Context, ivr platformcli
...
+		}
+		log.Printf("Deleted IVR %s", id)
+		return nil
+	}
+
+	if originalConfig == nil {
+		return fmt.Errorf("cannot reset IVR %s without a copy of the original configuration", id)
+	}
+
+	originalConfig.Version = nil
```
HemanthDogiparthi12, line 380: why is this assigned to nil

charliecon: The object originalConfig was read using a get earlier in the process, before a couple update happens and the value of version was incremented a few times, so trying to do a put with an old version number was causing weird behaviour. Better to just omit it and let the API compute it

carnellj-genesys (approved):

carnellj-genesys:
Very nice.  Looks good.

PR #775, opened 2024-01-02
Trunk Package Refactor.

kavinbalagen:


charliecon (commented):
If you run `go vet ./genesyscloud/...` at the root of the project, you'll see some messages. I saw 

```log
genesyscloud/telephony_providers_edges_trunk/genesyscloud_telephony_providers_edges_trunk_proxy.go:42:29: literal copies lock value from *clientConfig: github.com/mypurecloud/platform-client-sdk-go/v116/platformclientv2.Configuration contains github.com/mypurecloud/platform-client-sdk-go/v116/platformclientv2.APIClient contains github.com/hashicorp/go-retryablehttp.Client contains sync.Once contains sync.Mutex

```

but I wouldn't be surprised if you see a cyclic error message after this one is resolved. Looks like there's a cyclic relationship between `telephony` and `telephony_providers_edges_trunk` 

Also I noticed that `go vet` produced 

```log
# terraform-provider-genesyscloud/genesyscloud/tfexporter
genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go:35:2: struct field tag `-` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair
genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go:42:2: struct field tag `-` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair
genesyscloud/tfexporter/resource_genesyscloud_tf_export_test.go:47:2: struct field tag `-` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair

```

...on this branch. That has already been taken care of on main, so you won't need to worry about that

genesyscloud/telephony_providers_edges_trunk/resource_genesyscloud_telephony_providers_edges_trunk.go
```diff
@@ -18,53 +18,13 @@ import (
...
-		},
-	}
-}
-
 func createTrunk(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	trunkBaseSettingsId := d.Get("trunk_base_settings_id").(string)
 
 	sdkConfig := meta.(*gcloud.ProviderMeta).ClientConfig
-	edgesAPI := platformclientv2.NewTelephonyProvidersEdgeApiWithConfig(sdkConfig)
+	tp := newTrunkProxy(sdkConfig)
```
charliecon, line 25: We should be calling getTruntProxy here so we're not creating a new object every time. Same for the instances below

charliecon (approved):

genesyscloud/telephony_providers_edges_trunk/resource_genesyscloud_telephony_providers_edges_trunk.go
kavinbalagen: addressed

PR #781, opened 2024-01-08
Feat/add utilization labels

tsantos-genesys:
This adds a new resource/data-source for Utilization labels.

I did my best in trying to update the necessary places, but likely missed something.

My main point of concern in this PR are the tests, because **the Utilization Label feature is not yet released**. The endpoints are in PREVIEW in public-api, and only test orgs (like mine) have it enabled.

That would mean that the tests I added would only pass on those orgs, that have it enabled, and break for everybody else, like maybe your test pipeline.

So what do we do with those tests while the feature is not yet released?

I tried to call out in a few places in code comment/descriptions that the thing is not released. If there's a better way, please let me know.


genesyscloud/resource_genesyscloud_init.go
```diff
@@ -49,15 +49,14 @@ func registerDataSources(l registrar.Registrar) {
 	l.RegisterDataSource("genesyscloud_routing_skill", dataSourceRoutingSkill())
 	l.RegisterDataSource("genesyscloud_routing_skill_group", dataSourceRoutingSkillGroup())
 	l.RegisterDataSource("genesyscloud_routing_email_domain", DataSourceRoutingEmailDomain())
+	l.RegisterDataSource("genesyscloud_routing_utilization_label", dataSourceRoutingUtilizationLabel())
 	l.RegisterDataSource("genesyscloud_routing_wrapupcode", DataSourceRoutingWrapupcode())
-	l.RegisterResource("genesyscloud_routing_wrapupcode", ResourceRoutingWrapupCode())
```
tsantos-genesys, line 53: I've removed this resource because it was in the middle of the data sources, and wrapupcode resource is already added below.

genesyscloud/resource_genesyscloud_init.go
```diff
@@ -49,15 +49,14 @@ func registerDataSources(l registrar.Registrar) {
...
 	l.RegisterDataSource("genesyscloud_user", DataSourceUser())
 	l.RegisterDataSource("genesyscloud_widget_deployment", dataSourceWidgetDeployments())
 }
 
 func registerResources(l registrar.Registrar) {
 
-	l.RegisterResource("genesyscloud_routing_wrapupcode", ResourceRoutingWrapupCode())
```
tsantos-genesys, line 60: ...and I've removed this one because there's already another one down below, in line 108

genesyscloud/resource_genesyscloud_init_test.go
```diff
@@ -41,7 +40,6 @@ func (r *registerTestInstance) registerTestResources() {
 	providerResources["genesyscloud_group"] = ResourceGroup()
 	providerResources["genesyscloud_group_roles"] = ResourceGroupRoles()
 	providerResources["genesyscloud_idp_adfs"] = ResourceIdpAdfs()
-	providerResources["genesyscloud_routing_wrapupcode"] = ResourceRoutingWrapupCode()
```
tsantos-genesys, line 44: Same deal here, already added below.

If I made a mistake, please let me know.

charliecon (commented):
I noticed there are two data source examples in the docs folder. Everything inside the docs folder is generated based on the contents inside `examples/` and the details provided to schema objects. You just need to run `go generate` from the root of the project. Of course, if you did that already you can ignore this comment. It's just a common mistake made by new contributors

genesyscloud/data_source_genesyscloud_routing_utilization_label.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		if len(*labels.Entities) == 1 {
+			label := (*labels.Entities)[0]
+			d.SetId(*label.Id)
+			return nil
+		}
+
+		return retry.NonRetryableError(fmt.Errorf("%d labels found with name %s. The data source might be misconfigured", len(*labels.Entities), name))
+	})
```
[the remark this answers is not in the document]
tsantos-genesys: If the api is given "abc", it will return exact match "abc" or nothing.

I added this because the API does support partial matching, using asterisks. So if given "*abc*" it would return "abcd" + whatever else.

My reasoning was if the person writing the data_source wrote "*abc*", then they'd be matching on unexpected things.

Could just throw an error if the input has * instead?

I'll make that change, let me know what you think.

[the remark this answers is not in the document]
charliecon: Ah okay, I'm used to endpoints implementing partial match no the name. Sounds good.

tsantos-genesys:
@charliecon I've addressed the previous comments (I hope).

Thanks for letting me know about the generated docs.

I'd still like guidance into the tests and the feature not being available, as mentioned in the PR description.

Thanks!

charliecon:
> @charliecon I've addressed the previous comments (I hope).
> 
> Thanks for letting me know about the generated docs.
> 
> I'd still like guidance into the tests and the feature not being available, as mentioned in the PR description.
> 
> Thanks!

You can add a function to the tests to verify that the feature is implemented in the target org before continuing. 

Inside the test it would look like:

```golang
if err := testRoutingUtilLabels(); err != nil {
    t.Skipf("%v", err) // be sure to skip the test and not fail it 
}
```

And the function would simply just attempt an api request and inspect the response object:

```golang
func testRoutingUtilLabels(c *platformclientv2.Configuration) error {
	api := platformclientv2.NewRoutingApiWithConfig(sdkConfig) // the variable sdkConfig exists at a package level in ./genesyscloud and is already authorized
	_, resp, _ := api.GetRoutingUtilizationLabels(100, 1, "", "")
	if resp.StatusCode == 501 {
		return fmt.Errorf("feature is not yet implemented in this org.")
	}
	return nil
}
```

tsantos-genesys:
Thanks for the guidance, @charliecon, I've added the skip.

charliecon (approved):

carnellj-genesys (approved):

PR #791, opened 2024-01-11
feat/direct-routing-remove-requirement-of-backup-queue

jonesb-genesys:
As we get ready for beta we decided the requirement of a backup queue no longer makes sense. We can use the current queue as backup if one is not supplied.


carnellj-genesys (approved):

dginty4 (approved):

charliecon (approved):
Looks good. You will just need to run `go generate` from the root of the project to update the docs

PR #792, opened 2024-01-12
No jira: fix outbound tests

charliecon:
This is to fix outbound test failures like:

```log
 resource_genesyscloud_outbound_sequence_test.go:41: POST https://login.mypurecloud.com/oauth/token giving up after 1 attempt(s): Post "https://login.mypurecloud.com/oauth/token": net/http: cannot rewind body after connection loss
```

I suspect this is occurring when we try to call the function `AuthorizeSdk` inside the acceptance tests. Instead, this function is being called once from TestMain to save the config to a package-level variable on initialisation of the package, as is already done for the genesyscloud package.

### DIFF
--- a/genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign.go
+++ b/genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign.go
@@ -64,8 +64,8 @@ func getAllAuthOutboundCampaign(ctx context.Context, clientConfig *platformclien
 
 // createOutboundCampaign is used by the outbound_campaign resource to create Genesys cloud outbound campaign
 func createOutboundCampaign(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	sdkConfig := meta.(*gcloud.ProviderMeta).ClientConfig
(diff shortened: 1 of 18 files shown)


genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign.go
```diff
@@ -180,38 +180,28 @@ func updateOutboundCampaign(ctx context.Context, d *schema.ResourceData, meta in
...
-			}
-			// Handles updating the campaign based on what is set in ResourceData.campaign_status
-			diagErr := updateOutboundCampaignStatus(ctx, d.Id(), proxy, *outboundCampaign, "off")
-			if diagErr != nil {
-				return resp, diagErr
-			}
-			return resp, nil
-		})
-		if diagErr != nil {
+		log.Printf("Turning off Outbound Campaign before deletion")
```
charliecon, line 190: I refactored the above code into the proxy to reduce function size. Also, the code above was making the mistake of retrying on 400, where really we should be updating and then retrying until the field CampaignStatus != "off" . The 400 appeared when we tried to delete it while it was still "on"

carnellj-genesys (approved):

PR #793, opened 2024-01-12
Feat/Adding label support to routing_utilization

tsantos-genesys:
This allows configuring label utilization in the Org utilization (routing_utilization resource).

This is one I had spoken to John Carnell about, since it uses a field that's not public yet, so I had to call the api directly instead of using the existing platform-go-sdk call. It's in my to do list to rework that call once the field is public.

One complicated point in this PR involves how OrgUtilization works in the service that owns it. You can have 10 labels in system, configure only 2 of them, for example, and the response will contain all 10 labels (2 with the custom configuration + 8 with default config).

Translating that to terraform would mean that the configured resource has 2 label_utilizations, but the returned resource has 10...It took me a lot of trouble to get that working here for OrgUtilization, but the same concept will apply to my next PR (for AgentUtilization), and it did not work there at all. It really doesn't like the output not matching the input...

So I just removed the extra labels from the response and everything works. So if you configure 2, the resource will have 2, even if the actual thing in the service has all 10.

I don't know how y'all feel about that, so I'm pointing it out in this description.

I look forward to seeing your feedback. Thanks.

### DIFF
--- a/examples/resources/genesyscloud_routing_utilization/resource.tf
+++ b/examples/resources/genesyscloud_routing_utilization/resource.tf
@@ -1,4 +1,4 @@
-resource "genesyscloud_routing_utilization" "org-utililzation" {
+resource "genesyscloud_routing_utilization" "org-utilization" {
   call {
     maximum_capacity = 1
     include_non_acd  = true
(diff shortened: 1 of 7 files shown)


examples/resources/genesyscloud_routing_utilization/resource.tf
```diff
@@ -23,4 +23,13 @@ resource "genesyscloud_routing_utilization" "org-utililzation" {
     include_non_acd           = false
     interruptible_media_types = ["call", "chat"]
   }
+  label_utilizations {
```
HemanthDogiparthi12, line 26: can we have this representation to be array of label_utilization , instead of having multiple attributes with the same name ?

HemanthDogiparthi12: ok. I think its fine.

carnellj-genesys (approved):

charliecon (approved):

PR #796, opened 2024-01-16
DEVTOOLING-319: plugin crash due to deleted/not exported skills referenced by an outbound ruleset

willjeuniaux-genesys:
**Ticket:**
https://inindca.atlassian.net/browse/DEVTOOLING-319

**Scenario:**
Creation of a resource 'genesyscloud_tf_export'

**Issue analysis:**
The plugin crashes in function RuleSetSkillPropertyResolver of resource_exporter_custom.go if an outbound ruleset references a skill that has either been deleted or has not been exported. The current build assumes the skill is always in the skill map and therefore causes a segmentation fault in this specific scenario when trying to access the ResourceIDMetamap var SanitizedResourceMap.

**Fix:**
In function RuleSetSkillPropertyResolver, we need to make sure the skill is in the skill map before processing it. Fixing that bug is only one part of the problem though, as if a ruleset references a skill that no longer exists in GC, the exported file will be invalid and an error will be thrown during a terraform apply.

That is why I considered the entire ruleset to be invalid if it is referencing such skill. The second part of the fix is to remove the ruleset from the export. This is where function filterOutboundRulesets comes into play. It will parse every ruleset, and check its rules' actions & conditions to search for skills. Each time it finds a skill, it will verify if that skill actually exists. If it doesn't, we filter the ruleset out of the array and evaluate the next ruleset.

If we run into an unexpected scenario where filterOutboundRulesets doesn't filter a ruleset that should have been filtered out, the segmentation fault in function RuleSetSkillPropertyResolver will not occur. Instead, It will write an error code (e.g. 'skill_#id_not_found') to the exported file. 

**Trace**:

2024-01-16T15:31:40.087+0100 [INFO]  provider.terraform-provider-genesyscloud.exe: 2024/01/16 15:31:40 Removing ruleset 7c596858-bbc7-4190-8f3a-d3908c234186, the skill cfb08f13-f8fb-4290-93d1-2dabd7640f09 used in action does not exist in GC anymore: timestamp="2024-01-16T15:31:40.087+0100"
2024-01-16T15:31:40.087+0100 [INFO]  provider.terraform-provider-genesyscloud.exe: 2024/01/16 15:31:40 Removing ruleset 37abd0a4-60b9-4767-9068-21862e5d04f2, the skill English used in condition does not exist in GC anymore: timestamp="2024-01-16T15:31:40.087+0100"
2024-01-16T15:31:40.087+0100 [INFO]  provider.terraform-provider-genesyscloud.exe: 2024/01/16 15:31:40 Removing ruleset 7fb60a82-a44e-44db-a8dd-8aae00e1635e, the skill 29c31d2a-5265-48d1-8254-a908ab4a67ec used in action does not exist in GC anymore: timestamp="2024-01-16T15:31:40.087+0100"

[removedRuleSets.zip](https://github.com/MyPureCloud/terraform-provider-genesyscloud/files/13983346/removedRuleSets.zip)


genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
```diff
@@ -131,3 +144,59 @@ func deleteOutboundRuleset(ctx context.Context, d *schema.ResourceData, meta int
 		return retry.RetryableError(fmt.Errorf("Outbound Ruleset %s still exists", d.Id()))
 	})
 }
+
+// filterOutboundRuleset filters rulesets by removing the ones that are referencing skills that no longer exist in GC
+func filterOutboundRulesets(ruleSets []platformclientv2.Ruleset, skillMap resourceExporter.ResourceIDMetaMap) ([]platformclientv2.Ruleset, diag.Diagnostics) {
+	var filteredRuleSets []platformclientv2.Ruleset
+	log.Printf("Filtering outbound rule sets")
+
+RuleSetLoop:
```
carnellj-genesys, line 153: I didnt realize GoLang supported labels :).

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
```diff
@@ -131,3 +144,59 @@ func deleteOutboundRuleset(ctx context.Context, d *schema.ResourceData, meta int
...
+	for _, ruleSet := range ruleSets {
```
carnellj-genesys, line 154: I would highly recommend you break this function up.  In Go, if use see multi-level for and if statements, it is usually telling you that your code is too complex.  I would at least take that inner set of if and for loops and refactor it out to a single method call.  It will make it easier to write unit tests and prove that your logic is solid.

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
```diff
@@ -131,3 +144,59 @@ func deleteOutboundRuleset(ctx context.Context, d *schema.ResourceData, meta int
...
+		for _, rule := range *ruleSet.Rules {
+			// look through rule actions to check if the referenced skills exist in our skill map or not
+			for _, action := range *rule.Actions {
+				if action.ActionTypeName != nil && strings.ToLower(*action.ActionTypeName) == "set_skills" && action.Properties != nil {
+					for id, value := range *action.Properties {
```
carnellj-genesys, line 159: This block is a good candidate to break out into a a set of smaller methods and then write unit tests.

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
```diff
@@ -131,3 +144,59 @@ func deleteOutboundRuleset(ctx context.Context, d *schema.ResourceData, meta int
...
+						var found bool
+						for _, value := range skillMap {
+							if value.Name == *condition.Value {
+								found = true
+								break
+							}
+						}
+						if !found { // skill name referenced by rule condition is not found in the skill map, we filter the rulset out and evaluate the next one
+							log.Printf("Removing ruleset %s, the skill %s used in condition does not exist in GC anymore", *ruleSet.Id, *condition.Value)
+							continue RuleSetLoop
```
carnellj-genesys, line 192: Found this.  I don't know.  This feels a little dirty.  Its not a goto, but...... :)

carnellj-genesys (changes requested):
Overall looks good.  Please see my comments about the function size and maybe throw some unit tests on the extract methods.  Thanks for the commit.

- John

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
willjeuniaux-genesys: Yup, looked pretty cool! removed though ;)

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
willjeuniaux-genesys: Agreed, got overly excited about labeled loops. Removed

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
```diff
@@ -131,3 +144,59 @@ func deleteOutboundRuleset(ctx context.Context, d *schema.ResourceData, meta int
...
+									continue RuleSetLoop
+								}
+							}
+						}
+					}
+				}
+			}
+			// look through rule conditions to check if the referenced skills exist in our skill map or not
+			for _, condition := range *rule.Conditions {
+				if condition.AttributeName != nil && strings.ToLower(*condition.AttributeName) == "skill" {
```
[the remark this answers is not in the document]
willjeuniaux-genesys: Fixed! :)

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset.go
```diff
@@ -131,3 +142,38 @@ func deleteOutboundRuleset(ctx context.Context, d *schema.ResourceData, meta int
...
+	var filteredRuleSets []platformclientv2.Ruleset
+	log.Printf("Filtering outbound rule sets")
+
+	for _, ruleSet := range ruleSets {
+		var foundDeleted bool
+
+		for _, rule := range *ruleSet.Rules {
+			var err error
+
+			if foundDeleted, err = doesRuleActionsRefDeletedSkill(rule, skillMap); err != nil {
```
carnellj-genesys, line 157: @williamjeuniaux Looks better.  I do have a question though could the two conditions be anded together.  If I am reading this correctly you are breaking, the doesRuleActionsRef and doesRuleConditionsRef both cause a break in the loop.  Why not do an and in a single if and break if either is found true.

It just seems like we could simplify this code a little bit.  Sorry not trying to bit nitpicky

willjeuniaux-genesys: That's ok! The idea is I didn't want to evaluate the conditions at all if there was a deleted skill in the actions. If i add an AND, i believe it will evaluate both.

I could add an OR, so it only evaluates conditions if actions returns false, but then I am getting another problem; it considers the expression as a single value context. Since i am returning two values (bool + error)  the compiler complains.

I could get rid of the error handling and just return a bool, but i am not a huge fan of this idea. Thoughts?

carnellj-genesys (approved):

HemanthDogiparthi12 (approved):

genesyscloud/outbound_ruleset/resource_genesyscloud_outbound_ruleset_unit_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"testing"
+
+	resourceExporter "terraform-provider-genesyscloud/genesyscloud/resource_exporter"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mypurecloud/platform-client-sdk-go/v119/platformclientv2"
+)
+
+func TestDoesRuleConditionsRefDeletedSkill(t *testing.T) {
```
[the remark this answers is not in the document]
carnellj-genesys: @williamjeuniaux Last thing and then we can merge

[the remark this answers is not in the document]
willjeuniaux-genesys: Done @charliecon @carnellj-genesys

PR #802, opened 2024-01-17
Add in Rrule check

HemanthDogiparthi12:


carnellj-genesys (approved):
Just some small questions and changes

genesyscloud/resource_genesyscloud_architect_schedules.go
```diff
@@ -93,9 +96,10 @@ func ResourceArchitectSchedules() *schema.Resource {
 				ValidateDiagFunc: validateLocalDateTimes,
 			},
 			"rrule": {
-				Description: "An iCal Recurrence Rule (RRULE) string.",
-				Type:        schema.TypeString,
-				Optional:    true,
+				Description:      "An iCal Recurrence Rule (RRULE) string.",
+				Type:             schema.TypeString,
```
charliecon, line 100: I noticed there is an extra bit of info in the description on API explorer: "An iCal Recurrence Rule (RRULE) string. It is required to be set for schedules determining when upgrades to the Edge software can be applied." Probably no harm to add it in

genesyscloud/tfexporter/export_common.go
```diff
@@ -219,6 +220,40 @@ func sanitizeE164Number(number string) string {
 	}
 	return number
 }
+func sanitizeRrule(input string) string {
+	attributeRegex := map[string]*regexp.Regexp{
```
charliecon, line 224: unit tests for this function could be good to make sure the reg expressions work as expected

genesyscloud/resource_exporter/resource_exporter.go
```diff
@@ -194,7 +194,19 @@ func (r *ResourceExporter) IsJsonEncodable(attribute string) bool {
...
+	}
+	return lists.ItemInSlice(attribute, values)
+}
+
+func (r *ResourceExporter) IsAttributeRrule(attribute string) bool {
+	values, exists := r.CustomValidateExports["rrule"]
+	if !exists {
+		return false
+	}
+	return lists.ItemInSlice(attribute, values)
```
charliecon, line 209: Could add unit tests here as well

charliecon (approved):

genesyscloud/tfexporter/export_common.go
charliecon: My bad. I assumed that was for `removeTrailingZeros`

PR #820, opened 2024-01-25
[DEVTOOLING-353] Include the flow type in the resource name export

bbbco:
Fixes https://inindca.atlassian.net/browse/DEVTOOLING-353

Now, on export, it outputs like so:

```
resource "genesyscloud_flow" "INQUEUECALL_inqueue" {
  file_content_hash = "${filesha256(var.genesyscloud_flow_INQUEUECALL_inqueue_filepath)}"
  filepath          = "${var.genesyscloud_flow_INQUEUECALL_inqueue_filepath}"
}

resource "genesyscloud_flow" "INQUEUECALL_Default_In-Queue_Flow" {
  file_content_hash = "${filesha256(var.genesyscloud_flow_INQUEUECALL_Default_In-Queue_Flow_filepath)}"
  filepath          = "${var.genesyscloud_flow_INQUEUECALL_Default_In-Queue_Flow_filepath}"
}

resource "genesyscloud_flow" "VOICEMAIL_Default_Voicemail_Flow" {
  file_content_hash = "${filesha256(var.genesyscloud_flow_VOICEMAIL_Default_Voicemail_Flow_filepath)}"
  filepath          = "${var.genesyscloud_flow_VOICEMAIL_Default_Voicemail_Flow_filepath}"
}

variable "genesyscloud_flow_INQUEUECALL_inqueue_filepath" {
  description = "YAML file path for flow configuration. Note: Changing the flow name will result in the creation of a new flow with a new GUID, while the original flow will persist in your org."
}
variable "genesyscloud_flow_INQUEUECALL_Default_In-Queue_Flow_filepath" {
  description = "YAML file path for flow configuration. Note: Changing the flow name will result in the creation of a new flow with a new GUID, while the original flow will persist in your org."
}
variable "genesyscloud_flow_VOICEMAIL_Default_Voicemail_Flow_filepath" {
  description = "YAML file path for flow configuration. Note: Changing the flow name will result in the creation of a new flow with a new GUID, while the original flow will persist in your org."
}
```

PR #821, opened 2024-01-26
DEVTOOLING-355: Underscore in flow names prevent accurate assignment of the flow config yaml with the terraform resource

willjeuniaux-genesys:
Ticket: https://inindca.atlassian.net/browse/DEVTOOLING-355

Here is an example of files structure after downloading the 4 flows described on the ticket using Archy:

```
./flows/commonModule/custom flow_v1-0.yaml
./flows/commonModule/custom_flow_v1-0.yaml
./flows/inboundCall/custom flow_v1-0.yaml
./flows/inboundCall/custom_flow_v1-0.yaml
```

In order to programmatically map each file to the correct terraform resource in a convenient way, we need to make modifications to include the flow type & name as attributes of the genesyscloud_flow resource. Below is the new proposed format.

genesyscloud.tf:

```
resource "genesyscloud_flow" "custom_flow_1141841684" {
  file_content_hash = "${filesha256(var.genesyscloud_flow_custom_flow_1141841684_filepath)}"
  filepath          = "${var.genesyscloud_flow_custom_flow_1141841684_filepath}"
  flow_name         = "custom flow"
  flow_type         = "INBOUNDCALL"
}

resource "genesyscloud_flow" "custom_flow" {
  file_content_hash = "${filesha256(var.genesyscloud_flow_custom_flow_filepath)}"
  filepath          = "${var.genesyscloud_flow_custom_flow_filepath}"
  flow_name         = "custom_flow"
  flow_type         = "INBOUNDCALL"
}

resource "genesyscloud_flow" "custom_flow_1141841684_723871156" {
  file_content_hash = "${filesha256(var.genesyscloud_flow_custom_flow_1141841684_723871156_filepath)}"
  filepath          = "${var.genesyscloud_flow_custom_flow_1141841684_723871156_filepath}"
  flow_name         = "custom flow"
  flow_type         = "COMMONMODULE"
}

resource "genesyscloud_flow" "custom_flow_4191204515" {
  file_content_hash = "${filesha256(var.genesyscloud_flow_custom_flow_4191204515_filepath)}"
  filepath          = "${var.genesyscloud_flow_custom_flow_4191204515_filepath}"
  flow_name         = "custom_flow"
  flow_type         = "COMMONMODULE"
}
```

terraform.tfvars:

```
genesyscloud_flow_custom_flow_1141841684_filepath           = "./flows/inboundCall/custom flow_v1-0.yaml"
genesyscloud_flow_custom_flow_filepath                      = "./flows/inboundCall/custom_flow_v1-0.yaml"
genesyscloud_flow_custom_flow_1141841684_723871156_filepath = "./flows/commonModule/custom flow_v1-0.yaml"
genesyscloud_flow_custom_flow_4191204515_filepath           = "./flows/commonModule/custom_flow_v1-0.yaml"
```

Considerations:

- The new attributes are currently being marked as Optional in the Schema because I wasn't sure if marking them as Required wouldn't break any acceptance tests which I've not been exposed to yet. However since it is not technically required to create a flow, Optional seemed fine?
- The flow_name & flow_type attributes intent is to be read-only. There is no expectation that the value set on the resource would overwrite what is within the yaml file for that flow during the creation process. Should the attribute name be suffixed by "readonly" to show the intent better?

----

After exporting the resources, the IDE currently highlights both flow_name & flow_type and specifies the following message "Unexpected attribute: An attribute named "flow_name/flow_type" is not expected here". I think I modified the resource Schema correctly. The export worked fine and the following terraform apply was a success too. What I am missing here?

Thanks,
Will


HemanthDogiparthi12:
> "Unexpected attribute: An attribute named "flow_name/flow_type" is not expected here".

If your export and tf apply after that is working fine. I dont see any reason. Can you elaborate where you are seeing this error in IDE? is it warning markup shown by the IDE ?

genesyscloud/resource_genesyscloud_flow.go
```diff
@@ -92,6 +92,16 @@ func ResourceFlow() *schema.Resource {
...
+			"flow_name": {
+				Description: `Genesys Cloud flow name`,
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"flow_type": {
+				Description: `Genesys Cloud flow type`,
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
```
charliecon, line 104: I would mark this as Computed. If the user provides no value and one is read into the schema in the read function, they should see an error.

genesyscloud/resource_genesyscloud_flow.go
```diff
@@ -92,6 +92,16 @@ func ResourceFlow() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"flow_name": {
+				Description: `Genesys Cloud flow name`,
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
```
charliecon, line 99: Same here. `Computed: true`

charliecon: I have another concern about this field. If the end user wants to use it to define the name of the flow (which they will assume is possible when they see the field) , they might try to do so even though the flow name in the yaml file is different. That should also cause errors.

One possible workaround: You could specify in the docs that the name has to match the name field in the yaml file, and do some validation to ensure that is the case. There is also the possibility that they are using the substitutions map to populate the name field, in which case you would need to check inside that map, find the key, and again verify the two names are the same.

charliecon: (Same for flow_type)

willjeuniaux-genesys:
> You'll need to run `go generate` to update flow docs

i can't run go generate from a fork it looks like

data source entitled "genesyscloud-fork", or "genesyscloud-fork_genesyscloud_architect_datatable" does not exist
Error executing command: unable to generate website: error rendering static website: unable to render templated website to static markdown: unable to render template "data-sources\\genesyscloud_architect_datatable.md.tmpl": unable to execute template: template: docTemplate:6:18: executing "docTemplate" at <plainmarkdown>: invalid value; expected string

genesyscloud/resource_genesyscloud_flow.go
charliecon: That is do-able since the exported contents is guaranteed to have the correct flow name and type values in the resource... but then the problem is that the exported flow file could be edited by the user to change the name. In that case, the user will have to remember to change the resource flow_name field before applying the export config.

charliecon:
> > You'll need to run `go generate` to update flow docs
> 
> It updated a bunch of readme files but ended erroring out at the end with this:
> 
> data source entitled "genesyscloud-fork", or "genesyscloud-fork_genesyscloud_architect_datatable" does not exist Error executing command: unable to generate website: error rendering static website: unable to render templated website to static markdown: unable to render template "data-sources\genesyscloud_architect_datatable.md.tmpl": unable to execute template: template: docTemplate:6:18: executing "docTemplate" at : invalid value; expected string

That's strange. No worries though. We can do that part if it's not working

willjeuniaux-genesys:
> > > You'll need to run `go generate` to update flow docs
> > 
> > 
> > It updated a bunch of readme files but ended erroring out at the end with this:
> > data source entitled "genesyscloud-fork", or "genesyscloud-fork_genesyscloud_architect_datatable" does not exist Error executing command: unable to generate website: error rendering static website: unable to render templated website to static markdown: unable to render template "data-sources\genesyscloud_architect_datatable.md.tmpl": unable to execute template: template: docTemplate:6:18: executing "docTemplate" at : invalid value; expected string
> 
> That's strange. No worries though. We can do that part if it's not working

Updated my reply above in the meanwhile, it actually deleted the readme files, ;) yep that sounds like a plan, thanks!

genesyscloud/resource_genesyscloud_flow.go
bbbco: @HemanthDogiparthi12 I submitted a PR last week to update the resource name with the flowtype as you mentioned: https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/820

HemanthDogiparthi12: Yes. Thanks @bbbco 
Hi @williamjeuniaux Based on the updated resource names , do you think we still need flow name and flow type attributes ?

willjeuniaux-genesys: Hi @HemanthDogiparthi12 

PR 820 doesn't solve the name problem.

What special character do you recommend using? From what I can gather, it looks like it may not be best practices and could lead to issues while referencing the resource. I also checked if GC wasn't allowing hyphen "-", but they do, so we can't replace the space by it.

I do understand this might create ambiguity without the proper documentation. I started implementing changes following Charlie's comment about documenting the behavior and doing the necessary validation to make sure the values are the same as the ones in the yaml. If you think of a better way, I'll be happy to change it of course, let me know.

The flow type attribute is a nice to have, since the type is already in the name of the resource. However, from a structural perspective, it appears more logical to incorporate the type as an attribute of the flow object.

willjeuniaux-genesys:
I updated the code based on the comments. Will add a unit test if the approach is validated, thanks!

kavinbalagen (approved):

willjeuniaux-genesys:
@carnellj-genesys @kavinbalagen @charliecon @HemanthDogiparthi12 
I added unit tests

PR #822, opened 2024-01-26
DEVTOOLING-344: error http 412 when creating Responsemanagement Response

willjeuniaux-genesys:
Ticket: https://inindca.atlassian.net/browse/DEVTOOLING-344

Fix: Implemented retries when creating the entity

Test:
I did reapply the same config that previously generated an error, did not encounter an error this time. Therefore I can validate that there is nothing wrong with the resources. It was definitely a random issue with GC at that time.


charliecon (approved):

carnellj-genesys:
Want to check with the API owner and figure out why they are returning a 412 and whether it is appropriate to retry

bbbco:
https://developer.genesys.cloud/organization/responsemanagement-apis#post-api-v2-responsemanagement-responses 
> 412 - Precondition Failed
Not sure retrying would fix this type of error, right?

willjeuniaux-genesys:
> https://developer.genesys.cloud/organization/responsemanagement-apis#post-api-v2-responsemanagement-responses
> 
> > 412 - Precondition Failed
> 
> Not sure retrying would fix this type of error, right?

Well, the error message indicates it is unable to process the request at this time. Which typically means there is nothing wrong with the payload and something else is wrong on the server. The error code seems indeed odd though. Out of all the other responsemanagement responses that tried to be created on that run, only this one failed. When I tried creating this one on my own server, no errors at all, so based on my testing I do believe a retry would have fixed it.

willjeuniaux-genesys:
> Want to check with the API owner and figure out why they are returning a 412 and whether it is appropriate to retry

Agreed, they should be able to provide information about what happened with the collerationID in the ticket.

genesyscloud/resource_genesyscloud_responsemanagement_response.go
```diff
@@ -275,14 +275,24 @@ func createResponsemanagementResponse(ctx context.Context, d *schema.ResourceDat
...
 
-	log.Printf("Creating Responsemanagement Response %s", name)
-	responsemanagementResponse, _, err := responseManagementApi.PostResponsemanagementResponses(sdkresponse, "")
-	if err != nil {
-		return diag.Errorf("Failed to create Responsemanagement Response %s: %s", name, err)
+	diagErr := WithRetries(ctx, 30*time.Second, func() *retry.RetryError {
+		log.Printf("Creating Responsemanagement Response %s", name)
+		responsemanagementResponse, resp, err := responseManagementApi.PostResponsemanagementResponses(sdkresponse, "")
+		if err != nil {
+			if IsStatus412(resp) {
```
carnellj-genesys, line 282: @dginty4 @charliecon Do you guys know who the API owners of Response Management.  PR looks good, but I am wondering why they are returning a 412 and whether it is appropriate to retry on a 412.

carnellj-genesys (approved):

HemanthDogiparthi12 (approved):

PR #827, opened 2024-01-29
Feat/direct routing remove requirement of backup queue 

jonesb-genesys:
Follow-up on a previous PR. Turns out I did not know Golang supplies a "" for an empty string and does not allow passing a null string. Made the change to support that, added computed to the field as the backend will use the current queue id as backup if none is supplied, and added a test for the situation. Also updated the docs using go generate


genesyscloud/resource_genesyscloud_routing_queue.go
```diff
@@ -1308,6 +1309,16 @@ func buildSdkDirectRouting(d *schema.ResourceData) *platformclientv2.Directrouti
 			UseAgentAddressOutbound: &messageUseAgentAddressOutbound,
 		}
 
+		if backupQueueID == "" || backupQueueID == NullValue {
```
dginty4, line 1312: I don't think your second condition is needed. `NullValue` is a string with a value of null so `backupQueueID == NullValue` is checking if `backupQueueID == "null"`. I don't know a situation this would be used since if `backup_queue_id` was not supplied `backupQueueID` would be an empty string

genesyscloud/resource_genesyscloud_routing_queue.go
```diff
@@ -1308,6 +1309,16 @@ func buildSdkDirectRouting(d *schema.ResourceData) *platformclientv2.Directrouti
...
+			return &platformclientv2.Directrouting{
+				CallMediaSettings:    callSettings,
+				EmailMediaSettings:   emailSettings,
+				MessageMediaSettings: messageSettings,
+				WaitForAgent:         &waitForAgent,
+				AgentWaitSeconds:     &agentWaitSeconds,
+			}
+		}
+
 		return &platformclientv2.Directrouting{
```
charliecon, line 1322: If settingsMap["backup_queue_id"] could be nil, I would recommend defining the Directrouting object as a variable at the beginning, and setting the backup queue id like this:

```golang
if backUpQueueId, ok := settingsMap["backup_queue_id"].(string); ok && backUpQueueId != "" {
  directRouting.BackupQueueId = &backUpQueueId
}
```

It is more readable than having the two returns statements at the bottom, one having one more field than the other

HemanthDogiparthi12 (approved):

jonesb-genesys:
Awesome thanks for the reviews and the improvement suggestions!

jonesb-genesys:
Anything else needed here? Thanks!

carnellj-genesys (approved):

PR #846, opened 2024-02-12
Devtooling 374

kavinbalagen:


genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (p *edgeGroupProxy) deleteEdgeGroup(ctx context.Context, edgeGroupId string) (*platformclientv2.APIResponse, error) {
+	return p.deleteEdgeGroupAttr(ctx, p, edgeGroupId)
+}
+
+func (p *edgeGroupProxy) putEdgeGroup(ctx context.Context, edgeGroupId string, body platformclientv2.Edgegroup) (*platformclientv2.Edgegroup, *platformclientv2.APIResponse, error) {
+	return p.putEdgeGroupAttr(ctx, p, edgeGroupId, body)
+}
+
+func (p *edgeGroupProxy) postEdgeGroup(ctx context.Context, body platformclientv2.Edgegroup) (*platformclientv2.Edgegroup, *platformclientv2.APIResponse, error) {
```
dginty4, line 62: we generally call the post functions `createEdgeGroup` instead, same with `putEdgeGroup` should be update

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+	return edgeGroup, resp, nil
+}
+
+func getAllEdgeGroupsFn(ctx context.Context, p *edgeGroupProxy) (*[]platformclientv2.Edgegroup, error) {
+	const pageSize = 100
+	var allEdgeGroups []platformclientv2.Edgegroup
+	for pageNum := 1; ; pageNum++ {
```
dginty4, line 102: Instead of starting with the first page, call the first page separately then use the `pageCount` returned by that api call as the condition in the for loop

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func getEdgeGroupProxy(clientConfig *platformclientv2.Configuration) *edgeGroupProxy {
+	if internalProxy == nil {
+		internalProxy = newEdgeGroupProxy(clientConfig)
+	}
+	return internalProxy
+}
+
+func (p *edgeGroupProxy) getEdgeGroup(ctx context.Context, edgeGroupId string) (*platformclientv2.Edgegroup, *platformclientv2.APIResponse, error) {
```
dginty4, line 50: this should be `getEdgeGroupById`

genesyscloud/telephony_providers_edges_extension_pool/genesyscloud_telephony_providers_edges_extension_pool_proxy.go
```diff
@@ -0,0 +1,125 @@
+package telephony_providers_edges_extension_pool
```
dginty4, line 1: My comments for edge groups apply here too

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
kavinbalagen: addressed

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
kavinbalagen: addressed

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
kavinbalagen: addressed

genesyscloud/telephony_providers_edges_extension_pool/genesyscloud_telephony_providers_edges_extension_pool_proxy.go
kavinbalagen: addressed

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	if edgeGroups.Entities != nil && len(*edgeGroups.Entities) > 0 {
+		for _, edgeGroup := range *edgeGroups.Entities {
+			if edgeGroup.State != nil && *edgeGroup.State != "deleted" {
+				allEdgeGroups = append(allEdgeGroups, edgeGroup)
+			}
+		}
+	}
+
+	if *edgeGroups.PageCount < 2 {
```
dginty4, line 153: You don't need this if statement, if there's no page then the loop won't run

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (p *edgeGroupProxy) getEdgeGroupByName(ctx context.Context, edgeGroupName string, managed bool) (*[]platformclientv2.Edgegroup, error) {
+	return p.getEdgeGroupByNameAttr(ctx, p, edgeGroupName, managed)
+}
+
+func getEdgeGroupByNameFn(ctx context.Context, p *edgeGroupProxy, edgeGroupName string, managed bool) (*[]platformclientv2.Edgegroup, error) {
+	const pageSize = 100
+	var pageNum = 1
+	var allEdgeGroups []platformclientv2.Edgegroup
```
dginty4, line 80: There shouldn't be an array, we should only return one value, the object with the name that matches `name`

kavinbalagen: that didn't work apparently. the test cases kept failing, it said it wasn't found

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
```diff
@@ -0,0 +1,172 @@
...
+
+	edgeGroups, _, err := p.edgesApi.GetTelephonyProvidersEdgesEdgegroups(pageSize, pageNum, edgeGroupName, "", managed)
```
dginty4, line 82: You don't need to repeat this looping logic, calling `getAllEdgeGroupsFn` will return all the egeGroups

kavinbalagen: actually, we're passing the edgegroupName and managed parameter here. so this is slightly different

genesyscloud/telephony_providers_edges_edge_group/genesyscloud_telephony_providers_edges_edge_group_proxy.go
kavinbalagen: addressed

genesyscloud/telephony_providers_edges_edge_group/resource_genesyscloud_telephony_providers_edges_edge_group.go
```diff
@@ -172,44 +172,17 @@ func readEdgeGroup(ctx context.Context, d *schema.ResourceData, meta interface{}
...
-
-		if edgeGroups.Entities == nil || len(*edgeGroups.Entities) == 0 {
-			break
-		}
-
-		for _, edgeGroup := range *edgeGroups.Entities {
-			if edgeGroup.State != nil && *edgeGroup.State != "deleted" {
-				resources[*edgeGroup.Id] = &resourceExporter.ResourceMeta{Name: *edgeGroup.Name}
-			}
+	if edgeGroups != nil {
```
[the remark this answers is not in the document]
kavinbalagen: wouldn't *edgeGroups on nil cause any errors?

PR #856, opened 2024-02-15
Replace Data Source

HemanthDogiparthi12:


genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -65,6 +65,15 @@ func ResourceTfExport() *schema.Resource {
 				ForceNew:      true,
 				ConflictsWith: []string{"resource_types", "exclude_filter_resources"},
 			},
+			"replace_with_datasource": {
+				Description: "Include only resources that match either a resource type or a resource type::regular expression.  See export guide for additional information",
```
dginty4, line 69: I think we should add a line to the start here to explain what this property is a bit more. Something like, `Resources to be exported as a data source`

dginty4 (approved):

kavinbalagen (approved):

PR #883, opened 2024-02-27
DEVTOOLING-421: Removed the code from DEVTOOLING-310 and DEVTOOLING-2…

carnellj-genesys:
…82.  After talking with the integrations team we realized this code does not help anything and needs to be removed

### DIFF
--- a/genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
+++ b/genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"regexp"
-	"strings"
 	"terraform-provider-genesyscloud/genesyscloud/provider"
 	"terraform-provider-genesyscloud/genesyscloud/util"
 	"time"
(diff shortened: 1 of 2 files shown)


genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
```diff
@@ -54,24 +52,6 @@ func getAllCredentials(ctx context.Context, clientConfig *platformclientv2.Confi
 	for _, cred := range *credentials {
 		log.Printf("Dealing with credential id : %s", *cred.Id)
 		if cred.Name != nil { // Credential is possible to have no name
-
```
carnellj-genesys, line 57: William originally added this code because we were trying to leverage the naming standards for integrations to go figure out the integration credential associated with the integration.  However, after chatting with the integration team they let us know that this is not possible.  So we need to remove the code below because it does not add value.

dginty4 (approved):

charliecon (approved):

HemanthDogiparthi12 (approved):

PR #891, opened 2024-03-01
DEVTOOLING-425: Retry logic on flow upload 

charliecon:


kavinbalagen (approved):

genesyscloud/architect_flow/resource_genesyscloud_flow.go
```diff
@@ -113,10 +113,31 @@ func updateFlow(ctx context.Context, d *schema.ResourceData, meta interface{}) d
...
+	uploadErr := util.WithRetries(ctx, 20*time.Second, func() *retry.RetryError {
+		uploadStartTime := time.Now()
+		_, err = s3Uploader.Upload()
+		uploadDuration := time.Since(uploadStartTime)
+		if err == nil {
+			return nil
+		}
+		// Retrying when encountering the error from DEVTOOLING-425
+		if strings.Contains(fmt.Sprintf("%v", err), "An existing connection was forcibly closed by the remote host") {
+			log.Printf("failed to upload file %s after %d nanoseconds (%v seconds). Error: %v", filePath, uploadDuration.Nanoseconds(), uploadDuration.Seconds(), err)
```
carnellj-genesys, line 130: If anything I would record milliseconds and seconds.  Not nanoseconds.  A network upload is almost never going to take nanoseconds.

genesyscloud/architect_flow/resource_genesyscloud_flow.go
```diff
@@ -113,10 +113,31 @@ func updateFlow(ctx context.Context, d *schema.ResourceData, meta interface{}) d
...
+
+	uploadErr := util.WithRetries(ctx, 20*time.Second, func() *retry.RetryError {
+		uploadStartTime := time.Now()
+		_, err = s3Uploader.Upload()
+		uploadDuration := time.Since(uploadStartTime)
+		if err == nil {
+			return nil
+		}
+		// Retrying when encountering the error from DEVTOOLING-425
+		if strings.Contains(fmt.Sprintf("%v", err), "An existing connection was forcibly closed by the remote host") {
```
carnellj-genesys, line 129: I am just wondering if we should retry on any error?  It seems like if we have any kind of problem we should retry

carnellj-genesys (approved):
ook good.  Just have one small

genesyscloud/architect_flow/resource_genesyscloud_flow.go
charliecon: I'm going to change it because there is also a chance aws could change the error message they return, then this code wouldn't work anymore

genesyscloud/architect_flow/resource_genesyscloud_flow.go
charliecon: Done

genesyscloud/architect_flow/resource_genesyscloud_flow.go
charliecon: Done

genesyscloud/architect_flow/resource_genesyscloud_flow.go
```diff
@@ -113,10 +113,27 @@ func updateFlow(ctx context.Context, d *schema.ResourceData, meta interface{}) d
...
-	_, err = s3Uploader.Upload()
+	fileInfo, err := os.Stat(filePath)
 	if err != nil {
+		log.Printf("failed to read file information. Path: '%s' Error: %v", filePath, err)
+	}
+
+	uploadErr := util.WithRetries(ctx, 20*time.Second, func() *retry.RetryError {
+		uploadStartTime := time.Now()
+		_, err = s3Uploader.Upload()
+		uploadDuration := time.Since(uploadStartTime)
```
HemanthDogiparthi12, line 124: minor one : move this to inside error block

charliecon: Done

HemanthDogiparthi12 (approved):

PR #899, opened 2024-03-05
DEVTOOLING-438: Adding a data source to look up workitem status ids.

carnellj-genesys:


genesyscloud/task_management_workbin/data_source_genesyscloud_task_management_workbin.go
```diff
@@ -21,7 +21,7 @@ import (
 // dataSourceTaskManagementWorkbinRead retrieves by name the id in question
 func dataSourceTaskManagementWorkbinRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
-	proxy := newTaskManagementWorkbinProxy(sdkConfig)
+	proxy := getTaskManagementWorkbinProxy(sdkConfig)
```
carnellj-genesys, line 24: Found a few spots where Prince was use doing a new to get the proxy instead of a getXXX

genesyscloud/task_management_worktype/data_source_genesyscloud_task_management_unit_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v121/platformclientv2"
+	"github.com/stretchr/testify/assert"
+	"terraform-provider-genesyscloud/genesyscloud/provider"
+	"testing"
+)
+
+// TestAccDataSourceTaskManagementWorktypeStatus tests the retrieve of a status id.
```
carnellj-genesys, line 13: Decided to make this a unit test because I already have a test that covers almost all of this test.

genesyscloud/task_management_worktype/data_source_genesyscloud_task_management_worktype.go
```diff
@@ -40,3 +40,32 @@ func dataSourceTaskManagementWorktypeRead(ctx context.Context, d *schema.Resourc
 		return nil
 	})
 }
+
+func dataSourceTaskManagementWorktypeStatusRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
carnellj-genesys, line 44: This is the new data source

genesyscloud/task_management_worktype/genesyscloud_task_management_worktype_proxy.go
```diff
@@ -158,8 +165,8 @@ func getAllTaskManagementWorktypeFn(ctx context.Context, p *taskManagementWorkty
 	return &allWorktypes, nil
 }
 
-// getTaskManagementWorktypeIdByNameFn is an implementation of the function to get a Genesys Cloud task management worktype by name
-func getTaskManagementWorktypeIdByNameFn(ctx context.Context, p *taskManagementWorktypeProxy, name string) (id string, retryable bool, err error) {
+// Looksup a worktype by name
+func getWorkType(name string, p *taskManagementWorktypeProxy) (*platformclientv2.Worktype, bool, error) {
```
carnellj-genesys, line 169: The two data sources do almost the same actions.  The difference is that the data original data source we retrieved only the work type id.  The new data source retrieves the work type and then cycles through to match on the status.  Since they are doing the same API calls under the covers, I  refactored the calls out to our API to be in a single shared method called getWorkType.

genesyscloud/task_management_worktype/data_source_genesyscloud_task_management_unit_test.go
```diff
@@ -0,0 +1,73 @@
...
+// I am writing a unit test here as we already have the basic test coverage in another test.
+func TestAccDataSourceTaskManagementWorktypeStatus(t *testing.T) {
```
charliecon, line 15: This should be TestUnitDataSource...

genesyscloud/task_management_worktype/data_source_genesyscloud_task_management_worktype.go
```diff
@@ -40,3 +40,32 @@ func dataSourceTaskManagementWorktypeRead(ctx context.Context, d *schema.Resourc
...
+			return retry.RetryableError(fmt.Errorf("no task management worktype found with name %s", workTypeName))
+		}
+
+		for _, status := range *workType.Statuses {
+			if *status.Name == statusName {
+				d.SetId(*status.Id)
+				return nil
+			}
+		}
+		return nil
```
charliecon, line 69: Shouldn't we return an error down here if there is no match on the status name?

genesyscloud/task_management_worktype/data_source_genesyscloud_task_management_unit_test.go
carnellj-genesys: Thank you!

charliecon (approved):
`go generate` needs to be ran again. Looks good otherwise

PR #912, opened 2024-03-13
Devtooling 287 cache sub branch

HemanthDogiparthi12:
Introducing the Cache Solution by @carnellj  for Data Table and DataTable Rows to start with

and added new custom Generic Methods in resource_cache so that it will be only a method call while setting and getting cache of each individual resource

added new test cases of the methods introduced.


dginty4:
Is this a different approach to Johns? Also you might need to rebase since I upgraded the go sdk

HemanthDogiparthi12:
> Is this a different approach to Johns or an extention of his? Also you might need to rebase since I upgraded the go sdk

Hi @dginty4 Its an extension. The rebase is done now.

genesyscloud/tfexporter_state/tfexporter_state.go
```diff
@@ -0,0 +1,32 @@
...
+a signal that any resources being exported should be reading their data from each resource's internal cache rather then the API.
+*/
+var exportState bool
+var once sync.Once
+
+// ActivateExporterState will be used to indicate that caching should be used to process requests.
+// We are setting this as an environment variable so we can experiment with it, without creating an attribute
+// on the resource
+func ActivateExporterState() {
+	cxCacheActive := os.Getenv("ENABLE_CX_CACHE")
```
carnellj-genesys, line 20: We need to refactor this.  We should not have an environment variable here as I put here only to be able turn caching on or if.  Make sure you remove the if cxCacheActive check and just leave the once.Do call

genesyscloud/resource_cache/resource_cache_test.go
```diff
@@ -0,0 +1,47 @@
+package resource_cache
+
+import (
+	"os"
+	"terraform-provider-genesyscloud/genesyscloud/tfexporter_state"
+	"testing"
+)
+
+func TestUnitWithoutExporterState(t *testing.T) {
+	os.Setenv("ENABLE_CX_CACHE", "")
```
carnellj-genesys, line 10: This needs to be removed

genesyscloud/resource_cache/resource_cache_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Test GetCache for non-existent key
+	valPtr = GetCache(cache, "nonexistent")
+	if valPtr != nil {
+		t.Errorf("Expected nil value from the Cache")
+	}
+}
+
+func TestUnitSetCacheAndGetCache(t *testing.T) {
+	os.Setenv("ENABLE_CX_CACHE", "true")
```
carnellj-genesys, line 30: This needs to be removed

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row_proxy.go
```diff
@@ -99,6 +107,7 @@ func getAllArchitectDatatableFn(ctx context.Context, p *architectDatatableRowPro
 
 	for _, table := range *tables.Entities {
 		totalRecords = append(totalRecords, table)
+		rc.SetCache(p.dataTableCache, *table.Id, *ConvertDatatable(table))
```
carnellj-genesys, line 110: Just out of curiosity.  Why are you passing in the cache into this method?  I think on the original example we had the happening directly on the cache.

carnellj-genesys (changes requested):
Overall looks really good.  Please review and make changes.   We can chat in the morning and determine if we are ready for release.

genesyscloud/tfexporter_state/tfexporter_state.go
HemanthDogiparthi12: removed the env check

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row_proxy.go
```diff
@@ -113,13 +122,29 @@ func getAllArchitectDatatableFn(ctx context.Context, p *architectDatatableRowPro
...
 
 	return &totalRecords, nil
 }
 
+func ConvertDatatable(master platformclientv2.Datatable) *Datatable {
+	var datatable Datatable
+	err := mapstructure.Decode(master, &datatable)
+	if err != nil {
+		return nil
+	}
```
charliecon, line 137: Are we losing some info here by not handling the error? I think we should at least log the error, even if it shouldn't cause the process to fail

go.mod
```diff
@@ -1,6 +1,8 @@
 module terraform-provider-genesyscloud
 
-go 1.20
+go 1.21
+
+toolchain go1.21.4
```
charliecon, line 5: This is happening because of the go version you're using. If you use go1.20 (make sure GOROOT in your Goland settings is pointing to 1.20) then this won't automatically upgrade to 1.21

charliecon (approved):

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row_proxy.go
HemanthDogiparthi12: Just added a wrapper method , 

it removes the dependency of tfexporter from all the proxies.
also removes few boiler plate code like exporter checks and added few log statements added when there is a cache miss.

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row_proxy.go
HemanthDogiparthi12: edgecase where platformclientv2.Datatable cannot be converted to our custom DataTable. I am returning null in this scenario and the cache miss happens during getCache and the export proceeds to do an API call. Will add a log statement here.

carnellj-genesys (approved):

PR #927, opened 2024-03-21
Devtooling 485 fix group delete v2

carnellj-genesys:


genesyscloud/group/data_source_genesyscloud_group.go
```diff
@@ -11,47 +11,26 @@ import (
...
 	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
-		groups, _, getErr := groupsAPI.PostGroupsSearch(platformclientv2.Groupsearchrequest{
-			Query: &[]platformclientv2.Groupsearchcriteria{searchCriteria},
-		})
+		groups, _, getErr := gp.getGroupsByName(ctx, nameStr)
 		if getErr != nil {
 			return retry.NonRetryableError(fmt.Errorf("Error requesting group %s: %s", nameStr, getErr))
 		}
 
+		if *groups.Total > 1 {
```
carnellj-genesys, line 28: I added this.  This was a potential bug because if we ever got back more then one group we would always grab the first one and this could be wrong

genesyscloud/group/genesyscloud_group_proxy.go
```diff
@@ -0,0 +1,185 @@
...
+
+	for pageNum := 2; pageNum <= *groups.PageCount; pageNum++ {
+		groups, resp, getErr := p.groupsApi.GetGroups(pageSize, pageNum, nil, nil, "")
+		if getErr != nil {
+			return nil, resp, fmt.Errorf("Failed to get page of groups: %v", getErr)
+		}
+		allGroups = append(allGroups, *groups.Entities...)
+	}
+
+	for _, group := range allGroups {
```
carnellj-genesys, line 180: @HemanthDogiparthi12 Added caching to this code.  Nice job on the caching, very clean

genesyscloud/group/genesyscloud_group_proxy.go
```diff
@@ -0,0 +1,185 @@
...
+func updateGroupFn(ctx context.Context, p *groupProxy, id string, group *platformclientv2.Groupupdate) (*platformclientv2.Group, *platformclientv2.APIResponse, error) {
+	return p.groupsApi.PutGroup(id, *group)
+}
+
+func deleteGroupFn(ctx context.Context, p *groupProxy, id string) (*platformclientv2.APIResponse, error) {
+	return p.groupsApi.DeleteGroup(id)
+}
+
+func getGroupByIdFn(ctx context.Context, p *groupProxy, id string) (*platformclientv2.Group, *platformclientv2.APIResponse, error) {
+	group := rc.GetCache(p.groupCache, id)
```
carnellj-genesys, line 113: mmmmm...... Caching ....... Best homer slobbering

genesyscloud/group/resource_genesyscloud_group.go
```diff
@@ -0,0 +1,333 @@
...
+			return retry.NonRetryableError(fmt.Errorf("Error deleting group %s: %s", d.Id(), err))
+		}
+
+		if group.State != nil && *group.State == "deleted" {
+			log.Printf("Group %s deleted", name)
+			return nil
+		}
+
+		/*
+		  This extra delete call is being added here because of  DEVTOOLING-485.  Basically we are in a transition
```
carnellj-genesys, line 219: This was the bug fix that started this whole thing.  I am adding an extra delete, because sometimes when we delete a record directory is quietly failing the delete so if we find the record after the delete, we are going to retry

genesyscloud/group/resource_genesyscloud_group.go
```diff
@@ -0,0 +1,333 @@
...
+			maxMembersPerRequest := 50
+			membersToRemoveList := lists.SliceDifference(existingMemberIds, configMemberIds)
+			chunkedMemberIdsDelete := chunks.ChunkBy(membersToRemoveList, maxMembersPerRequest)
+
+			chunkProcessor := func(membersToRemove []string) diag.Diagnostics {
+				if len(membersToRemove) > 0 {
+					if diagErr := util.RetryWhen(util.IsVersionMismatch, func() (*platformclientv2.APIResponse, diag.Diagnostics) {
+						_, resp, err := gp.deleteGroupMembers(ctx, d.Id(), strings.Join(membersToRemove, ","))
+						if err != nil {
+							return resp, util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to remove members from group %s: %s", d.Id(), err), resp)
```
carnellj-genesys, line 255: As part of this work I went through and started using the diagnostics pattern to get the log format in the way we want.

genesyscloud/group/resource_genesyscloud_group_test.go
```diff
@@ -54,7 +55,7 @@ func TestAccResourceGroupBasic(t *testing.T) {
 			},
 			{
 				// Update group
-				Config: GenerateUserWithCustomAttrs(testUserResource, testUserEmail, testUserName) + generateGroupResource(
+				Config: generateUserWithCustomAttrs(testUserResource, testUserEmail, testUserName) + GenerateGroupResource(
```
carnellj-genesys, line 58: I had to duplicate functions to break cyclic dependencies.  As you look through the code you mind find duplicate functions across packages.  We will need to clean this up at some point and centralize into a single package, but for now I am just moving functions locally.

genesyscloud/group_roles/resource_genesyscloud_group_roles_schema.go
```diff
@@ -78,3 +81,27 @@ func GroupRolesExporter() *resourceExporter.ResourceExporter {
 		},
 	}
 }
+
```
carnellj-genesys, line 84: Probably should make this part of the proxy.  Thoughts

genesyscloud/resource_genesyscloud_orgauthorization_pairing_test.go
```diff
@@ -164,3 +165,41 @@ func TestAccResourceOrgAuthorizationPairing(t *testing.T) {
 		},
 	})
 }
+
+func generateBasicGroupResource(resourceID string, name string, nestedBlocks ...string) string {
```
carnellj-genesys, line 169: Duplicated code to break dependencies

genesyscloud/resource_genesyscloud_routing_queue_test.go
```diff
@@ -56,15 +57,15 @@ func TestAccResourceRoutingQueueBasic(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				// Create
-				Config: GenerateUserWithCustomAttrs(testUserResource, testUserEmail, testUserName) + GenerateRoutingSkillResource(queueSkillResource, queueSkillName) +
+				Config: generateUserWithCustomAttrs(testUserResource, testUserEmail, testUserName) + GenerateRoutingSkillResource(queueSkillResource, queueSkillName) +
 					generateGroupResource(
 						bullseyeMemberGroupName,
-						"MySeries6Group",
+						"MySeries6Groupv2",
```
carnellj-genesys, line 63: Changed this name because a crash caused this resource to hand around in the test environment

genesyscloud/group_roles/resource_genesyscloud_group_roles_schema.go
```diff
@@ -78,3 +81,27 @@ func GroupRolesExporter() *resourceExporter.ResourceExporter {
...
+//Duplicated this from the group roles package to break a cyclical dependency.  We should be asking ourselves why we are doing this at some point
```
charliecon, line 85: *from groups package (?)

genesyscloud/group_roles/resource_genesyscloud_group_roles_schema.go
charliecon: I agree unless we plan on refactoring it to another package instead of duplicating - in that case I don't think it's necessary to bake it into the proxy

PR #932, opened 2024-03-25
DEVTOOLING-492: Resolved nil interface in buildAppConversations function

BrianMoyles:


carnellj-genesys (approved):

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
```diff
@@ -79,9 +79,17 @@ func buildMessengerApps(apps []interface{}) *platformclientv2.Messengerapps {
 	}
 
 	app := apps[0].(map[string]interface{})
+	conversations, conversationsOk := app["conversations"].([]interface{})
+	knowledge, knowledgeOk := app["knowledge"].([]interface{})
+
+	if !conversationsOk || !knowledgeOk {
+		fmt.Println("error in build Messenger Apps")
+		return nil
+	}
```
charliecon, line 88: I don't think we should return in this case. If `conversations` and `knowledge` are optional fields, then the user should be able to omit them from their tf config, and then they'll appear as nil in the resource schema data and `knowledgeOk` will be `false`. 

I would check `if ok { messengerApp.Conversations = buildAppConversation(conversations) }`

carnellj-genesys: Nice catch Charlie

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
```diff
@@ -69,15 +69,13 @@ func buildAppKnowledge(knowledge []interface{}) *platformclientv2.Knowledge {
...
-
 	return ret
 }
 
 func buildMessengerApps(apps []interface{}) *platformclientv2.Messengerapps {
 	if len(apps) < 1 {
 		return nil
 	}
-
 	app := apps[0].(map[string]interface{})
```
charliecon, line 79: Same as with `buildAppKnowledge` above, the only item in apps could be nil, so we'll want to check for that with `if app, ok := apps[0].(map[string]interface{})`  

This can happen when all of the fields in the map are optional, so the end user could define messenger apps like this (even though they don't have any good reason to do so):

```hcl
messenger {
  apps {}
}
```

Because there are no fields, it doesn't register as a map and instead one nil item is added to the apps list.

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
```diff
@@ -69,15 +69,13 @@ func buildAppKnowledge(knowledge []interface{}) *platformclientv2.Knowledge {
...
 	return &platformclientv2.Messengerapps{
 		Conversations: buildAppConversations(app["conversations"].([]interface{})),
```
charliecon, line 81: If `app["conversations"]` is nil because the user didn't define it at all, then this line will throw the error `panic: interface conversion: interface {} is nil, not []interface {}`

So I would define your `&platformclientv2.Messengerapps` object as a variable `m`, set these fields conditionally 

```golang
if c, ok := app["conversations"].([]interface{}); ok {
  m.Conversations = c
}

// same thing for knowledge

return m
```

charliecon:
This file you're working on seems to have loads of lines that could throw interface conversion panics. It is probably worth going through the other functions and having a look at them. Otherwise we'll probably see another ticket pop up in the future pointing us back here

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
```diff
@@ -75,23 +74,22 @@ func buildAppKnowledge(knowledge []interface{}) *platformclientv2.Knowledge {
 }
 
 func buildMessengerApps(apps []interface{}) *platformclientv2.Messengerapps {
-	if len(apps) < 1 {
+	if len(apps) < 1 || (len(apps) == 1 && apps[0] == nil) {
 		return nil
 	}
 
+	m := platformclientv2.Messengerapps{}
```
charliecon, line 81: Apologies, I wrote `m` as the var name to keep the code example short. I would go with a more descriptive name here, like `messengerApps`

charliecon (approved):

PR #935, opened 2024-03-25
NO-JIRA:  Adding lint configuration

carnellj-genesys:


golangci.yaml
```diff
@@ -0,0 +1,24 @@
+linters:
```
carnellj-genesys, line 1: Added the linting configuration  for the tool and the ran it against one package to try it out.

genesyscloud/architect_datatable/resource_genesyscloud_architect_datatable_proxy.go
```diff
@@ -170,9 +168,7 @@ func getAllArchitectDatatableFn(ctx context.Context, p *architectDatatableProxy)
 			break
 		}
 
-		for _, table := range *tables.Entities {
-			totalRecords = append(totalRecords, table)
-		}
+		totalRecords = append(totalRecords, *tables.Entities...)
```
carnellj-genesys, line 171: Simplified the code for appending entities

genesyscloud/architect_datatable/resource_genesyscloud_architect_datatable_proxy.go
```diff
@@ -74,7 +74,7 @@ func createOrUpdateArchitectDatatableFn(ctx context.Context, p *architectDatatab
 	// create path and map variables
 	path := p.architectApi.Configuration.BasePath + "/api/v2/flows/datatables"
 
-	if createAction == false {
+	if !createAction {
```
carnellj-genesys, line 77: Fixed per comments from Linter

genesyscloud/architect_datatable/resource_genesyscloud_architect_datatable.go
```diff
@@ -115,19 +115,19 @@ func readArchitectDatatable(ctx context.Context, d *schema.ResourceData, meta in
...
 		}
 		cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, ResourceArchitectDatatable())
-		d.Set("name", *datatable.Name)
-		d.Set("division_id", *datatable.Division.Id)
+		_ = d.Set("name", *datatable.Name)
+		_ = d.Set("division_id", *datatable.Division.Id)
 
 		if datatable.Description != nil {
-			d.Set("description", *datatable.Description)
+			_ = d.Set("description", *datatable.Description)
```
carnellj-genesys, line 122: Explicitly showed I was not going to use the error code. This makes it easier for devs to realize an error code does get returning, but we are specifically not doing anything with it.

genesyscloud/architect_datatable/genesyscloud_architect_datatable_init_test.go
```diff
@@ -20,11 +20,15 @@ type registerTestInstance struct {
 
 // registerTestResources registers all resources used in the tests
 func (r *registerTestInstance) registerTestResources() {
+	r.resourceMapMutex.Lock()
```
carnellj-genesys, line 23: These were missed during the code review.  We should be locking the map so we avoid any potential concurrent map issues,

dginty4 (approved):

genesyscloud/architect_datatable/genesyscloud_architect_datatable_init_test.go
dginty4: Was it the linter that caught this?

carnellj-genesys: It caught that the variables were unused.   Once I knew they were unused, I realized that we need to include the locking in there.

PR #945, opened 2024-03-29
feat/Devtooling-448 oauth cred cache

carnellj-genesys:


genesyscloud/integration_credential/genesyscloud_integration_credential_init_test.go
```diff
@@ -31,6 +33,7 @@ func (r *registerTestInstance) registerTestResources() {
 	defer r.resourceMapMutex.Unlock()
 
 	providerResources["genesyscloud_integration_credential"] = ResourceIntegrationCredential()
+	providerResources["genesyscloud_oauth_client"] = oauth.ResourceOAuthClient()
```
carnellj-genesys, line 36: Adding a dependency for the genesys_oauth_client so we can test the situation where we are adding a Genesys Cloud OAuthClient

genesyscloud/integration_credential/genesyscloud_integration_credential_init_test.go
```diff
@@ -39,6 +42,7 @@ func (r *registerTestInstance) registerTestDataSources() {
 	defer r.datasourceMapMutex.Unlock()
 
 	providerDataSources["genesyscloud_integration_credential"] = DataSourceIntegrationCredential()
+	providerDataSources["genesyscloud_auth_role"] = auth_role.DataSourceAuthRole()
```
carnellj-genesys, line 45: Added a data source for an auth role look up

genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
```diff
@@ -82,16 +83,24 @@ func getAllCredentials(ctx context.Context, clientConfig *platformclientv2.Confi
 func createCredential(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	name := d.Get("name").(string)
 	cred_type := d.Get("credential_type_name").(string)
+	fields := buildCredentialFields(d)
+	_, secretFieldPresent := fields["clientSecret"]
```
carnellj-genesys, line 87: Checking to see if the secretField is set in the fields attribute.  If it is we are not going to attempt to read it later on from the cache.

genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
```diff
@@ -82,16 +83,24 @@ func getAllCredentials(ctx context.Context, clientConfig *platformclientv2.Confi
...
 
 	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
 	ip := getIntegrationCredsProxy(sdkConfig)
 
+	//If if is a Genesys Cloud OAuth Client and the user has not provided a secret field we should look for the
+	//item in the cache DEVTOOLING-448
+	if cred_type == "pureCloudOAuthClient" && !secretFieldPresent {
```
carnellj-genesys, line 94: Here is where we check to see if we are dealing with a Genesys Cloud AuthClient and whether the user has set the secret field has been set.  If the secretField has not been set, then the code will try to retrieve an OAuthClient  based on the client id from the the oauthClient proxy and see if there is already an OauthClient present in the code.

genesyscloud/integration_credential/resource_genesyscloud_integration_credential.go
```diff
@@ -104,6 +113,15 @@ func createCredential(ctx context.Context, d *schema.ResourceData, meta interfac
 	return readCredential(ctx, d, meta)
 }
 
+func retrieveCachedOauthClientSecret(sdkConfig *platformclientv2.Configuration, fields map[string]string) {
```
carnellj-genesys, line 116: If the clientId is in the field will retrieve the secret from the cached oAuthClient

genesyscloud/integration_credential/resource_genesyscloud_integration_credential_test.go
```diff
@@ -161,6 +161,74 @@ func TestAccResourceCredential(t *testing.T) {
 	})
 }
 
+// Resource Credential DEVTOOLING-448
+// This tests to make sure that we can successfully create an integration credential for a Genesys Cloud oauth client without providing a client secret
+func TestAccGenesysCloudOAuthResourceCredential(t *testing.T) {
```
carnellj-genesys, line 166: This test case will attempt to create a Genesys Cloud OAuth Client and then create an integration credential with only a clientId.  If it creates successfully, the test should pass.

genesyscloud/integration_credential/resource_genesyscloud_integration_credential_test.go
```diff
@@ -161,6 +161,74 @@ func TestAccResourceCredential(t *testing.T) {
...
+				),
+			},
+		},
+		CheckDestroy: testVerifyCredentialDestroyed,
+	})
+}
+
+// TestAccGenesysCloudOAuthResourceCredentialWithSecret will check to make sure we can still create Genesys Cloud
+// integration credential by providing the oauth client and id and secret.  This is how we would normally do it.
+func TestAccGenesysCloudOAuthResourceCredentialWithSecret(t *testing.T) {
```
carnellj-genesys, line 194: This function is going to make sure we can still create a Genesys Cloud integration credential and still be able to set the clientId and clientSecret.

genesyscloud/oauth_client/resource_genesyscloud_oauth_client_proxy.go
```diff
@@ -76,8 +92,24 @@ func (o *oauthClientProxy) getIntegrationCredential(ctx context.Context, id stri
...
 }
 
+func (o *oauthClientProxy) GetCachedOAuthClient(clientId string) platformclientv2.Oauthclient {
+	o.createdClientCacheLock.Lock()
+	defer o.createdClientCacheLock.Unlock()
+	return o.createdClientCache[clientId]
+}
+
 func (o *oauthClientProxy) createOAuthClient(ctx context.Context, oauthClient platformclientv2.Oauthclientrequest) (*platformclientv2.Oauthclient, *platformclientv2.APIResponse, error) {
-	return o.createOAuthClientAttr(ctx, o, oauthClient)
```
carnellj-genesys, line 80: When we create the OAuthClient we cache it.

charliecon (approved):
Should we use this new cache map for the oauth_client export as well?

genesyscloud/integration_credential/genesyscloud_integration_credential_init_test.go
```diff
@@ -31,6 +33,7 @@ func (r *registerTestInstance) registerTestResources() {
 	defer r.resourceMapMutex.Unlock()
```
carnellj-genesys, line 33: Added a defer to fix the lock

genesyscloud/oauth_client/resource_genesyscloud_oauth_client.go
```diff
@@ -49,15 +49,21 @@ func createOAuthClient(ctx context.Context, d *schema.ResourceData, meta interfa
...
-	oauthClientProxy := getOAuthClientProxy(sdkConfig)
+	oauthClientProxy := GetOAuthClientProxy(sdkConfig)
 
 	roles, diagErr := buildOAuthRoles(d)
 	if diagErr != nil {
 		return diagErr
 	}
 
+	//Before we create the oauth client take any roles on it and update the user
+	diagErr = updateTerraformUserWithRole(ctx, sdkConfig, roles)
```
carnellj-genesys, line 60: This is a new function.  Basically, if we create an oauth client we need to make sure any roles that were assigned to the oauth client are assigned to the user running the process.  This process will only work if the user is running user-based access token (e.g. like in CX accelerators.). If the user is running under an oauth credential and they do not have the role assigned to the oauth client, the code will fail with an error.

genesyscloud/oauth_client/resource_genesyscloud_oauth_client.go
```diff
@@ -93,18 +101,69 @@ func createOAuthClient(ctx context.Context, d *schema.ResourceData, meta interfa
...
+		_ = d.Set("integration_credential_id", *credential.Id)
+		_ = d.Set("integration_credential_name", *credential.Name)
 	}
 
 	d.SetId(*client.Id)
 	log.Printf("Created oauth client %s %s", name, *client.Id)
 	return readOAuthClient(ctx, d, meta)
 }
 
+func updateTerraformUserWithRole(ctx context.Context, sdkConfig *platformclientv2.Configuration, addedRoles *[]platformclientv2.Roledivision) diag.Diagnostics {
```
carnellj-genesys, line 113: This is the new function to update the role

genesyscloud/oauth_client/resource_genesyscloud_oauth_client.go
```diff
@@ -93,18 +101,97 @@ func createOAuthClient(ctx context.Context, d *schema.ResourceData, meta interfa
...
+
+	var totalRoles []string
+	//Step #4  - Concat the addedRoles
+	for _, role := range *addedRoles {
+		totalRoles = append(totalRoles, *role.RoleId)
+	}
+
+	for _, role := range *userRoles.Roles {
+		totalRoles = append(totalRoles, *role.Id)
+	}
```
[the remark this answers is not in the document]
carnellj-genesys: I dont think I can be roles is a complex object. If I just wanted to a combined list of Role objects I could do this, but the reality is that we are trying to collect all of the roleIds so I have to walk through each one and the specific role id.

genesyscloud/oauth_client/resource_genesyscloud_oauth_client.go
```diff
@@ -93,18 +101,69 @@ func createOAuthClient(ctx context.Context, d *schema.ResourceData, meta interfa
...
+
+	//Step #5 - Update addedRoles
+	_, resp, err = op.UpdateTerraformUserRoles(ctx, *terraformUser.Id, totalRoles)
+	if err != nil {
+		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to update the terraform user addedRoles running this terraform code %s", err), resp)
+	}
+
```
carnellj-genesys, line 155: This sleep is a bit of ugliness.

genesyscloud/util/util_strings.go
```diff
@@ -9,6 +9,10 @@ var matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
 var matchAllCap = regexp.MustCompile("([a-z0-9])([A-Z])")
 var matchUnderscore = regexp.MustCompile("_")
 
+func ToStrPtr(s string) *string {
```
carnellj-genesys, line 12: Useful little method I wrote to convert a string to *String pointer.  That way you don't have to assign a String to a variable just to pass it into a struct

dginty4: We already have this function in the go sdk, `platformclientv2.String()`. There's also one's for integers, booleans and floats

HemanthDogiparthi12 (approved):

carnellj-genesys:
> Should we use this new cache map for the oauth_client export as well?

No.  I kept separate on purpose.

genesyscloud/util/util_strings.go
carnellj-genesys: Nice.  I did not know this.  Removing this function and replacing ToPtrStr with the platform call

PR #948, opened 2024-04-02
Feat/add flow log level resource

ben-howarth-genesys:
adds support for setting flow log level resource


docs/resources/flow_loglevel.md
```diff
@@ -0,0 +1,56 @@
...
+## Schema
+
+### Required
+
+- `flow_id` (String) The flowId for this characteristics set
+- `flow_log_level` (String) The logLevel for this characteristics set
+
+### Read-Only
+
+- `flow_characteristics` (List of Object) Shows what characteristics are enabled for this log level (see [below for nested schema](#nestedatt--flow_characteristics))
```
charliecon, line 40: Why is this field listed as a read only attribute? Is this something the user should be able to configure using this resource?

examples/resources/genesyscloud_flow_logLevel/apis.md
```diff
@@ -0,0 +1,6 @@
+* todo add links when available
+* [GET /api/v2/flows/instances/settings/loglevels]()
+* [POST /api/v2/flows/{flowId}/instances/settings/loglevels/{flowId}]()
+* [GET /api/v2/flows/{flowId}/instances/settings/loglevels/{flowId}]()
+* [PUT /api/v2/flows/{flowId}/instances/settings/loglevels/{flowId}]()
+* [DELETE /api/v2/flows/{flowId}/instances/settings/loglevels/{flowId}]()
```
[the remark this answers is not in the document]
dginty4: These endpoints are preview apis, you could link to them on the preview api's page on api explorer, https://developer.genesys.cloud/platform/preview-apis

genesyscloud/flow_logLevel/resource_genesyscloud_flow_logLevel.go
```diff
@@ -0,0 +1,153 @@
...
+	ep := getFlowLogLevelProxy(clientConfig)
+	resources := make(resourceExporter.ResourceIDMetaMap)
+
+	flowLogLevels, err := ep.getAllFlowLogLevels(ctx)
+	if err != nil {
+		return nil, diag.Errorf("Failed to get flow log levels: %v", err)
+	}
+
+	for _, flowLogLevel := range *flowLogLevels {
+		log.Printf("Dealing with flow log level id : %s", *flowLogLevel.Id)
```
dginty4, line 52: You can probably remove this log, they tend to be unnessary

genesyscloud/flow_logLevel/resource_genesyscloud_flow_logLevel.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		flowLogLevel := flowSettingsResponse.LogLevelCharacteristics
+
+		cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, ResourceFlowLoglevel())
+		resourcedata.SetNillableValue(d, "flow_log_level", flowLogLevel.Level)
+		resourcedata.SetNillableValueWithInterfaceArrayWithFunc(d, "flow_characteristics", flowLogLevel.Characteristics, flattenFlowCharacteristics)
+
+		log.Printf("Read flow log level %s", flowId)
+		checkState := cc.CheckState()
```
dginty4, line 103: Instead of getting the value returned from `cc.CheckState()` and printing it, you can do `return cc.CheckState()`

genesyscloud/flow_logLevel/resource_genesyscloud_flow_logLevel.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	return util.WithRetries(ctx, 180*time.Second, func() *retry.RetryError {
+		_, respCode, err := ep.getFlowLogLevelById(ctx, flowId)
+
+		if err == nil {
+			return retry.NonRetryableError(fmt.Errorf("Error deleting flow log level %s: %s", flowId, err))
+		}
+		if util.IsStatus404ByInt(respCode) {
+			// Success  : External contact deleted
```
dginty4, line 146: Just a wrong comment here

genesyscloud/flow_logLevel/resource_genesyscloud_flow_logLevel_test.go
```diff
@@ -0,0 +1,125 @@
...
+	for _, rs := range state.RootModule().Resources {
+		if rs.Type != "genesyscloud_flow_loglevel" {
+			continue
+		}
+		expandArray := []string{"logLevelCharacteristics.characteristics"}
+		flowLogLevel, resp, err := architectAPI.GetFlowInstancesSettingsLoglevels(rs.Primary.ID, expandArray)
+		if flowLogLevel != nil {
+			return fmt.Errorf("flowLogLevel for flowId (%s) still exists", rs.Primary.ID)
+		} else if util.IsStatus404(resp) {
+			// Language not found as expected
```
dginty4, line 116: Wrong comment here, there's also another a few lines down

genesyscloud/flow_logLevel/resource_genesyscloud_flow_logLevel_schema.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+	}
+}
+
+// FlowLoglevelExporter returns the resourceExporter object used to hold the genesyscloud_flow_logLevels exporter's config
+func FlowLoglevelExporter() *resourceExporter.ResourceExporter {
+	return &resourceExporter.ResourceExporter{
+		GetResourcesFunc: provider.GetAllWithPooledClient(getAllFlowLogLevels),
+		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
+			"external_organization": {}, //Need to add this when we external orgs implemented
```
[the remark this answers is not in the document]
ben-howarth-genesys: No, I'll remove the whole thing

docs/resources/flow_loglevel.md
ben-howarth-genesys: these are returned by the api and are configured based on the log level that the user sets, I included them as they're illustrative of the log level after an operation but maybe it's something we don't need?

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_utils.go
```diff
@@ -0,0 +1,83 @@
+package flow_loglevel
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v125/platformclientv2"
+	"terraform-provider-genesyscloud/genesyscloud/util/resourcedata"
+)
+
+func generateFlowLogLevelResource(
```
dginty4, line 10: This function can be in the test file. We only need to have these functions outside of the test file when they are public since functions can't be public in a test file. In this case though since the function is private it can stay in the test file

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_schema.go
```diff
@@ -0,0 +1,101 @@
...
+			"flow_log_level": {
+				Description: "The logLevel for this characteristics set",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"flow_characteristics": {
+				Description: "Shows what characteristics are enabled for this log level",
+				Type:        schema.TypeList,
+				Computed:    true,
+				Elem:        flowCharacteristics,
```
charliecon, line 97: I would add `MaxItems: 1`  since the user shouldn't define more than one `flow_characteristics` block

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_utils.go
```diff
@@ -0,0 +1,83 @@
...
+		return &platformclientv2.Flowloglevel{
+			Level: &level,
+		}
+	}
+
+}
+
+// getFlowLogLevelCharacteristicsFromResourceData maps data from schema ResourceData object to a platformclientv2.Flowcharacteristics
+func getFlowLogLevelCharacteristicsFromResourceData(d *schema.ResourceData) *platformclientv2.Flowcharacteristics {
+	characteristics := d.Get("flow_characteristics").([]interface{})[0].(map[string]interface{})
```
charliecon, line 49: A panic could be thrown here. Since all the fields in `flow_characteristics` are optional, the user could define the block like this `flow_characteristics{}` , then the list would have one item and it will be nil. 

As a work around, you can just define the map variable like this

```golang
characteristics, ok := d.Get("flow_characteristics").([]interface{})[0].(map[string]interface{})
if !ok {
  return nil
}
```

docs/resources/flow_loglevel.md
charliecon: No, it looks like we need them. If you can configure them on a POST request, then they belong here as well. I'm just confused as to why they're listed as read only here. Did you use the `go generate` command for these docs? Nothing should be written manually inside the docs folder because this command will always overwrite it

genesyscloud/flow_loglevel/genesyscloud_flow_loglevel_proxy.go
dginty4, line 1: There's a few comments in this file that just say external contacts. it might be easier to do a find on the package just to find any incorrect comments and stuff

examples/resources/genesyscloud_flow_loglevel/resource.tf
```diff
@@ -0,0 +1,4 @@
+resource "genesyscloud_flow_loglevel" "flowLogLevel" {
```
dginty4, line 1: I would add every attribute the user can set to this example. It helps to show the full structure of the resource

genesyscloud/flow_loglevel/genesyscloud_flow_loglevel_proxy.go
```diff
@@ -0,0 +1,164 @@
...
+func getAllFlowLogLevelsFn(ctx context.Context, p *flowLogLevelProxy) (*[]platformclientv2.Flowsettingsresponse, *platformclientv2.APIResponse, error) {
+	const pageSize = 100
+	var totalFlowLogLevels []platformclientv2.Flowsettingsresponse
+
+	flowSettingsResponse, apiResponse, err := p.architectApi.GetFlowsInstancesSettingsLoglevels([]string{"logLevelCharacteristics.characteristics"}, 1, pageSize)
+	if err != nil {
+		return nil, apiResponse, fmt.Errorf("Failed to get page of flows: %v", err)
+	}
+
+	for _, flowLogLevel := range *flowSettingsResponse.Entities {
```
dginty4, line 131: We are using a new syntax when looping through an array and adding it to another array. You can do this instead
`totalFlowLogLevels = append(totalFlowLogLevels, *flowSettingsResponse.Entities...)`

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// readFlowLogLevels is used by the flow_loglevel resource to read a flow log level from genesys cloud.
+func readFlowLogLevel(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	ep := getFlowLogLevelProxy(sdkConfig)
+	flowId := d.Get("flow_id").(string)
+
+	log.Printf("Reading readFlowLogLevel with flowId %s", flowId)
+	if flowId == "" {
```
dginty4, line 84: You don't need this check since flow_id is required in the schema this will always exists

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel.go
```diff
@@ -0,0 +1,158 @@
...
+		log.Printf("flow log level with blank flowId %s", flowId)
+		return diag.Errorf("flowId: %s not found ", flowId)
+	}
+	return util.WithRetriesForRead(ctx, d, func() *retry.RetryError {
+		flowSettingsResponse, apiResponse, err := ep.getFlowLogLevelById(ctx, flowId)
+		if err != nil {
+			log.Print(err)
```
dginty4, line 91: You don't need to log the error since it's being returned

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// updateFlowLogLevels is used by the flow_loglevel resource to update an flow log level in Genesys Cloud
+func updateFlowLogLevel(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	ep := getFlowLogLevelProxy(sdkConfig)
+	flowId := d.Get("flow_id").(string)
+	log.Printf("Updating flow log level for flow %s", flowId)
+
+	_, apiResponse, err := ep.getFlowLogLevelById(ctx, flowId)
```
dginty4, line 116: Why are you getting flow log level here?

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_test.go
```diff
@@ -0,0 +1,123 @@
...
+					resource.TestCheckResourceAttr("genesyscloud_flow_loglevel."+resourceId, "flow_characteristics.0.event_other", "false"),
+					resource.TestCheckResourceAttr("genesyscloud_flow_loglevel."+resourceId, "flow_characteristics.0.event_warning", "true"),
+					resource.TestCheckResourceAttr("genesyscloud_flow_loglevel."+resourceId, "flow_characteristics.0.execution_input_outputs", "false"),
+					resource.TestCheckResourceAttr("genesyscloud_flow_loglevel."+resourceId, "flow_characteristics.0.execution_items", "true"),
+					resource.TestCheckResourceAttr("genesyscloud_flow_loglevel."+resourceId, "flow_characteristics.0.names", "false"),
+					resource.TestCheckResourceAttr("genesyscloud_flow_loglevel."+resourceId, "flow_characteristics.0.variables", "false"),
+				),
+			},
+			{
+				// Create using only flow log level
```
dginty4, line 61: I think you might need to update these comments on the test steps?

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_utils.go
```diff
@@ -0,0 +1,83 @@
...
+
+// getFlowLogLevelSettingsRequestFromResourceData maps data from schema ResourceData object to a platformclientv2.Flowloglevelrequest
+func getFlowLogLevelSettingsRequestFromResourceData(d *schema.ResourceData) platformclientv2.Flowloglevelrequest {
+	return platformclientv2.Flowloglevelrequest{
+		LogLevelCharacteristics: getFlowLogLevelFromResourceData(d),
+	}
+}
+
+// getFlowLogLevelFromResourceData maps data from schema ResourceData object to a platformclientv2.Flowloglevel
+func getFlowLogLevelFromResourceData(d *schema.ResourceData) *platformclientv2.Flowloglevel {
```
dginty4, line 32: I think you can change the logic of this function just to make it shorter. First you can create the `platformclientv2.Flowloglevel` struct with the `Level`. After that do your check to see if you need to add the `Characteristics`, you don't need an else that way and can just have one return statement. P.S. There's a function built into our go sdk that can return a pointer to a string, you can use it for required attributes, e.g. 
```
logLevel := platformclientv2.Flowloglevel{
	Level:   platformclientv2.String(d.Get("flow_log_level").(string)),
}
```

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_utils.go
```diff
@@ -0,0 +1,83 @@
...
+	  flow_id					= "%s"
+	  flow_log_level 			= "%s"
+	}`,
+		resourceId,
+		flowId,
+		flowLoglevel)
+}
+
+// getFlowLogLevelSettingsRequestFromResourceData maps data from schema ResourceData object to a platformclientv2.Flowloglevelrequest
+func getFlowLogLevelSettingsRequestFromResourceData(d *schema.ResourceData) platformclientv2.Flowloglevelrequest {
```
dginty4, line 25: This function isn't really doing a lot, It would be easier if you create the `Flowloglevelrequest` directly in the create and update.

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel.go
ben-howarth-genesys: it's an update method so verifying that the flow exists before taking that action was the idea. I'll remove it though, seems redundant looking at it now

examples/resources/genesyscloud_flow_loglevel/resource.tf
ben-howarth-genesys: every possible flow log level or all the sub fields? The flowLogLevel characteristics can't be set individually, they're just set and returned based on the flow_log_level value a user provides

docs/resources/flow_loglevel.md
ben-howarth-genesys: you can include them on a post request but they do not have an impact. The characteristics are set by the `flow_log_level` field even though they are accepted by the post request and returned by it. I haven't edited anything here manually but I'll run `go generate` again

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel.go
dginty4: I can understand why you thought this, there are a lot of update functions that do a get before the update. the reason they do this is if the genesyscloud object has a version property so we need to do a get to the the current version before updating. In this case though there's no version so you can leave it out

dginty4:
@ben-howarth-genesys Would you be able to clarify something for me please. 
Screenshot 2024-04-11 at 16 29 02 (2)
What properties on the `FlowLogLevel` object should the user set? My understanding from some of the conversations here is that they should only set `level` and the others are read-only? Am I correct?

dginty4:
@ben-howarth-genesys In that case I think you can remove the characteristics attribute from the resource. I understand why you added it but terraform is a tool used to manage infrastructure thus terraform resources only need to include attributes that impact behaviour. For this reason, read-only attributes should not be included unless they are referenced by another resource, but this is rare.  That is my understanding of it anyway, @charliecon  @carnellj-genesys @HemanthDogiparthi12 would you agree or no?

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_utils.go
dginty4, line 1: This has become very small, you can probably add `getFlowLogLevelFromResourceData` to the resource file resource_genesyscloud_flow_loglevel.go and remove this

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_test.go
```diff
@@ -0,0 +1,112 @@
...
+					flowId,
+					flowLoglevelBase,
+					resourceId,
+				),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("genesyscloud_flow_loglevel."+resourceId, "flow_log_level", flowLoglevelBase),
+				),
+			},
+			{
+				// Create using flow log level All
```
dginty4, line 53: This test step will actually update the resource. Resources are not deleted between test steps so for each test step after the first one, the update function is called. You don't need to change the test or anything I would just update the comment to say update, same for the one below. I know this is just a comment but I don't want someone to look at this is the future and be confused.

genesyscloud/flow_loglevel/resource_genesyscloud_flow_loglevel_schema.go
```diff
@@ -0,0 +1,50 @@
...
+		DeleteContext: provider.DeleteWithPooledClient(deleteFlowLogLevel),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"flow_id": {
+				Description: "The flowId for this characteristics set",
+				Type:        schema.TypeString,
+				Required:    true,
```
dginty4, line 41: You'll need to add `ForceNew: true` to this attribute so that the resource will be recreated if the flow_id changes

genesyscloud/flow_loglevel/genesyscloud_flow_loglevel_proxy.go
```diff
@@ -0,0 +1,160 @@
...
+// getAllFlowLogLevelsFn is the implementation for retrieving all flow log levels in Genesys Cloud
+func getAllFlowLogLevelsFn(ctx context.Context, p *flowLogLevelProxy) (*[]platformclientv2.Flowsettingsresponse, *platformclientv2.APIResponse, error) {
+	const pageSize = 100
+	var totalFlowLogLevels []platformclientv2.Flowsettingsresponse
+
+	flowSettingsResponse, apiResponse, err := p.architectApi.GetFlowsInstancesSettingsLoglevels(nil, 1, pageSize)
+	if err != nil {
+		return nil, apiResponse, fmt.Errorf("Failed to get page of flows: %v", err)
+	}
+
```
dginty4, line 130: You're just missing one check here, you need to check if `flowSettingsResponse` is nil or if the length of the `Entities` is 0, there should be example in other resource. You'll also need this check in the loop below

dginty4:
@ben-howarth-genesys I just left a few more comments, after that it should be good. Just one more thing though. There's a test file for the exporter `tfexporter/tf_exporter_resource_test.go`, you will need to add your resource to that file. You'll see two list in the file, add the resource to one list and the exporter to the second

genesyscloud/flow_loglevel/genesyscloud_flow_loglevel_proxy.go
```diff
@@ -0,0 +1,168 @@
...
+	if flowSettingsResponse.Entities == nil || len(*flowSettingsResponse.Entities) == 0 {
+		return nil, nil, fmt.Errorf("No flow log level found")
+	}
```
charliecon, line 133: One thing before we merge - I don't believe you should return an error here. No entities being returned isn't an error in the API or the terraform provider. You can just `return &totalFlowLogLevels, apiResponse, nil`

genesyscloud/flow_loglevel/genesyscloud_flow_loglevel_proxy.go
```diff
@@ -0,0 +1,168 @@
...
+
+	totalFlowLogLevels = append(totalFlowLogLevels, *flowSettingsResponse.Entities...)
+
+	for pageNum := 2; pageNum <= *flowSettingsResponse.PageCount; pageNum++ {
+		flowSettingsResponse, apiResponse, err := p.architectApi.GetFlowsInstancesSettingsLoglevels(nil, pageNum, pageSize)
+		if err != nil {
+			return nil, apiResponse, fmt.Errorf("Failed to get page %d of flow log levels: %v", pageNum, err)
+		}
+		if flowSettingsResponse.Entities == nil || len(*flowSettingsResponse.Entities) == 0 {
+			return nil, nil, fmt.Errorf("No flow log level found on page %d of flow log levels", pageNum)
```
charliecon, line 143: Same here

charliecon (approved):

dginty4 (approved):
There's just a few merge conflicts to resolve and you should be good. I have put `do not merge` on the PR. We just need to enable the feature toggle for these endpoints in our test org before we can merge this.

PR #957, opened 2024-04-05
Bug/devtooling-530: Fixing panic during export caused by MemberGroupsResolver

charliecon:


genesyscloud/group/resource_genesyscloud_group_schema.go
```diff
@@ -109,9 +109,9 @@ func ResourceGroup() *schema.Resource {
 			"owner_ids": {
 				Description: "IDs of owners of the group.",
 				Type:        schema.TypeList,
-				Required:    true,
+				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
-				MinItems:    1,
+				Computed:    true,
```
charliecon, line 114: owner_ids shouldn't have been required. You can create a group without this field, it will just be populated with one item - the creating user's ID. I had to change this in this PR because it was preventing me from making sure the groups exporter was working properly

carnellj-genesys: @charliecon Why is this one marked as Do Not Merge.  Are we ready to get this resolved

charliecon: It's marked as "do not merge" because, at the moment, the target branch is main and it is open as a hotfix PR. We were waiting on a response from the forum user before deciding if it should go out as a hotfix or with the regular schedule. Since they still haven't responded, I'll resolve any conflicts and change the target branch to `dev`

charliecon: Made all the updates. This PR is ready to go into dev 👍🏻

dginty4:
I might be misunderstanding this but was owner_ids not made required to avoid "orphan" groups that nobody can access? https://inindca.atlassian.net/browse/DEVTOOLING-321

dginty4:
> > I might be misunderstanding this but was owner_ids not made required to avoid "orphan" groups that nobody can access? https://inindca.atlassian.net/browse/DEVTOOLING-321
> 
> I couldn't apply exported groups to another org because some of them would have an empty owner_ids array (probably weren't created using terraform), and then the apply would fail saying that I'm missing a required argument.
> 
> Cesar has since created this ticket - https://inindca.atlassian.net/browse/DEVTOOLING-475 , which describes how the solution is setting visibility to "public" , not forcing the user to provide an owner

Ah ok I understand. Should we add a note explaining that if the group is not public and doesn't have any owners that it will be orphaned? Or maybe check it in the code?

charliecon:
> Ah ok I understand. Should we add a note explaining that if the group is not public and doesn't have any owners that it will be orphaned? Or maybe check it in the code?

@dginty4 @carnellj-genesys 

I don't think it's necessary. The description for `visibility` and the fact that it defaults to `public` should be enough to safeguard, or at least to put the onus on the user.

dginty4 (approved):

PR #958, opened 2024-04-08
feat/Devtooling-531: Add proxy and cache to routing queue

dginty4:
This PR adds a proxy to routing queues so I can add caching. This is needed because in the conditional_group_routing and outbound_email_address resources, the get all is also getting all routing queues which leads to a lot of api calls. I am making the proxy in routing queues public so the get all call and the cache can be shared with conditional_group_routing and outbound_email_address


genesyscloud/routing_queue/genesyscloud_routing_queue_proxy.go
```diff
@@ -0,0 +1,163 @@
...
+	queues, resp, getErr := p.routingApi.GetRoutingQueues(1, pageSize, "", "", nil, nil, nil, false)
+	if getErr != nil {
+		return nil, resp, fmt.Errorf("failed to get first page of queues: %v", getErr)
+	}
+
+	// Check if the routing queue cache is populated with all the data, if it is, return that instead
+	// If the size of the cache is the same as the total number of queues, the cache is up-to-date
+	if rc.GetCacheSize(p.RoutingQueueCache) == *queues.Total {
+		return rc.GetCache(p.RoutingQueueCache), nil, nil
+	} else if rc.GetCacheSize(p.RoutingQueueCache) != *queues.Total && rc.GetCacheSize(p.RoutingQueueCache) != 0 {
```
charliecon, line 87: Should this not be done after collecting all the other pages? This is just the first page of queues

dginty4: No I put it here intentionally, my logic is when I do the first api call we can use the `Total` value that is returned to check if the number of routing queues in the org is equal to the number of routing_queues in the cache. If it is then I assume the cache is up to date with all the data and we don't need to do the rest of the api calls adn can use the cache instead

charliecon: Ah right of course. Missed the total part, assumed you were checking length

genesyscloud/routing_queue/genesyscloud_routing_queue_proxy.go
```diff
@@ -0,0 +1,163 @@
...
+	const pageSize = 100
+
+	queues, resp, getErr := p.routingApi.GetRoutingQueues(1, pageSize, "", "", nil, nil, nil, false)
+	if getErr != nil {
+		return nil, resp, fmt.Errorf("failed to get first page of queues: %v", getErr)
+	}
+
+	// Check if the routing queue cache is populated with all the data, if it is, return that instead
+	// If the size of the cache is the same as the total number of queues, the cache is up-to-date
+	if rc.GetCacheSize(p.RoutingQueueCache) == *queues.Total {
```
HemanthDogiparthi12, line 85: do we need cache for getAll ?

dginty4: The reason I have it like this is because this getAll will eventually be used by 3 resources, `routing_queues`, `routing_queues_conditional_group_routing` and `routing_queue_outbound_email_address`. If I cache the get all when the first resource call it, it stops the other two resources having to make the same call

HemanthDogiparthi12 (approved):

charliecon (approved):

carnellj-genesys (approved):

PR #969, opened 2024-04-11
Feat/devtooling-499: Added getAll Functionality to ResponseManagement_ResponseAsset

BrianMoyles:


genesyscloud/responsemanagement_response/resource_genesyscloud_responsemanagement_response_test.go
```diff
@@ -48,6 +48,7 @@ func TestAccResourceResponseManagementResponseFooterField(t *testing.T) {
 		assetResource = "asset-resource"
 		fileName      = "yeti-img.png"
 		fullPath      = fmt.Sprintf("%s/%s", testFilesDir, fileName)
+		filePath      = "test_responseasset_data/yeti-img.png"
```
charliecon, line 51: I think we should start using `filepath.Join` for test cases like this. I'm pretty sure this wouldn't work on a windows machine because they use \ instead of / 

I had to deal with a bug in the CLI recently that was caused by manually built path strings (code I wrote) instead of using a package that would account for different operating systems.

genesyscloud/responsemanagement_response/resource_genesyscloud_responsemanagement_response_test.go
```diff
@@ -233,6 +236,7 @@ func TestAccResourceResponseManagementResponseMessaging(t *testing.T) {
...
```
charliecon, line 239: same here about using filepath.Join

genesyscloud/responsemanagement_responseasset/data_source_genesyscloud_responsemanagement_responseasset_test.go
```diff
@@ -16,6 +16,7 @@ func TestAccDataSourceResponseManagementResponseAsset(t *testing.T) {
 		testDirName  = "test_responseasset_data"
 		fileName     = fmt.Sprintf("%s/yeti-img.png", testDirName)
 		dataSourceId = "resp_asset_data"
+		filePath     = "test_responseasset_data/yeti-img.png"
```
charliecon, line 19: Same here

charliecon: and two lines above

genesyscloud/responsemanagement_responseasset/resource_genesyscloud_responsemanagement_responseasset_test.go
```diff
@@ -23,6 +23,8 @@ func TestAccResourceResponseManagementResponseAsset(t *testing.T) {
 		fileName2          = "genesys-img.png"
 		fullPath1          = fmt.Sprintf("%s/%s", testFilesDir, fileName1)
 		fullPath2          = fmt.Sprintf("%s/%s", testFilesDir, fileName2)
+		filePath1          = "test_responseasset_data/yeti-img.png"
+		filePath2          = "test_responseasset_data/genesys-img.png"
```
charliecon, line 27: Same here

genesyscloud/responsemanagement_responseasset/resource_genesyscloud_responsemanagement_responseasset_utils.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return err
+	}
+
+	log.Println("Details: ", fullPath, fileName, subDirectory)
+	if err := files.DownloadExportFile(fullPath, fileName, *data.ContentLocation); err != nil {
+		return err
+	}
+
+	configMap["filepath"] = path.Join(subDirectory, fileName)
```
charliecon, line 35: Here we should be updating filename, since it is actually used in the api call to create the asset

genesyscloud/responsemanagement_responseasset/resource_genesyscloud_responsemanagement_responseasset_utils.go
```diff
@@ -0,0 +1,48 @@
...
+	"path"
+	"terraform-provider-genesyscloud/genesyscloud/provider"
+	"terraform-provider-genesyscloud/genesyscloud/util/files"
+)
+
+func responsemanagementResponseassetResolver(responseAssetId, exportDirectory, subDirectory string, configMap map[string]interface{}, meta interface{}) error {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := getRespManagementRespAssetProxy(sdkConfig)
+
+	fileName := fmt.Sprintf("asset-%s.jpeg", responseAssetId)
```
charliecon, line 17: For users exporting assets, it looks like the filename is really useful for differentiating them, since they don't have a regular name. The only other way, from a terraform perspective, is to open the image files. If we could keep the filename in there, I think it would be better. So something like `"{the value of filename}-{responseAssetId}.jpeg"`
You'll just have to trim off any path directories that prefix the file name

charliecon (commented):
Looks good. One other thing - we should add the `file_content_hash` field to this resource. There are examples of it in the scripts and flows resource. It stores the hash sum of the file, so that if the user were to delete the jpeg and replace it with another (with the same path and file name) after having already deployed, then the resource will catch that something has changed and force a recreation

genesyscloud/responsemanagement_responseasset/genesyscloud_responsemanagement_responseasset_proxy.go
```diff
@@ -80,6 +91,48 @@ func (p *responsemanagementResponseassetProxy) deleteRespManagementRespAsset(ctx
...
+		return nil, resp, fmt.Errorf("Failed to get response asset search request: %v", err)
+	}
+
+	if responseAssets.Results == nil || len(*responseAssets.Results) == 0 {
+		return &allResponseAssets, resp, nil
+	}
+	allResponseAssets = append(allResponseAssets, *responseAssets.Results...)
+
+	for pageNum := 2; pageNum <= *responseAssets.PageCount; pageNum++ {
+		responseAssets, resp, err := p.responseManagementApi.PostResponsemanagementResponseassetsSearch(platformclientv2.Responseassetsearchrequest{
```
carnellj-genesys, line 114: Does this resource upload assets to Genesys Cloud?  If so how do we pull the assets back down?

carnellj-genesys (changes requested):

genesyscloud/responsemanagement_responseasset/genesyscloud_responsemanagement_responseasset_proxy.go
BrianMoyles: that is being handled in the resolver in the utils file through the 
`files.DownloadExportFile(fullPath, fileName, *data.ContentLocation)`

dginty4 (approved):

PR #974, opened 2024-04-15
Feature/devtooling 63

HemanthDogiparthi12:


genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
```diff
@@ -33,6 +33,27 @@ var (
 		},
 	}
 
+	agentOwnedRoutingResource = &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"enable_agent_owned_callbacks": {
+				Description: "Enable Agent Owned Callbacks",
+				Type:        schema.TypeBool,
+				Required:    true,
+				Default:     false,
```
charliecon, line 42: I don't think it is necessary to have a default value if the field is required, since they will have to provide a value anyway. Also the default (or zero) value for all bool variables in golang is false anyway

charliecon (approved):

genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -43,9 +46,13 @@ func GenerateRoutingQueueResource(
 		callingPartyName,
 		callingPartyNumber,
 		enableTranscription,
+		scoringMethod,
 		suppressInQueueCallRecording,
 		enableManualAssignment,
 		strings.Join(nestedBlocks, "\n"))
+
+	log.Print(f)
```
dginty4, line 54: Why are you printing this?

HemanthDogiparthi12: unwanted log. removed it

genesyscloud/routing_queue/resource_genesyscloud_routing_queue.go
```diff
@@ -71,6 +72,7 @@ func createQueue(ctx context.Context, d *schema.ResourceData, meta interface{})
 		Bullseye:                     buildSdkBullseyeSettings(d),
 		ConditionalGroupRouting:      conditionalGroupRouting,
 		AcwSettings:                  buildSdkAcwSettings(d),
+		AgentOwnedRouting:            constructAgentOwnedRouting(d),
```
dginty4, line 75: Can just to ` buildAgentOwnedRouting` just to keep everything consistent

dginty4: Sorry I just noticed buildAgentOwnedRouting already exists

dginty4 (approved):

PR #989, opened 2024-04-22
DEVTOOLING-445: User Roles not returning duplicate roles

dginty4:
I had to update the go sdk to get this change since this feature was only just released


charliecon (approved):
Just need to update the sdk version in the new conditional_group_routing package

PR #1014, opened 2024-05-02
Bug/devtooling 444: Custom resolver function for outbound campaign script_id

charliecon:


genesyscloud/resource_cache/resource_cache.go
```diff
@@ -25,6 +26,12 @@ func SetCache[T any](cache CacheInterface[T], key string, value T) {
 	}
 }
 
+func DeleteCacheItem[T any](cache CacheInterface[T], key string) {
```
carnellj-genesys, line 29: I am not sure why we have a DeleteCache Item method.  What is the use case you are running into?  Just trying to understand because generally the cache is used for the exporter and I am not sure I see the use case for the delete.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -1221,6 +1221,29 @@ func (g *GenesysCloudResourceExporter) sanitizeConfigMap(
 			} else {
 				configMap[key] = escapeString(val.(string))
 			}
+
+			// custom function to resolve the field to a specific data source depending on the value
+			if refAttrCustomResolver, ok := exporter.CustomAttributeResolver[currAttr]; ok {
```
carnellj-genesys, line 1226: Oof.  This is a pretty deep and wide block of code.  Should we break this out into its own function and then have unit tests to test the logic?

genesyscloud/resource_cache/resource_cache.go
charliecon: In my acceptance tests, I created an export resource and a campaign. The `deleteOutboundCampaign` method was getting stuck because it deletes the campaign and then reads by ID with retries until it receives a 404. Because the export resource is also present, `IsExporterActive` returns true, and the campaign is read from cache. That causes the test to fail and claim that the campaign still exists, even though it does not. 

This can happen with any resource that does reads to confirm that the object has been deleted before exiting the deleteContext function.

I haven't been able to recreate this behaviour using the locally compiled binary, so I'm not sure yet what the difference is with test cases that allows this to happen.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
charliecon: You're right. I didn't stop to consider it had gotten a bit out of control 😄 I'll refactor it

carnellj-genesys (approved):

dginty4:
There's a data source test failing, `TestAccDataSourceScriptPublishedDefaults`. is it related to this work?

charliecon:
> There's a scripts data source test failing, `TestAccDataSourceScriptPublishedDefaults`. is it related to this work?

It wasn't related to this PR, but I fixed it there anyway

dginty4 (approved):

PR #1015, opened 2024-05-03
bug/devtooling 560 - Nil error in telephony_providers_edges_phone

BrianMoyles:
A log.Printf line was causing a nil pointer error. This line was executed before an API call was handled for error checks. 
This PR removes the line that throws the error and makes a few small changes with a linter

### DIFF
--- a/genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone.go
+++ b/genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone.go
@@ -42,10 +42,11 @@ func createPhone(ctx context.Context, d *schema.ResourceData, meta interface{}) 
 	if err != nil {
 		return diag.Errorf("failed to create phone %v: %v", *phoneConfig.Name, err)
 	}
+
 	log.Printf("Creating phone %s", *phoneConfig.Name)
+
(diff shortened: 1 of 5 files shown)


dginty4 (approved):

genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -42,10 +42,11 @@ func createPhone(ctx context.Context, d *schema.ResourceData, meta interface{})
 	if err != nil {
 		return diag.Errorf("failed to create phone %v: %v", *phoneConfig.Name, err)
 	}
+
 	log.Printf("Creating phone %s", *phoneConfig.Name)
+
 	diagErr := util.RetryWhen(util.IsStatus404, func() (*platformclientv2.APIResponse, diag.Diagnostics) {
 		phone, resp, err := pp.createPhone(ctx, phoneConfig)
-		log.Printf("Completed call to create phone name %s with status code %d, correlation id %s and err %s", *phoneConfig.Name, resp.StatusCode, resp.CorrelationID, err)
 		if err != nil {
```
charliecon, line 50: I think we could keep this log statement. It provides some useful info. It would just need to go after the error check

charliecon (approved):

carnellj-genesys (approved):

PR #1027, opened 2024-05-09
Pre-verify if the tfState is in sync with the exporter

HemanthDogiparthi12:


dginty4 (approved):

genesyscloud/util/feature_toggles/state_comparison.go
```diff
@@ -0,0 +1,19 @@
+package feature_toggles
+
+import "os"
+
+const enableStateComparison = "ENABLE_EXPORTER_STATE_COMPARISON"
```
charliecon, line 5: @HemanthDogiparthi12 @dginty4 @carnellj-genesys 

Is there somewhere in the docs where we should mention this env variable so users know about it? If not, we could still recommend it in the log output if they don't have it set

charliecon (approved):

genesyscloud/util/feature_toggles/state_comparison.go
HemanthDogiparthi12: right now this is used for troubleshooting some of the problems Cesar reported. hence added it as env variable instead of adding this as a new property in exporter schema.

PR #1045, opened 2024-05-15
DEVTOOLING-595: Integration Nil pointer exception

dginty4:


genesyscloud/integration/resource_genesyscloud_integration.go
```diff
@@ -71,7 +71,7 @@ func createIntegration(ctx context.Context, d *schema.ResourceData, meta interfa
 	}
 	integration, resp, err := ip.createIntegration(ctx, createIntegrationReq)
 	if err != nil {
-		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to create integration %s error: %s", *createIntegrationReq.Name, err), resp)
```
dginty4, line 74: This log didn't print the name attribute originally as it has not been set hence the nil pointer. It was changed by mistake when we were adding the new diag errors

charliecon (approved):

PR #1048, opened 2024-05-20
Feat/devtooling-477: Site Outbound Routes Resource

dginty4:
Moved the outbound routes property on the sites resource out into it's own resource. Made some changes to the sites proxy as well


charliecon:
I think the example resource and apis.md file aren't getting picked up because of a difference between the value of `resourceName` and the resource name in the path to the example files. I think it was because one says "route" and the other "routes"

dginty4:
> I think the example resource and apis.md file aren't getting picked up because of a difference between the value of `resourceName` and the resource name in the path to the example files. I think it was because one says "route" and the other "routes"

Thanks Charlie, I meant to go back and make that more consistent, will do that now

genesyscloud/telephony_providers_edges_site_outbound_route/resource_genesyscloud_telephony_providers_edges_site_outbound_route.go
```diff
@@ -0,0 +1,247 @@
...
+			}
+		} else {
+			// Add the outbound route if not already present
+			_, resp, err := proxy.createSiteOutboundRoute(ctx, d.Id(), &outboundRoute)
+			if err != nil {
+				return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to add outbound route to site %s error: %s", d.Id(), err), resp)
+			}
+		}
+	}
+
```
charliecon, line 185: This function is very big. I'd recommend having a function that returns the three lists - routes to create, update, and delete. And then calling functions to perform those operations. It could all go behind the proxy as well

genesyscloud/telephony_providers_edges_site_outbound_route/resource_genesyscloud_telephony_providers_edges_site_outbound_route_test.go
```diff
@@ -0,0 +1,226 @@
...
+	"terraform-provider-genesyscloud/genesyscloud/provider"
+	"terraform-provider-genesyscloud/genesyscloud/telephony"
+	telephonyProvidersEdgesSite "terraform-provider-genesyscloud/genesyscloud/telephony_providers_edges_site"
+	"terraform-provider-genesyscloud/genesyscloud/util"
+	featureToggles "terraform-provider-genesyscloud/genesyscloud/util/feature_toggles"
+	"testing"
+)
+
+func TestAccResourceSiteOutboundRoutes(t *testing.T) {
+	t.Parallel()
```
charliecon, line 20: If you're going to remove the t.Skip() from the old tests, then you'd want to remove this t.Parallel line and add a defer func to unset the env variable

dginty4:
@charliecon Think I fixed everything you mentioned

charliecon (approved):

carnellj-genesys (approved):

PR #1060, opened 2024-05-28
Refactor/devtooling 156

shrutisuryawanshigenesys:


genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+)
+
+/*
+The resource_genesyscloud_idp_adfs.go contains all of the methods that perform the core logic for a resource.
+*/
+
+// getAllAuthIdpAdfs retrieves all of the idp adfs via Terraform in the Genesys Cloud and is used for the exporter
+func getAllAuthIdpAdfss(ctx context.Context, clientConfig *platformclientv2.Configuration) (resourceExporter.ResourceIDMetaMap, diag.Diagnostics) {
+	proxy := newIdpAdfsProxy(clientConfig)
```
charliecon, line 31: this should be `getIdpAdfsProxy`. We should only really be calling the `newProxy` method from inside the proxy file

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs.go
```diff
@@ -0,0 +1,145 @@
...
+	resources := make(resourceExporter.ResourceIDMetaMap)
+
+	_, resp, err := proxy.getIdpAdfs(ctx)
+	if err != nil {
+		if util.IsStatus404(resp) {
+			// Don't export if config doesn't exist
+			return resources, nil
+		}
+		return nil, util.BuildAPIDiagnosticError("genesyscloud_idp_adfs", fmt.Sprintf("Failed to get IDP ADFS error: %s", err), resp)
```
charliecon, line 40: You can use `resourceName` here instead of the hard-coded string "genesyscloud_idp_adfs". Same with the other instances below

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs_schema.go
```diff
@@ -0,0 +1,99 @@
...
+3.  The datasource schema definitions for the idp_adfs datasource.
+4.  The resource exporter configuration for the idp_adfs exporter.
+*/
+const resourceName = "genesyscloud_idp_adfs"
+
+// SetRegistrar registers all of the resources, datasources and exporters in the package
+func SetRegistrar(regInstance registrar.Registrar) {
+	regInstance.RegisterResource(resourceName, ResourceIdpAdfs())
+	regInstance.RegisterExporter(resourceName, IdpAdfsExporter())
+}
```
charliecon, line 25: You'll need to import this new package `idp_adfs` to main.go, and call this function from there. You'll see how it is done with all the other packages there. After that, run `go generate` from the command line to generate the docs. If nothing changes, that's a good sign.

charliecon (changes requested):

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs_schema.go
```diff
@@ -0,0 +1,99 @@
...
+				Description: ``,
+				Optional:    true,
+				Type:        schema.TypeString,
+			},
+			`target_uri`: {
+				Description: ``,
+				Optional:    true,
+				Type:        schema.TypeString,
+			},
+			`slo_u_r_i`: {
```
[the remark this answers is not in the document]
charliecon: "Provided by ADSF on app creation" can be the description here. It says it in the input box in the UI before you enter a value. I don't know why it isn't the official description in the API

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs_schema.go
```diff
@@ -0,0 +1,99 @@
...
+			`relying_party_identifier`: {
+				Description: ``,
+				Optional:    true,
+				Type:        schema.TypeString,
+			},
+			`certificate`: {
+				Description: ``,
+				Optional:    true,
+				Type:        schema.TypeString,
+			},
```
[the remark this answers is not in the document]
charliecon: To be clear, you can remove `certificate`, singular

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs_schema.go
```diff
@@ -0,0 +1,99 @@
...
+			`slo_u_r_i`: {
+				Description: ``,
+				Optional:    true,
+				Type:        schema.TypeString,
+			},
+			`slo_binding`: {
+				Description: ``,
+				Optional:    true,
+				Type:        schema.TypeString,
+			},
```
charliecon, line 70: According to the UI, there can only be two values for this field `slo_binding`: "HTTP Redirect" or "HTTP Post". You can add a validation function to this field to enforce this rule. Here is an example - [`dialing_mode` in the campaigns resource](https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_schema.go#L72) 

And for the description, "Valid values: HTTP Redirect, HTTP Post" will suffice. It's better than nothing

charliecon (commented):
just requested a few changes to the schema. Once they're done, you'll have to run `go generate` again

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs.go
```diff
@@ -0,0 +1,145 @@
...
+
+// deleteIdpAdfs is used by the idp_adfs resource to delete an idp adfs from Genesys cloud
+func deleteIdpAdfs(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := getIdpAdfsProxy(sdkConfig)
+
+	_, err := proxy.deleteIdpAdfs(ctx, d.Id())
+	if err != nil {
+		return diag.Errorf("Failed to delete idp adfs %s: %s", d.Id(), err)
+	}
```
charliecon, line 121: You can change this to return the `BuildAPIDiagnosticError` function. You'll need to replace `_` with `resp` to catch the API Response object

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs_init_test.go
```diff
@@ -0,0 +1,51 @@
+package idp_adfs
+
```
charliecon, line 2: There are a few places in this file where the comments reference architect_grammar when it should be idp_adfs

genesyscloud/idp_adfs/resource_genesyscloud_idp_adfs_schema.go
```diff
@@ -0,0 +1,97 @@
...
+func SetRegistrar(regInstance registrar.Registrar) {
+	regInstance.RegisterResource(resourceName, ResourceIdpAdfs())
+	regInstance.RegisterExporter(resourceName, IdpAdfsExporter())
+}
+
+// ResourceIdpAdfs registers the genesyscloud_idp_adfs resource with Terraform
+func ResourceIdpAdfs() *schema.Resource {
+	return &schema.Resource{
+		Description: `Genesys Cloud idp adfs`,
+
```
charliecon, line 32: If you look here - https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/resource_genesyscloud_idp_adfs.go#L50 , you'll see what this description was before. I think it is better to keep it how it was because it provides more info to the end user. You'll need to run `go generate` again after making that change.

charliecon (approved):

PR #1064, opened 2024-05-30
Cache for Data Sources

HemanthDogiparthi12:


charliecon (approved):

genesyscloud/data_source_genesyscloud_routing_skill.go
```diff
@@ -28,24 +32,67 @@ func dataSourceRoutingSkill() *schema.Resource {
...
+	d.SetId(queueId)
+	return nil
+}
+
+func hydrateRoutingSkillCacheFn(c *rc.DataSourceCache) error {
+	log.Printf("hydrating cache for data source genesyscloud_routing_skill")
+
+	routingApi := platformclientv2.NewRoutingApiWithConfig(c.ClientConfig)
+
+	for pageNum := 1; ; pageNum++ {
```
dginty4, line 63: Should we be sticking to the new way of doing this? Where we do one api call first to get the number of pages and then loop up to that

genesyscloud/data_source_genesyscloud_routing_skill.go
```diff
@@ -28,24 +32,67 @@ func dataSourceRoutingSkill() *schema.Resource {
...
+	const pageSize = 100
+	skillId := ""
+	routingAPI := platformclientv2.NewRoutingApiWithConfig(c.ClientConfig)
 
 	// Find first non-deleted skill by name. Retry in case new skill is not yet indexed by search
-	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
-		for pageNum := 1; pageNum <= pageCount; pageNum++ {
-			const pageSize = 100
+	diag := util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
+		for pageNum := 1; ; pageNum++ {
```
dginty4, line 95: Same here

dginty4 (approved):

genesyscloud/data_source_genesyscloud_routing_skill.go
HemanthDogiparthi12: I thought to do it that way, but really find little advantage more than an additional call.  what do you think ?

kavinbalagen (approved):

PR #1065, opened 2024-05-31
Feat/devtooling 610

shrutisuryawanshigenesys:


dginty4 (approved):

genesyscloud/routing_email_route/resource_genesyscloud_routing_email_route_schema.go
```diff
@@ -151,6 +152,25 @@ func ResourceRoutingEmailRoute() *schema.Resource {
 	}
 }
 
+func DataSourceRoutingEmailRoute() *schema.Resource {
+	return &schema.Resource{
+		Description: "Data source for Genesys Cloud Routing Email Route. Select a routing email by name.",
```
charliecon, line 157: Looks like this should say "Select a routing email route by pattern and domain ID." (you'll need to run go generate after this change)

charliecon (commented):
You'll need to add an example data source to the examples folder. The path would be `examples/data-sources/<resourceName>/data-source.tf`

carnellj-genesys (approved):

PR #1067, opened 2024-06-03
[JM-1457]: Add resource for journey views

Rafat001:
Added journey views for following journey views APIs
```
GET /api/v2/journey/views/{viewId}
POST /api/v2/journey/views //Create Journey View
POST /api/v2/journey/views/{viewId}/versions //Update Journey View
DELETE /api/v2/journey/views/{viewId} 
```

Exporter and Data Source were not added since we do not have a working get all resource API. we will add it once https://inindca.atlassian.net/browse/JM-109 is done.


genesyscloud/journey_views/resource_genesyscloud_journey_views_schema.go
```diff
@@ -0,0 +1,213 @@
...
+func SetRegistrar(regInstance registrar.Registrar) {
+	regInstance.RegisterResource(resourceName, ResourceGroup())
+	/***
+	TODO: Add DataSource and Exporter once we are done with https://inindca.atlassian.net/browse/JM-109
+	regInstance.RegisterDataSource(resourceName, DataSourceGroup())
+	regInstance.RegisterExporter(resourceName, JourneyViewExporter())
+	***/
+}
+
+func ResourceGroup() *schema.Resource {
```
dginty4, line 172: This function should be called `ResourceJourneyViews`

genesyscloud/journey_views/resource_genesyscloud_journey_views_schema.go
```diff
@@ -0,0 +1,213 @@
...
+		DeleteContext: provider.DeleteWithPooledClient(deleteJourneyView),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "JourneyView name.",
+				Type:        schema.TypeString,
+				Optional:    true,
```
dginty4, line 188: Are you sure name should be optional? This api might be different but most api's require name

genesyscloud/journey_views/resource_genesyscloud_journey_views_test.go
```diff
@@ -0,0 +1,167 @@
...
+    duration = "%s"
+    name = "%s"
+    %s
+}`, journeyResource, duration, name, func() string {
+		if elementsBlock != "" {
+			return elementsBlock
+		}
+		return ""
+	}())
+	println(jv)
```
dginty4, line 105: You can remove this print and the one below, don't think they're needed

genesyscloud/journey_views/resource_genesyscloud_journey_views_utils.go
```diff
@@ -0,0 +1,253 @@
...
+			}
+			element.FollowedBy = &followedBy
+		}
+		elements = append(elements, element)
+	}
+
+	return &elements, nil
+}
+
+func parseJourneyViewElementAttributes(attributesSlice []interface{}) platformclientv2.Journeyviewelementattributes {
```
dginty4, line 57: Instead of parse we tend to use build for this type of functions, so this would be `buildJourneyViewElementAttributes`, same for the functions below

genesyscloud/journey_views/resource_genesyscloud_journey_views_utils.go
```diff
@@ -0,0 +1,253 @@
...
+			timeConstraint.Value = getIntPointerFromInterface(timeConstraintMap["value"])
+		}
+		if !ok {
+			log.Printf("wrong type 2, correct %T", timeConstraintMap)
+		}
+	}
+	return &timeConstraint
+}
+
+func getStringPointerFromInterface(val interface{}) *string {
```
dginty4, line 147: You don't need this function, when you read an interface value you can cast it to a type, we have a built in function in the go sdk that can create a pointer to the string. So an example of this would be 
```
timeConstraint.Unit = platformclientv2.String(timeConstraintMap["unit"].(string))
```

The same goes when working with booleans and integers, cast and then use the corresponding function in the go sdk to create a pointer(If you want to do it in one line)

dginty4 (changes requested):
@Rafat001 I noticed when building structs you are setting every value on the struct without checking if the user provided that value, this can sometimes lead to nil pointers, inaccurate messages and other problems like that. Setting the value without checking if it is present is fine if the attribute is required in the schema but for optional attributes we need to verify it's presnt before trying to use it. 

Write a test case or two just to verify there is no problems if the user leaves out some optional attributes or even optional blocks

examples/resources/genesyscloud_journey_views/apis.md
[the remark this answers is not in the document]
Rafat001: Our APIs are still not published in https://developer.genesys.cloud/platform/preview-apis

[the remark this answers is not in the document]
dginty4: I think they are https://developer.genesys.cloud/platform/preview-apis#post-api-v2-journey-views

genesyscloud/journey_views/resource_genesyscloud_journey_views_schema.go
Rafat001: name is optional as of now. But yes I guess it's better to make it required, let me talk to my team so that we make the changes in the api

Rafat001:
> @Rafat001 I noticed when building structs you are setting every value on the struct without checking if the user provided that value, this can sometimes lead to nil pointers, inaccurate messages and other problems like that. Setting the value without checking if it is present is fine if the attribute is required in the schema but for optional attributes we need to verify it's presnt before trying to use it.
> 
> Write a test case or two just to verify there is no problems if the user leaves out some optional attributes or even optional blocks

Hi, I'm not sure if I got you right. I'm setting every value as nullable while building the JourneyView object from user's input. And also the first test creates a journey view without elements being passed since that's optional.

examples/resources/genesyscloud_journey_views/apis.md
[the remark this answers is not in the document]
Rafat001: Thanks, I'll update it

genesyscloud/journey_views/resource_genesyscloud_journey_views_utils.go
Rafat001: The function helps resolve any nil values but the sdk function will result a runtime panic for any optional value that's not provided

charliecon (approved):
Looks great! You'll just need to run `go generate` from the root of the project to update the documentation

genesyscloud/journey_views/resource_genesyscloud_journey_views_utils.go
```diff
@@ -0,0 +1,253 @@
...
+func buildJourneyViewLink(linkMap map[string]interface{}) platformclientv2.Journeyviewlink {
+	var link platformclientv2.Journeyviewlink
+	link.Id = getStringPointerFromInterface(linkMap["id"])
+	constraintWithinSlice, ok := linkMap["constraint_within"].([]interface{})
+	if ok {
+		link.ConstraintWithin = buildJourneyViewLinkTimeConstraint(constraintWithinSlice)
+	}
+	if !ok {
+		log.Printf("wrong type. correct type %T", constraintWithinSlice)
+	}
```
charliecon, line 115: This `!ok` block can be removed. How it is done for constraint_after below is perfect

genesyscloud/journey_views/resource_genesyscloud_journey_views_utils.go
dginty4: Ah ok, that's a good point. We might change the sdk functions in the future to have this functionality but for now we can use your functions

dginty4:
> > @Rafat001 I noticed when building structs you are setting every value on the struct without checking if the user provided that value, this can sometimes lead to nil pointers, inaccurate messages and other problems like that. Setting the value without checking if it is present is fine if the attribute is required in the schema but for optional attributes we need to verify it's presnt before trying to use it.
> > Write a test case or two just to verify there is no problems if the user leaves out some optional attributes or even optional blocks
> 
> Hi, I'm not sure if I got you right. I'm setting every value as nullable while building the JourneyView object from user's input. And also the first test creates a journey view without elements being passed since that's optional.

You can disregard this, I didn't notice that your functions like `getStringPointerFromInterface` handled nil values. Everything should be fine as is

PR #1075, opened 2024-06-06
fix api call strategy

HemanthDogiparthi12:


carnellj-genesys (approved):

kavinbalagen (approved):
looks good to me, maybe we can log the response in the error log when we are calling the API and checking for error.

PR #1083, opened 2024-06-10
Feat/devtooling 618

shrutisuryawanshigenesys:
zip the Exporter Artifacts in CX code


jenkins/tests/Jenkinsfile
```diff
@@ -679,6 +679,17 @@ pipeline {
...
+ 
+ def extractReason(String log) {
+    // Extract reason from the log
+    def reason = ""
+    def reasonPattern = /message="([^"]*)">/
+    def matcher = log =~ reasonPattern
+    if (matcher.find()) {
+        reason = matcher.group(1).trim()
+    }
+    return reason
```
charliecon, line 692: If you merge the latest version of dev into this branch, it should remove these changes. This method `extractReason` is already defined in this file , so it shouldn't be here. (If merging dev into this branch doesn't get rid of one of the methods, you can just make sure they are the same and manually remove one.)

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -479,6 +483,49 @@ func (g *GenesysCloudResourceExporter) generateOutputFiles() diag.Diagnostics {
 		}
 	}
 
+	if _, ok := g.d.GetOk("compress"); ok { //if true, compress directory name of where the export is going to occur
```
charliecon, line 486: Here is the description in the [docs for GetOk](https://pkg.go.dev/github.com/hashicorp/terraform/helper/schema#ResourceData.GetOk):

> GetOk returns the data for the given key and whether or not the key has been set to a non-zero value at some point.
>The first result will not necessarilly be nil if the value doesn't exist. The second result should be checked to determine this information.

I think the way you have it might actually work, since the second value indicates if `compress` has been set to a non zero value (which is true in the case of bools). Still the "at some point" confuses things a bit, and since the compress field has a `Default` value, there is no need to check if compress has been set to a value by the user.

I would use:

```golang
if compress := d.Get("compress").(bool); compress {
    // ...
}
```

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -479,6 +483,49 @@ func (g *GenesysCloudResourceExporter) generateOutputFiles() diag.Diagnostics {
...
+					log.Fatalln(ferr)
+				}
+				defer file.Close()
+
+				if _, ferr = io.Copy(w, file); ferr != nil {
+					log.Fatalln(ferr)
+				}
+			}
+		}
+		zipWriter.Close()
```
charliecon, line 524: I would move all the above into its own function. ~40 lines is a lot to add to the function `generateOutputFiles`. 

I would also return `diag.Diagnostics` from that new function, and anywhere where you have `log.Fatalln` or `panic` , change it to `return diag.Errorf("...")`. If the new function returns an error, `generateOutputFiles` should then return it. We don't panic/exit anywhere in our code. We try to pass the error back up the tree

charliecon (changes requested):

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -134,6 +140,13 @@ func ResourceTfExport() *schema.Resource {
 				Default:     true,
 				ForceNew:    true,
 			},
+			"compress": {
+				Description: "Flag to check if compression is allowed",
```
carnellj-genesys, line 144: I would rename this description to "Compress exported results using zip format"

docs/resources/tf_export.md
```diff
@@ -60,6 +60,7 @@ resource "genesyscloud_tf_export" "export" {
 
 ### Optional
 
+- `compress` (Boolean) Flag to check if compression is allowed Defaults to `false`.
```
carnellj-genesys, line 63: We should add a test case to test an export and then make sure it is zipped (e.g. with a non-zero zip file"

charliecon (approved):

HemanthDogiparthi12 (approved):

PR #1094, opened 2024-06-11
Feat/Fully remove unsupported attributes from journey segments

tsantos-genesys:
This PR https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1070 removed a few unsupported attributes from journey segments, but left one still there, as well as leaving some duplicated tests that didn't serve much purpose with all the fields gone.

Since that PR already removed some fields, I'm putting this one up to remove the other one (scope) and rework the tests that use that resource, trying to simplify them a bit.

I ran all the TestAccResourceJourneySegment* and TestAccResourceJourneyActionMap* tests, since they're the only ones that should have been affected by my changes here.

Apologies if I didn't follow the proper PR protocol here, please let me know if I made a mistake and I'll update it.


genesyscloud/resource_genesyscloud_journey_segment.go
```diff
@@ -52,18 +52,10 @@ var (
 				return r
 			}(), ""),
 		},
-		"scope": {
-			Description:  "The target entity that a segment applies to.Valid values: Session, Customer. Changing the scope attribute will cause the existing journey_segment to be dropped and recreated with new ID.",
-			Type:         schema.TypeString,
-			Required:     true,
-			ForceNew:     true, // scope can be only set during creation
-			ValidateFunc: validation.StringInSlice([]string{"Session"}, false),
```
tsantos-genesys, line 60: As mentioned, in the previous PR the value "Customer" was removed.

The API only allows Session, and the field is not required. Since only session scoped segments exist today, there's no need to declare the scope at all.

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -106,36 +107,3 @@ func testVerifyJourneySegmentsDestroyed(state *terraform.State) error {
 	// Success. All Journey segment destroyed
 	return nil
 }
-
-func customerSegmentationIsSupported(t *testing.T) (bool, string) {
```
tsantos-genesys, line 110: This was used before the previous PR to skip some tests, is no longer being used at all, so I'm removing it.

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -56,30 +51,36 @@ func setupJourneySegment(t *testing.T, testCaseName string) {
...
 	journeyApi := platformclientv2.NewJourneyApiWithConfig(sdkConfig)
 
-	pageCount := 1 // Needed because of broken journey common paging
-	for pageNum := 1; pageNum <= pageCount; pageNum++ {
-		const pageSize = 100
+	segmentsToDelete := make([]platformclientv2.Journeysegment, 0)
+
+	// go through all segments to find those to delete
+	const pageSize = 200
+	for pageNum := 1; ; pageNum++ {
```
tsantos-genesys, line 58: Fetching all segments first then deleting the needed ones, as opposed to deleting and fetching.

I think with the previous code it could have bugged if trying to delete segments at the end of a page, then fetching the next page. It would have skipped a few segments.

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -16,12 +15,8 @@ import (
 	"github.com/mypurecloud/platform-client-sdk-go/v130/platformclientv2"
 )
 
-func TestAccResourceJourneySegmentCustomer(t *testing.T) {
-	runResourceJourneySegmentTestCase(t, "basic_customer_attributes")
```
tsantos-genesys, line 20: This test was supposed to be related to customer scope. There's no scope anymore, so this was doing the same as the test below.

Got rid of it and left just one in place.

test/data/data_source/genesyscloud_journey_action_map/find_by_name/01_find_by_name.tf
```diff
@@ -19,7 +19,6 @@ resource "genesyscloud_journey_action_map" "terraform_test_-TEST-CASE-" {
 resource "genesyscloud_journey_segment" "terraform_test_-TEST-CASE-_action_map_dependency" {
   display_name            = "terraform_test_-TEST-CASE-_action_map_dependency"
   color                   = "#008000"
-  scope                   = "Session"
```
tsantos-genesys, line 22: Most of the files below this point just contain the removal of scope from test files.

A few of them I deleted the whole file altogether, because they weren't doing anything meaningful, given all the removed fields.

For example, there was a test step 04_update_remove_assignment_expirition_days.tf, but expiration days were removed in the last PR, so no need for it anymore.

genesyscloud/resource_genesyscloud_journey_segment.go
tsantos-genesys: If you consider this a breaking change because the field was required, let me know and I'll rework this PR to just update the tests / descriptions.

charliecon: I think it would better to add a deprecation message stating that the field is no longer in use by the API, won't do anything, and will be removed in a future version. At least that way we might be giving some context to any users that have implemented it in a system already.

charliecon (approved):
Thanks @tsantos-genesys - this is really great! The attention to detail when cleaning up the code & tests is much appreciated. 

I just have one comment one the removal of the field "scope". Besides that, all looks good

tsantos-genesys:
Thank you @charliecon , I agree with your suggestion.

I'll decline this PR and put up a new one, to make the changes easier to check.

PR #1096, opened 2024-06-12
increased member_division_ids maxItems to 1000

igorgordienko:
increased member_division_ids maxItems to 1000
divisions is an adjustable limit and can be increased as it is in my case.


dginty4 (approved):

carnellj-genesys:
While the number of divisions you can create is apparently configurable, the number of divisions you can assign to skill groups is a hard limit of 50.  I would work with product management to see if this can be increased in the long run, but for now this limitation of 50 should be kept in place in the Terraform provider.  Ping me if you have questions Ihor.  Thanks

PR #1099, opened 2024-06-12
refactor/DEVTOOLING-158 idp_gsuite

shrutisuryawanshigenesys:
Refactor idp_gsuite into a new package structure


dginty4 (changes requested):
It looks like you added the attributes `name`, `slo_binding` and `slo_uri` but your not actually using them anywhere. Make sure to be reading and setting these values where needed and add them to the test.

carnellj-genesys (approved):

genesyscloud/idp_gsuite/genesyscloud_idp_gsuite_proxy.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// deleteIdpGsuiteFn is an implementation function for deleting a Genesys Cloud idp gsuite
+func deleteIdpGsuiteFn(ctx context.Context, p *idpGsuiteProxy, id string) (response *platformclientv2.APIResponse, err error) {
+	_, resp, err := p.identityProviderApi.DeleteIdentityprovidersGsuite()
+	if err != nil {
+		return resp, fmt.Errorf("Failed to delete idp gsuite: %s", err)
+	}
+
+	return resp, nil
```
charliecon, line 87: No need to check the error or add context here (context will be added in the main delete function.) You can just `return resp, err` directly

charliecon (approved):
Same here about Declan's comments. Also there are some merge conflicts

charliecon (approved):

PR #1103, opened 2024-06-14
refactor/Tidy up journey_segment resource tests/doc

tsantos-genesys:
Follow up from declined PR https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1094.

This PR https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1070 removed a few unsupported attributes from journey segments, but left some duplicated tests that didn't serve much purpose with all the fields gone.

This pr is reworking the tests that use that resource, trying to simplify them a bit.

I ran all the TestAccResourceJourneySegment* and TestAccResourceJourneyActionMap* tests, since they're the only ones that should have been affected by my changes here.

Apologies if I didn't follow the proper PR protocol here, please let me know if I made a mistake and I'll update it.


genesyscloud/resource_genesyscloud_journey_segment.go
```diff
@@ -53,7 +53,7 @@ var (
 			}(), ""),
 		},
 		"scope": {
-			Description:  "The target entity that a segment applies to.Valid values: Session, Customer. Changing the scope attribute will cause the existing journey_segment to be dropped and recreated with new ID.",
+			Description:  "The target entity that a segment applies to. Valid values: Session",
```
tsantos-genesys, line 56: Just correcting the description and leaving the field in place.

Customer scoped segments will be rechartered, and when the time comes, added back here.

No impact to existing customers.

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -16,12 +15,8 @@ import (
 	"github.com/mypurecloud/platform-client-sdk-go/v130/platformclientv2"
 )
 
-func TestAccResourceJourneySegmentCustomer(t *testing.T) {
-	runResourceJourneySegmentTestCase(t, "basic_customer_attributes")
```
tsantos-genesys, line 20: This test was supposed to be related to customer scope. There's no scope anymore, so this was doing the same as the test below.

Got rid of it and left just one in place.

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -56,30 +51,36 @@ func setupJourneySegment(t *testing.T, testCaseName string) {
...
 	journeyApi := platformclientv2.NewJourneyApiWithConfig(sdkConfig)
 
-	pageCount := 1 // Needed because of broken journey common paging
-	for pageNum := 1; pageNum <= pageCount; pageNum++ {
-		const pageSize = 100
+	segmentsToDelete := make([]platformclientv2.Journeysegment, 0)
+
+	// go through all segments to find those to delete
+	const pageSize = 200
+	for pageNum := 1; ; pageNum++ {
```
tsantos-genesys, line 58: Fetching all segments first then deleting the needed ones, as opposed to deleting and fetching.

I think with the previous code it could have bugged if trying to delete segments at the end of a page, then fetching the next page. It would have skipped a few segments.

genesyscloud/resource_genesyscloud_journey_segment_test.go
```diff
@@ -106,36 +107,3 @@ func testVerifyJourneySegmentsDestroyed(state *terraform.State) error {
 	// Success. All Journey segment destroyed
 	return nil
 }
-
-func customerSegmentationIsSupported(t *testing.T) (bool, string) {
```
tsantos-genesys, line 110: This was used before the previous PR to skip some tests, is no longer being used at all, so I'm removing it.

charliecon (approved):
In your last PR, you were removing the lines that build and read the scope value from the API (getting the scope from the schema.ResourceData, setting it, etc.) You can introduce these changes in this PR also. I would just recommend adding a Deprecated message to the scope field to let the user know it won't do anything.  

Here is an example of what I mean - [genesyscloud/resource_genesyscloud_widget_deployment.go#L132](https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/dev/genesyscloud/resource_genesyscloud_widget_deployment.go#L132)

carnellj-genesys (approved):

PR #1111, opened 2024-06-19
no jira: Fixes #1086 + some minor refactoring

### LINKED ISSUE #1086: Random Data Resource loading issue with genesyscloud_routing_skill_group
I use Data Source to load skill for a queue.
I always have multiple Queues to be created which depends on multiple skill groups.
I first apply the skill group with Terraform using GC Provider.
When I use Data Source  of skill group during queue creation, It fails to load some of the skill groups even if it exists in UI.
E.g there there were total new 6 skill groups created through TF.
Later when using this for creating queue, 3 of them were identifiable, rest 3 its says "Error: context deadline exceeded"
I check for the spelling and its correct.
I tried to push this to a different org , then issue persists but this time , issue was with different set of skills groups.
If I hard code the id I got from state and comment the data source call, it works.
I tried different versions from 1.33 to 1.37 issue persists
```
data "genesyscloud_routing_skill_group" "Test123" { name = "Test123" }
skill_groups       = ["${genesyscloud_routing_skill_group.Test123.id}"]
```

In the logs it shows something like below
`data.genesyscloud_routing_skill_group.Test123: Still reading... [6m0s elapsed]`
Many retries with below
` Retry #1 for GET https://api.usw2.pure.cloud/api/v2/routing/skillgroups:`
Many 429 errors
`Response 429 Too Many Requests for request:GET https://api.usw2.pure.cloud/api/v2/routing/skillgroups:`

charliecon:
There were some mistakes in the skill group data source logic and in the exporter. The SDK methods must have been released since that code was written, so I converted from raw HTTP to using the SDK.

### DIFF
--- a/genesyscloud/data_source_genesyscloud_routing_skill_group.go
+++ b/genesyscloud/data_source_genesyscloud_routing_skill_group.go
@@ -2,18 +2,14 @@ package genesyscloud
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"net/http"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v131/platformclientv2"
 	"terraform-provider-genesyscloud/genesyscloud/provider"
 	"terraform-provider-genesyscloud/genesyscloud/util"
 	"time"
-
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
-
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/mypurecloud/platform-client-sdk-go/v131/platformclientv2"
 )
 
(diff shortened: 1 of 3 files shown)


dginty4 (approved):

genesyscloud/data_source_genesyscloud_routing_skill_group.go
```diff
@@ -31,52 +27,74 @@ func dataSourceRoutingSkillGroup() *schema.Resource {
...
 
-			allSkillGroups := &AllSkillGroups{}
+func getAllSkillGroupsByName(api *platformclientv2.RoutingApi, name string) ([]platformclientv2.Skillgroupdefinition, *platformclientv2.APIResponse, error) {
+	const pageSize = 100
+	var (
+		after          string
+		allSkillGroups []platformclientv2.Skillgroupdefinition
+	)
+
+	for {
```
HemanthDogiparthi12, line 74: is it not worth looking for pagecount and then looping through ?

some APIs do not properly return the Entities size object

HemanthDogiparthi12 (approved):

genesyscloud/data_source_genesyscloud_routing_skill_group.go
charliecon: This API doesn't return a page count. If there is another page, nextUri will exist in the response, if it is the last page, it won't. That is the only indicator

carnellj-genesys (approved):

PR #1112, opened 2024-06-20
Bug/devtooling 606 - Managed sites data source 

BrianMoyles:
Due to https://inindca.atlassian.net/browse/DEVTOOLING-606, this PR:
- Removes managed attribute from sites data source 
- Automatically sets managed sites to data on export 
- Searches both managed and unmanaged sites on getSiteByName function call
Due to import cycles, the genesyscloud_resource_exporter will read the DataSourceManagedSites on export to determine which resources should be exported as data


genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -1552,6 +1551,11 @@ func (g *GenesysCloudResourceExporter) resourceIdExists(refID string, existingRe
 }
 
 func (g *GenesysCloudResourceExporter) isDataSource(resType string, name string) bool {
+	for _, element := range telephony_providers_edges_site.DataSourceManagedSites {
```
HemanthDogiparthi12, line 1554: this loop will increase linearly ("telephony_providers_edges_site.DataSourceManagedSites"),
 for every resource we might need to add as a datsource in the future. 
 
 suggest to create a datasource slice in exporter and refer it in telephony_providers_edges_site and make sure you lock it when appending data to it. this way we need not come to this place everytime we introduce a new resource  based on attribute as datasource.

genesyscloud/telephony_providers_edges_site/resource_genesyscloud_telephony_providers_edges_site.go
```diff
@@ -43,8 +45,13 @@ func getAllSites(ctx context.Context, sdkConfig *platformclientv2.Configuration)
 	}
 	for _, managedSite := range *managedSites {
 		resources[*managedSite.Id] = &resourceExporter.ResourceMeta{Name: *managedSite.Name}
+		// When exporting managed sites, they must automatically be exported as data source 
+		// Managed sites are added to the ExportAsData []string in resource_exporter and formatted correctly for export as data source
+		if tfexporter_state.IsExporterActive() {
+			managedSiteDataSource := resourceName + "::" + strings.ReplaceAll(*managedSite.Name, " ", "_")
```
bbbco, line 51: Instead of `strings.ReplaceAll` shouldn't you use `SanitizeResourceName()`?

BrianMoyles:
New PR created due to issues

PR #1116, opened 2024-06-21
Merge unit tests & fix failing testcases

monishapadmavathi:
This PR includes,

- Adding Unit testcases to the pipeline
- Fix the failing testcases that were added newly


genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone_test.go
```diff
@@ -532,3 +530,26 @@ func deleteDidPool(config *platformclientv2.Configuration, id string) error {
...
+		}
+
+		for _, didPool := range *didPools.Entities {
+			if (didPool.StartPhoneNumber != nil && *didPool.StartPhoneNumber == number) ||
+				(didPool.EndPhoneNumber != nil && *didPool.EndPhoneNumber == number) {
+				edgesAPI.DeleteTelephonyProvidersEdgesDidpool(*didPool.Id)
+				time.Sleep(5 * time.Second)
+			}
+		}
+	}
```
charliecon, line 554: In another PR, a dev made a good point that its a bad idea to paginate and delete entities at the same time because the pages are going to change as you paginate through them. It's better to loop through the entities and add all to a list, and then after all have been collected that match the condition, loop through them and do the deletions

charliecon (approved):
Just one comment

dginty4 (approved):

genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone_test.go
HemanthDogiparthi12: +1

monishapadmavathi: Modified it to collect list of didpool first, then delete.

PR #1121, opened 2024-06-25
feat/COBROWSE-2427 pause field added

Anpeople:


carnellj-genesys (approved):

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
```diff
@@ -356,6 +356,26 @@ var (
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
+			"pause_criteria": {
+				Description: "List of pause criteria",
```
charliecon, line 360: The public API description is more useful here I think - "Pause criteria that will pause cobrowse if some of them are met in the user's URL". The docs will then indicate that it is a list

go.mod
```diff
@@ -1,6 +1,8 @@
 module terraform-provider-genesyscloud
 
-go 1.20
+go 1.21
+
+toolchain go1.22.3
```
charliecon, line 5: You'll have to undo this. `toolchain` causes problems with the tool we use to release, so we have to keep the go verison the same. Sometimes this update to go.mod happens automatically based on the go version you have installed and the value of the env var GOROOT

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_test.go
```diff
@@ -1043,8 +1067,9 @@ func generateWebDeploymentConfigCobrowseSettings(cbEnabled, cbAllowAgentControl
 		channels = [ %s ]
 		mask_selectors = [ %s ]
 		readonly_selectors = [ %s ]
+		pause_criteria = [ %s ]
 	}
```
charliecon, line 1071: We define items in a schema list like this: 

```golang
pause_criteria { // item one in the list
  url_fragment = ""
  condition   = ""
}

pause_criteria { // item two 
  url_fragment = ""
  condition   = ""
}
```

Here is an example of we usually use functions to Generate list items - https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/outbound_contact_list_contact/resource_genesyscloud_outbound_contact_list_contact_utils.go#L144

We have a parent `generateXYZ` resource whose final parameter is `nestedBlocks ...string,`. This allows you to pass as many calls to a function like `generatePauseCriteria`, that will generate the nested schema objects.

e.g.

```golang
var config = generateWebdeploymentConfigResource(
  resourceId,
  name,
  generatePauseCriteria(
    "url 1",
    "condition 1",
  ),
  generatePauseCriteria(
    "url 2",
    "condition 2",
  ),
)
```

How you have done it might be syntactically correct but it's a pretty big departure from the convention in this project, so it would be good to align it.

examples/resources/genesyscloud_webdeployments_configuration/resource.tf
```diff
@@ -75,6 +75,7 @@ resource "genesyscloud_webdeployments_configuration" "exampleConfiguration" {
     channels               = ["Webmessaging", "Voice"]
     mask_selectors         = [".my-class", "#my-id"]
     readonly_selectors     = [".my-class", "#my-id"]
+    pause_criteria         = []
```
charliecon, line 78: You can update this to 

```hcl
pause_criteria {
  url_fragment = "/sensitive"
  condition    = "includes"
}
```

charliecon (changes requested):
Be sure to re-run `go generate` from the root of the project after these changes are made

examples/resources/genesyscloud_webdeployments_configuration/resource.tf
Anpeople: pause_criteria         = [{
                                url_fragment = "/sensitive"
                                condition    = "includes"
                              }]
probably more correct, because its list

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
Anpeople: done

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_test.go
Anpeople: you want me to do something like 

func generateWebDeploymentConfigCobrowseSettings(cbEnabled, cbAllowAgentControl string, cbAllowAgentNavigation string, cbChannels []string, cbMaskSelectors []string, cbReadonlySelectors []string, pauseCriteriaBlocks ...string,) string {

	return fmt.Sprintf(`
	cobrowse {
		enabled = %s
		allow_agent_control = %s
		allow_agent_navigation = %s
		channels = [ %s ]
		mask_selectors = [ %s ]
		readonly_selectors = [ %s ]
		%s
	}
`, cbEnabled, cbAllowAgentControl, cbAllowAgentNavigation, strings.Join(cbChannels, ", "), strings.Join(cbMaskSelectors, ", "), strings.Join(cbReadonlySelectors, ", "), strings.Join(pauseCriteriaBlocks, "\n"))
}

func generatePauseCriteria(urlFragment, condition string) string {
	return fmt.Sprintf(`pause_criteria {
	url_fragment = "%s"
	condition = "%s"
}`, urlFragment, condition)
}
?

examples/resources/genesyscloud_webdeployments_configuration/resource.tf
charliecon: Nope. We actually had issues defining lists as in your example - terraform couldn't detect when a nested field was required or not. 

It is a bit strange, but that's how you define schema list items in HCL. Even if a json request body contains a field that is a map, we have to define it as a list of maps in the schema and set MaxItems: 1 to make sure the user doesn't try to define multiple.

go.sum
```diff
@@ -11,6 +11,7 @@ cloud.google.com/go/firestore v1.1.0/go.mod h1:ulACoGHTpvq5r8rxGJ4ddJZBZqakUQqCl
 cloud.google.com/go/pubsub v1.0.1/go.mod h1:R0Gpsv3s54REJCy4fxDixWD93lHJMoZTyQ2kNxGRt3I=
 cloud.google.com/go/storage v1.0.0/go.mod h1:IhtSnM/ZTZV8YYJWCY8RULGVqBDmpoyjwiyrjsg+URw=
```
charliecon, line 12: `go mod tidy` should clean up the changes below

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
Anpeople: I didnt know about go generate! great thing : D I did it manually

go.sum
Anpeople: I did it but its still here

Anpeople: if its critical, I could remove it manually

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_test.go
charliecon: Exactly 👍

go.sum
charliecon: It's alright. I'll check it out after we merge to dev. Maybe these changes should be here

charliecon (approved):

PR #1123, opened 2024-06-26
Cleanup/Use sdk instead of direct calls to deal in routing_utilization

tsantos-genesys:
Back when utilization labels were added to cx as code, some attributes weren't public in public-api, so they weren't in the sdk. To make the calls works, we switched to direct calls to public-api, instead of using sdk.

Now the feature is released, we can go back to using the sdk, simplifying the provider.

I'll call it out in the PR, there's a small amount of duplication here, because I'm only cleaning up routing_utilization and not touching user, in the interest of keeping the PR small.

I'll clean up user immediately after this PR, getting rid of said duplication.

### DIFF
--- a/genesyscloud/routing_utilization/genesyscloud_routing_utilization_proxy.go
+++ b/genesyscloud/routing_utilization/genesyscloud_routing_utilization_proxy.go
@@ -2,19 +2,15 @@ package routing_utilization
 
 import (
 	"context"
-	"fmt"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/mypurecloud/platform-client-sdk-go/v133/platformclientv2"
 )
 
 var internalProxy *routingUtilizationProxy
 
-type getRoutingUtilizationFunc func(ctx context.Context, p *routingUtilizationProxy) (*platformclientv2.APIResponse, error)
+type getRoutingUtilizationFunc func(ctx context.Context, p *routingUtilizationProxy) (*platformclientv2.Utilizationresponse, *platformclientv2.APIResponse, error)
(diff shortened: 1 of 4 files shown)


genesyscloud/routing_utilization/genesyscloud_routing_utilization_proxy.go
```diff
@@ -46,31 +38,20 @@ func getRoutingUtilizationProxy(clientConfig *platformclientv2.Configuration) *r
...
-	return p.updateDirectlyAttr(ctx, p, d, utilizationRequest)
-}
-
-// Calling the Utilization API directly while the label feature is not available.
-// Once it is, this code can go back to using platformclientv2's RoutingApi to make the call.
-func getRoutingUtilizationFn(ctx context.Context, p *routingUtilizationProxy) (*platformclientv2.APIResponse, error) {
-	apiClient := &p.routingApi.Configuration.APIClient
-	path := fmt.Sprintf("%s/api/v2/routing/utilization", p.routingApi.Configuration.BasePath)
-	headerParams := buildHeaderParams(p.routingApi)
-	resp, err := apiClient.CallAPI(path, "GET", nil, headerParams, nil, nil, "", nil)
```
tsantos-genesys, line 69: All of these direct calls are no longer needed.

genesyscloud/routing_utilization/resource_routing_utilization_utils.go
```diff
@@ -68,15 +129,16 @@ func FlattenUtilizationSetting(settings MediaUtilization) []interface{} {
 	return []interface{}{settingsMap}
 }
 
-func FilterAndFlattenLabelUtilizations(labelUtilizations map[string]LabelUtilization, originalLabelUtilizations []interface{}) []interface{} {
+// To be removed once resource_user is refactored to use FilterAndFlattenLabelUtilizations instead
```
tsantos-genesys, line 132: This is the small bit of duplication I mentioned, these Internal functions.

Like I said, I'll get rid of all of them in my next PR, just separated the work to keep each PR smaller and easier to review.

charliecon (approved):

PR #1124, opened 2024-06-26
feature/devtooling 608

kavinbalagen:


genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone_schema.go
```diff
@@ -30,6 +30,24 @@ func SetRegistrar(l registrar.Registrar) {
 
 // ResourcePhone registers the genesyscloud_telephony_providers_edges_phone resource with Terraform
 func ResourcePhone() *schema.Resource {
+	lineProperties := &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			`line_address`: {
+				Description:      `DID for standalone phones. Each phone number must be in an E.164 phone number format.`,
+				Optional:         true,
```
kavinbalagen, line 37: line_address is required for each line when the parent field lineProperites is specified otherwise line_address is automatically computed if lineProperites is not set. However terraform will not allow are key to be computed and required at the same time, hence i have left it as optional, the user will get an error if they try to update the remote_address without a line_address

carnellj-genesys (approved):

PR #1132, opened 2024-06-28
Bug/Devtooling-677: Error Migrating routes to independent resource

BrianMoyles:
When migrating to independent outbound routes resource, the error 
`OutboundRoute with name 'Default Outbound Route' already exists` can occur
This PR checks the existing routes within a site and only creates any new sites, to prevent the error.
This allows users to migrate existing routes, create new routes, or do both at the same time

### DIFF
--- a/genesyscloud/telephony_providers_edges_site_outbound_route/resource_genesyscloud_telephony_providers_edges_site_outbound_route.go
+++ b/genesyscloud/telephony_providers_edges_site_outbound_route/resource_genesyscloud_telephony_providers_edges_site_outbound_route.go
@@ -4,10 +4,14 @@ import (
 	"context"
 	"fmt"
 	"log"
+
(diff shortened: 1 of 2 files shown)


dginty4 (approved):

charliecon (commented):
Should the user not have to remove routes from the sites resource if they're going to migrate to the new outbound_routes resource? Otherwise what's the point of controlling them with a separate resource? Seems like the error is right, and they should make those changes. I might be missing something here

dginty4:
> Should the user not have to remove routes from the sites resource if they're going to migrate to the new outbound_routes resource? Otherwise what's the point of controlling them with a separate resource? Seems like the error is right, and they should make those changes. I might be missing something here

I don't think customer will like having to delete routes from a site and then recreating them just to move to the new independent resource. This could also cause problems if they have made references to existing routes outside of terraform

charliecon (approved):

PR #1135, opened 2024-06-28
Fix/devtooling 633

bbbco:
Fixes https://inindca.atlassian.net/browse/DEVTOOLING-633 by adding a new resource `genesyscloud_outbound_contact_list_template` and adding logic to the `genesyscloud_outbound_contactlistfilter` resource to figure out whether to use a contact list or a contact list template.


genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
```diff
@@ -0,0 +1,624 @@
...
+		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
+			"attempt_limit_id": {RefType: "genesyscloud_outbound_attempt_limit"},
+			"division_id":      {RefType: "genesyscloud_auth_division"},
+		},
+	}
+}
+
+func ResourceOutboundContactListTemplate() *schema.Resource {
+	return &schema.Resource{
+		Description: `Genesys Cloud Outbound Contact List Template`,
```
charliecon, line 139: To stay consistent with the over all package structure, could you put the schema definitions for the exporter, resource and data source into a file `resource_genesyscloud_outbound_contact_list_template_schema.go`. You can add the `resourceName` constant to that file as well and use it wherever the resource name string appears

genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
```diff
@@ -0,0 +1,624 @@
...
+		}
+		if sdkContactListTemplate.AutomaticTimeZoneMapping != nil {
+			_ = d.Set("automatic_time_zone_mapping", *sdkContactListTemplate.AutomaticTimeZoneMapping)
+		}
+		if sdkContactListTemplate.ZipCodeColumnName != nil {
+			_ = d.Set("zip_code_column_name", *sdkContactListTemplate.ZipCodeColumnName)
+		}
+		if sdkContactListTemplate.ColumnDataTypeSpecifications != nil {
+			_ = d.Set("column_data_type_specifications", flattenSdkOutboundContactListTemplateColumnDataTypeSpecifications(*sdkContactListTemplate.ColumnDataTypeSpecifications))
+		}
```
charliecon, line 364: You could implement the generic functions from the `resourcedata` package for most of these

genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
```diff
@@ -0,0 +1,624 @@
...
+func buildSdkOutboundContactListContactEmailAddressColumnSlice(contactEmailAddressColumn *schema.Set) *[]platformclientv2.Emailcolumn {
+	if contactEmailAddressColumn == nil {
+		return nil
+	}
+	sdkContactEmailAddressColumnSlice := make([]platformclientv2.Emailcolumn, 0)
+	contactEmailAddressColumnList := contactEmailAddressColumn.List()
+	for _, configEmailColumn := range contactEmailAddressColumnList {
+		var sdkContactEmailAddressColumn platformclientv2.Emailcolumn
+		contactEmailAddressColumnMap := configEmailColumn.(map[string]interface{})
+		if columnName := contactEmailAddressColumnMap["column_name"].(string); columnName != "" {
```
charliecon, line 460: This could throw an interface conversion panic. If `column_name` isn't in the map, `columnName` will be `nil` and casting nil to a string will throw the panic. If you write the if statement like this,

```golang
if columnName, _ := contact ...
```

and the key isn't present, the var will be the zero value (string in this case) for that type.

genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
```diff
@@ -0,0 +1,624 @@
...
+		if contactEmailAddressColumn.ColumnName != nil {
+			contactEmailAddressColumnMap["column_name"] = *contactEmailAddressColumn.ColumnName
+		}
+		if contactEmailAddressColumn.VarType != nil {
+			contactEmailAddressColumnMap["type"] = *contactEmailAddressColumn.VarType
+		}
+		if contactEmailAddressColumn.ContactableTimeColumn != nil {
+			contactEmailAddressColumnMap["contactable_time_column"] = *contactEmailAddressColumn.ContactableTimeColumn
+		}
+
```
charliecon, line 493: Could use resourcedata function `SetMapValueIfNotNil` for these

genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
```diff
@@ -0,0 +1,624 @@
...
+		if contactPhoneNumberColumn.ColumnName != nil {
+			contactPhoneNumberColumnMap["column_name"] = *contactPhoneNumberColumn.ColumnName
+		}
+		if contactPhoneNumberColumn.VarType != nil {
+			contactPhoneNumberColumnMap["type"] = *contactPhoneNumberColumn.VarType
+		}
+		if contactPhoneNumberColumn.CallableTimeColumn != nil {
+			contactPhoneNumberColumnMap["callable_time_column"] = *contactPhoneNumberColumn.CallableTimeColumn
+		}
+
```
charliecon, line 444: `SetMapValueIfNotNil`

genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
```diff
@@ -0,0 +1,624 @@
...
+		contactPhoneNumberColumnMap := configPhoneColumn.(map[string]interface{})
+		if columnName := contactPhoneNumberColumnMap["column_name"].(string); columnName != "" {
+			sdkContactPhoneNumberColumn.ColumnName = &columnName
+		}
+		if varType := contactPhoneNumberColumnMap["type"].(string); varType != "" {
+			sdkContactPhoneNumberColumn.VarType = &varType
+		}
+		if callableTimeColumn := contactPhoneNumberColumnMap["callable_time_column"].(string); callableTimeColumn != "" {
+			sdkContactPhoneNumberColumn.CallableTimeColumn = &callableTimeColumn
+		}
```
charliecon, line 419: Same thing about conversion panic

genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
```diff
@@ -0,0 +1,624 @@
...
+
+	return columnDataTypeSpecificationsSlice
+}
+
+// type OutboundContactListInstance struct{
+// }
+
+// func (*OutboundContactListInstance) ResourceOutboundContactList() *schema.Resource {
+// 	ResourceOutboundContactList() *schema.Resource
+// }
```
charliecon, line 567: Leftover comments

charliecon (changes requested):
I think you are missing the example resource/datasource. Also, could you implement a proxy file for this package?

bbbco:
Sure @charliecon I'll make these changes. FYI, I just did a straight copy from the `outbound_contact_list` resource for this new resource, so I'll make the same recommendation changes to that resource at the same time.

genesyscloud/outbound_contact_list_template/resource_genesyscloud_outbound_contact_list_template.go
bbbco: I just put the `resourceName` in the init file

charliecon (approved):
The example `genesyscloud_outbound_contact_list_template` data source is still missing, and you'll need to run `go generate` to render the examples inside `/docs/`. 

Besides that, looks good to go 👌

bbbco:
> The example `genesyscloud_outbound_contact_list_template` data source is still missing, and you'll need to run `go generate` to render the examples inside `/docs/`.
> 
> Besides that, looks good to go 👌

Thanks @charliecon ! I've added the missing data source example and ran `make docs` to generate the docs.
Thanks for the solid review!

PR #1137, opened 2024-07-01
feat/GPE-14463 add support for App stream type on Journey resources

michalk-genesys:
When configuring Journey resources like segments, outcomes or action maps, it is now possible to use App stream type. Also, the documentation around supported session types was updated, to specify the allowed values.


dginty4 (approved):

dginty4:
@michalk-genesys there's conflicts that need to be resolved. I think you might have branched off main instead of dev when you done your work

PR #1145, opened 2024-07-04
Feat/devtooling-645: Add roles_enabled to group resource

BrianMoyles:


dginty4 (approved):

genesyscloud/group/resource_genesyscloud_group_schema.go
```diff
@@ -120,6 +121,11 @@ func ResourceGroup() *schema.Resource {
 				Computed:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
+			"roles_enabled": {
+				Description: "Allow roles to be assigned to this group.",
+				Type:        schema.TypeBool,
+				Optional:    true,
```
HemanthDogiparthi12, line 127: default value ?

genesyscloud/group/resource_genesyscloud_group_schema.go
```diff
@@ -120,6 +121,12 @@ func ResourceGroup() *schema.Resource {
...
+				Default:     true,
```
charliecon, line 128: I might be over-reviewing, but the zero value of a bool in go is false, so I would just leave this out and go with that. That's the pattern we follow with a lot of fields across different resources. There's no need for a Default value anyway since golang won't allow for a nil value in for a boolean field (it's gonna be `false`), and the API doesn't specify that it should be `true` for any reason

charliecon (approved):

genesyscloud/group/resource_genesyscloud_group_schema.go
charliecon: Plus the user could expect that omitting the field will mean `false` since that's the pattern in other resources. But with this one all the sudden they have to explicitly provide `false`

dginty4: I think we should maybe have the default as true. If we default to false, once this goes out to customers, I assume their groups will no longer have roles (unless the customer catches it). This could cause problems if customers are doing role assignment via a group which from what I understand is the recommended way to assign roles.

PR #1150, opened 2024-07-10
Feat/devtooling 655: conversations_messaging_settings resource

BrianMoyles:


carnellj-genesys (approved):

genesyscloud/conversations_messaging_settings/resource_conversations_messaging_settings.go
```diff
@@ -0,0 +1,129 @@
...
+	"terraform-provider-genesyscloud/genesyscloud/util/resourcedata"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v133/platformclientv2"
+)
+
+func getAllAuthConversationsMessagingSettingss(ctx context.Context, clientConfig *platformclientv2.Configuration) (resourceExporter.ResourceIDMetaMap, diag.Diagnostics) {
```
charliecon, line 21: Settingss -> Settings

genesyscloud/conversations_messaging_settings/resource_conversations_messaging_settings.go
```diff
@@ -0,0 +1,129 @@
...
+	proxy := getConversationsMessagingSettingsProxy(clientConfig)
+	resources := make(resourceExporter.ResourceIDMetaMap)
+
+	messagingSettings, resp, err := proxy.getAllConversationsMessagingSettings(ctx)
+	if err != nil {
+		return nil, util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to get Conversations messaging Settings: %s", err), resp)
+	}
+
+	for _, messagingSetting := range *messagingSettings {
+		log.Printf("Dealing with messaging setting id : %s", *messagingSetting.Id)
```
charliecon, line 31: No need for this log statement. We found before that it clogs up the logs a good bit when you're running an export. Better to just provide all the details when one fails

genesyscloud/conversations_messaging_settings/resource_conversations_messaging_settings_schema.go
```diff
@@ -0,0 +1,161 @@
...
+				Elem:        contentSettingResource,
+			},
+			"event": {
+				Description: "Settings relating to events which may occur",
+				Optional:    true,
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Elem:        eventSettingResource,
+			},
+		},
```
charliecon, line 134: If i try to create one of these with only the name field, I see this error

```json
{
  "message": "Config field is required",
  "code": "invalid.messaging.setting",
  "status": 400,
  "messageWithParams": "Config field is required",
  "contextId": "2252da7f-e244-4a8a-a6a6-5ef3186da8d5",
  "details": [],
  "errors": []
}
```

I don't see any field "config", but it looks like it works if you provide either `content` or `event` (and the nested values until you eventually provide a value for `inbound` or `outbound`). I would use the attribute `ExactlyOneOf`. There is an example of it being used in `resource_genesyscloud_outbound_contactlistfilter_schema.go`

charliecon (commented):
Missing example data source and resource. One small merge conflict as well

charliecon (approved):

PR #1155, opened 2024-07-12
Bug/devtooling 637: Conditional group routing rules 

charliecon:


genesyscloud/routing_queue_conditional_group_routing/genesyscloud_routing_queue_conditional_group_routing_proxy.go
```diff
@@ -78,13 +86,13 @@ func getRoutingQueueConditionRoutingFn(ctx context.Context, p *routingQueueCondi
 		return queue.ConditionalGroupRouting.Rules, resp, nil
 	}
 
-	return nil, resp, fmt.Errorf("no conditional group routing rules found for queue %s", queueId)
+	return nil, resp, nil
```
charliecon, line 89: This was the main cause of the bug. There are scenarios where TF will read a queue that has no CGR rules yet, but it shouldn't fail because of that. For example, if the creation of the CGR rules fails for another reason, when I ran `terraform destroy` after that, it would fail with the error above. 

If there are no CGR rules on a queue even though there should be, Terraform will let us know, so I took out this error.

dginty4 (approved):

carnellj-genesys (approved):

PR #1156, opened 2024-07-15
feat/GPE-14606 add support for Outcome Quantile Conditions

michalk-genesys:
feat/GPE-14606 add support for Outcome Quantile Conditions when working with Action Maps

When setting up an Action Map with outcome triggers, Outcome Quantile Conditions should now be used instead of Outcome Probability Conditions.


genesyscloud/resource_genesyscloud_journey_action_map.go
```diff
@@ -636,6 +668,27 @@ func buildSdkOutcomeProbabilityCondition(outcomeProbabilityCondition map[string]
...
 
+func flattenOutcomeQuantileCondition(outcomeQuantileCondition *platformclientv2.Outcomequantilecondition) map[string]interface{} {
+	outcomeQuantileConditionMap := make(map[string]interface{})
+	outcomeQuantileConditionMap["outcome_id"] = *outcomeQuantileCondition.OutcomeId
+	outcomeQuantileConditionMap["max_quantile_threshold"] = *typeconv.Float32to64(outcomeQuantileCondition.MaxQuantileThreshold)
+	stringmap.SetValueIfNotNil(outcomeQuantileConditionMap, "fallback_quantile_threshold", typeconv.Float32to64(outcomeQuantileCondition.FallbackQuantileThreshold))
+	return outcomeQuantileConditionMap
+}
+
+func buildSdkOutcomeQuantileCondition(outcomeQuantileCondition map[string]interface{}) *platformclientv2.Outcomequantilecondition { // ???
```
jshankarc, line 679: Very minor, should we need `// ???` here?

michalk-genesys: Oh, thanks! Missed this!

EvelynHannonGenesys (approved):

jshankarc (approved):

charliecon (approved):
Looks good! Only change I would recommend - you could use the Deprecated field in the schema definition. [here's an example of it being used in another resource](https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/dev/genesyscloud/tfexporter/resource_genesyscloud_tf_export.go#L60)

This will give the end user a fancy yellow warning if they use the wrong field.

charliecon (approved):

PR #1159, opened 2024-07-16
feat/Devtooling 712 resource for supported content

shrutisuryawanshigenesys:


carnellj-genesys (approved):

genesyscloud/supported_content/genesyscloud_supported_content_proxy.go
```diff
@@ -0,0 +1,173 @@
...
+	supportedContents, resp, err := p.conversationsApi.GetConversationsMessagingSupportedcontent(pageSize, 1)
+	if err != nil {
+		return nil, resp, fmt.Errorf("Failed to get supported content: %v", err)
+	}
+	if supportedContents.Entities == nil || len(*supportedContents.Entities) == 0 {
+		return &allSupportedContents, resp, nil
+	}
+	for _, supportedContent := range *supportedContents.Entities {
+		allSupportedContents = append(allSupportedContents, supportedContent)
+	}
```
charliecon, line 114: You can replace this for loop with:

```golang
allSupportedContents = append(allSupportedContents, *supportedContents.Entities...)
```

It is not a big deal, but it is something the go linter package highlights. Eventually we'd like for that linter to run against the whole project as part of the workflow and produce zero warnings

genesyscloud/supported_content/genesyscloud_supported_content_proxy.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		if supportedContents.Entities == nil || len(*supportedContents.Entities) == 0 {
+			break
+		}
+
+		for _, supportedContent := range *supportedContents.Entities {
+			allSupportedContents = append(allSupportedContents, supportedContent)
+		}
+	}
```
charliecon, line 129: Same here

genesyscloud/supported_content/genesyscloud_supported_content_proxy.go
```diff
@@ -0,0 +1,173 @@
...
+
+// deleteSupportedContentFn is an implementation function for deleting a Genesys Cloud supported content
+func deleteSupportedContentFn(ctx context.Context, p *supportedContentProxy, id string) (response *platformclientv2.APIResponse, err error) {
+	resp, err := p.conversationsApi.DeleteConversationsMessagingSupportedcontentSupportedContentId(id)
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
+}
```
charliecon, line 173: This can be a one line function:

```golang
return p.conversationsApi.DeleteConversationsMessagingSupportedcontentSupportedContentId(id)
```

genesyscloud/supported_content/data_source_genesyscloud_supported_content.go
```diff
@@ -0,0 +1,43 @@
...
+	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
+		supportedContentId, retryable, resp, err := proxy.getSupportedContentIdByName(ctx, name)
+
+		if err != nil && !retryable {
+			retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("Error searching supported content %s: %s", name, err), resp))
+		}
+
+		if retryable {
+			retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("No supported content found with name %s", name), resp))
+		}
```
charliecon, line 38: `NonRetryableError` -> `RetryableError`

genesyscloud/supported_content/data_source_genesyscloud_supported_content_test.go
```diff
@@ -0,0 +1,62 @@
...
+	resourceId string,
+	name string,
+	dependsOn string,
+) string {
+	return fmt.Sprintf(`
+	data "genesyscloud_supported_content" "%s" {
+		name = "%s"
+		depends_on = [%s]
+	}
+	`, resourceId, name, dependsOn)
```
charliecon, line 61: You can use the resourceName var here

genesyscloud/supported_content/resource_genesyscloud_supported_content_schema.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+}
+
+// SupportedContentExporter returns the resourceExporter object used to hold the genesyscloud_supported_content exporter's config
+func SupportedContentExporter() *resourceExporter.ResourceExporter {
+	return &resourceExporter.ResourceExporter{
+		GetResourcesFunc: provider.GetAllWithPooledClient(getAllAuthSupportedContents),
+		RefAttrs:         map[string]*resourceExporter.RefAttrSettings{
+			// TODO: Add any reference attributes here
+		},
```
charliecon, line 110: You can remove this comment

genesyscloud/supported_content/resource_genesyscloud_supported_content_test.go
```diff
@@ -0,0 +1,131 @@
...
+	nestedBlocks ...string,
+) string {
+	return fmt.Sprintf(`resource "genesyscloud_supported_content" "%s" {
+		name = "%s"
+		media_types {
+			allow {
+				%s
+			}
+		}
+	} `, resourceId, name, strings.Join(nestedBlocks, "\n"))
```
charliecon, line 90: resourceName

genesyscloud/supported_content/resource_genesyscloud_supported_content_utils.go
```diff
@@ -0,0 +1,115 @@
...
+			continue
+		}
+
+		resourcedata.BuildSDKInterfaceArrayValueIfNotNil(&sdkMediaType.Allow, mediaTypesMap, "allow", buildMediaTypeAccesss)
+	}
+	return &sdkMediaType
+}
+
+// buildMediaTypeAccesss maps an []interface{} into a Genesys Cloud *[]platformclientv2.Mediatypeaccess
+func buildMediaTypeAccesss(mediaTypeAccesss []interface{}) *platformclientv2.Mediatypeaccess {
```
charliecon, line 39: `buildMediaTypeAccesss` -> `buildMediaTypeAccess`

charliecon: Also mediaTypeAccesss -> mediaTypeAccess

There are a few other instances below where there is an extra `s`

genesyscloud/supported_content/resource_genesyscloud_supported_content_test.go
```diff
@@ -0,0 +1,131 @@
...
+		outbound {
+			type="%s"
+		}	
+	`, outboundType)
+}
+
+func testVerifySupportedContentDestroyed(state *terraform.State) error {
+	supportContentApi := platformclientv2.NewConversationsApi()
+	for _, rs := range state.RootModule().Resources {
+		if rs.Type != "genesyscloud_supported_content" {
```
charliecon, line 116: `resourceName`

charliecon (changes requested):
You could implement caching in the proxy. The groups package implemented it, if you want an example to go off

genesyscloud/supported_content/resource_genesyscloud_supported_content_utils.go
```diff
@@ -0,0 +1,115 @@
...
+and unmarshal data into formats consumable by Terraform and/or Genesys Cloud.
+*/
+
+// getSupportedContentFromResourceData maps data from schema ResourceData object to a platformclientv2.Supportedcontent
+func getSupportedContentFromResourceData(d *schema.ResourceData) platformclientv2.Supportedcontent {
+	return platformclientv2.Supportedcontent{
+		Name:       platformclientv2.String(d.Get("name").(string)),
+		MediaTypes: buildMediaTypes(d.Get("media_types").([]interface{})),
+	}
+}
```
charliecon, line 21: The way this function is working, if the user only defines a value for `name`, the MediaTypes field will be defined as well but with nothing inside, resulting in the error: `Failed to create supported content: API Error: 400 - Missing the supportedContent.mediaTypes.allow field`

We want to make sure MediaTypes is nil if the user doesn't specify it in their resource. You can do this by defining `platformclientv2.Supportedcontent` as a variable 

```golang
var supportedContent platformclientv2.Supportedcontent
supportedContent.name = platformclientv2.String(d.Get("name").(string))
if mediaTypes, ok := d.Get("media_types").([]interface{}); ok && len(mediaTypes) > 0 {
  supportedContent.MediaTypes = &mediaTypes 
}
return supportedContent
```

genesyscloud/supported_content/resource_genesyscloud_supported_content_utils.go
```diff
@@ -0,0 +1,115 @@
...
+	resourcedata.SetMapInterfaceArrayWithFuncIfNotNil(mediaTypeAccessMap, "inbound", mediaTypeAccesss.Inbound, flattenMediaTypes)
+	resourcedata.SetMapInterfaceArrayWithFuncIfNotNil(mediaTypeAccessMap, "outbound", mediaTypeAccesss.Outbound, flattenMediaTypes)
+
+	mediaTypeAccessList = append(mediaTypeAccessList, mediaTypeAccessMap)
+
+	return mediaTypeAccessList
+}
+
+// flattenMediaTypess maps a Genesys Cloud *[]platformclientv2.Mediatypes into a []interface{}
+func flattenMediaTypess(mediaTypess *platformclientv2.Mediatypes) []interface{} {
```
charliecon, line 106: There is a function called `flattenMediaTypess` and another called `flattenMediaTypes`. 

Also, if you define an inbound item like this: `inbound{}` , leaving out the `type` field, you see this error: `panic: interface conversion: interface {} is nil, not map[string]interface {}`. This is because somewhere we're casting nil to a map. It's hard to tell where this is happening looking through this file though. I think the resource generator wrote this file in a way that makes it hard to read/understand. I think it's probably a good idea to just write these functions yourself so we know what the code is doing.

charliecon (changes requested):

charliecon:
@dginty4 @carnellj-genesys @HemanthDogiparthi12 What do we think of the resource name "genesyscloud_supported_content" ? The endpoint is `/api/v2/conversations/messaging/supportedcontent` , maybe "genesyscloud_conversations_messaging_supportedcontent" would be better? Its a bit long but I think it will help to keep the provider organised as it scales

dginty4:
> @dginty4 @carnellj-genesys @HemanthDogiparthi12 What do we think of the resource name "genesyscloud_supported_content" ? The endpoint is `/api/v2/conversations/messaging/supportedcontent` , maybe "genesyscloud_conversations_messaging_supportedcontent" would be better? Its a bit long but I think it will help to keep the provider organised as it scales

I would agree with you Charlie. Brians Resource for messaging settings was named `genesyscloud_conversations_messaging_settings` and is merged now so best to stick with that pattern to make it clear they are related

genesyscloud/supported_content/genesyscloud_supported_content_proxy.go
```diff
@@ -0,0 +1,177 @@
...
+
+	for _, content := range allSupportedContents {
+		rc.SetCache(p.supportedContentCache, *content.Id, content)
+	}
+
+	return &allSupportedContents, resp, nil
+}
+
+// getSupportedContentIdByNameFn is an implementation of the function to get a Genesys Cloud supported content by name
+func getSupportedContentIdByNameFn(ctx context.Context, p *supportedContentProxy, name string) (id string, retryable bool, response *platformclientv2.APIResponse, err error) {
```
HemanthDogiparthi12, line 140: for new resources better to have cache enabled . for datasource reads.

genesyscloud/conversations_messaging_supportedcontent/data_source_genesyscloud_conversations_messaging_supportedcontent.go
```diff
@@ -0,0 +1,114 @@
...
+		key = name.(string)
+	}
+
+	if d.Get("name").(string) == "" {
+		return util.BuildDiagnosticError(resourceName, "no user search field specified", nil)
+	}
+
+	if dataSourceSupportedContentCache == nil {
+		dataSourceSupportedContentCache = rc.NewDataSourceCache(sdkConfig, hydrateSupportedContentCacheFn, getSupportedContentIdByName)
+	}
```
charliecon, line 40: `name` is a required field, so you can just write `key := d.Get("name").(string)` and don't do the checks. If the user does not provide a value for it, they will see an error about it being a required field, and this code will not be reached.

genesyscloud/conversations_messaging_supportedcontent/data_source_genesyscloud_conversations_messaging_supportedcontent_test.go
```diff
@@ -0,0 +1,62 @@
...
+	resourceId string,
+	name string,
+	dependsOn string,
+) string {
+	return fmt.Sprintf(`
+	data "`+resourceName+`" "%s" {
+		name = "%s"
+		depends_on = [%s]
+	}
+	`, resourceId, name, dependsOn)
```
charliecon, line 61: You might as well format resourceName with `%s`, so it looks like 

```golang
return fmt.Sprintf(`
	data "%s" "%s" {
		name = "%s"
		depends_on = [%s]
	}
	`, resourceName, resourceId, name, dependsOn)
```

genesyscloud/conversations_messaging_supportedcontent/resource_genesyscloud_conversations_messaging_supportedcontent_test.go
```diff
@@ -0,0 +1,131 @@
...
+		CheckDestroy: testVerifySupportedContentDestroyed,
+	})
+}
+
+func GenerateSupportedContentResource(
+	resourceId string,
+	name string,
+	nestedBlocks ...string,
+) string {
+	return fmt.Sprintf(`resource "`+resourceName+`" "%s" {
```
charliecon, line 83: Same here

charliecon (approved):
Looks good! Just some small comments. There is a merge conflict on main also

PR #1164, opened 2024-07-17
Bug/devtooling 707 - Worktype status independent resource

dginty4:
Moving task management worktype status into it's own resource as we can't properly manage statuses as a nested block on the worktype resource


genesyscloud/task_management_worktype_status/data_source_genesyscloud_task_management_worktype_status.go
```diff
@@ -0,0 +1,43 @@
...
+
+/*
+   The data_source_genesyscloud_task_management_worktype_status.go contains the data source implementation
+   for the resource.
+*/
+
+// dataSourceTaskManagementWorktypeStatusRead retrieves by name the id in question
+func dataSourceTaskManagementWorktypeStatusRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := newTaskManagementWorktypeStatusProxy(sdkConfig)
```
charliecon, line 24: Should be calling the getProxy function here

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_test.go
```diff
@@ -0,0 +1,174 @@
...
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+		CheckDestroy: testVerifyTaskManagementWorktypeStatusDestroyed,
+	})
+}
+
+func testVerifyTaskManagementWorktypeStatusDestroyed(state *terraform.State) error {
+	taskManagementApi := platformclientv2.NewTaskManagementApi()
```
charliecon, line 150: Why don't we need to provide a Configuration object to authorise the call? I'm not requesting a change, just curious how this works

carnellj-genesys (approved):

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_test.go
dginty4: We don't really tend to pass in a configuration when creating apis in the verify destroyed function. I was looking at the code there and when you don't pass in a configuration it will create a default config. I think because it's a test function we don't want it using a configuration from the configuration pool in the provider

PR #1167, opened 2024-07-18
Devtooling 625: Adding the id fields to question_groups, questions, answer_options as READ-ONLY

charliecon:


genesyscloud/resource_genesyscloud_quality_forms_evaluation.go
```diff
@@ -16,17 +16,25 @@ import (
...
 
+const readOnlyWarningMessage = "READ-ONLY. No value should be provided for this field as one will be generated by the API."
+
 var (
 	evaluationFormQuestionGroup = &schema.Resource{
 		Schema: map[string]*schema.Schema{
+			"id": {
+				Description: fmt.Sprintf("ID of the question group. %s", readOnlyWarningMessage),
+				Type:        schema.TypeString,
+				Optional:    true,
```
bbbco, line 35: I believe you only need to set `Computed: true` and don't need to set `Optional: true` if the user is not intended to set/update this value. See the docs for `Optional` and `Computed` behaviors here: https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors

charliecon: I was trying to find a way to mark an attribute as read-only! Didn't find that article. Thanks, Brian!

bbbco: Woohoo! Glad to help!

HemanthDogiparthi12 (approved):

HemanthDogiparthi12 (commented):
Think the export tests are failing in question_groups. Can you correct them.

carnellj-genesys (approved):

PR #1170, opened 2024-07-19
Tf state for DataSources

HemanthDogiparthi12:


charliecon (approved):
It would be good to have a test case that verifies the exported state file is correct i.e. that `"mode": "data"` and not `"mode": "managed"` for the objects specified in `replace_with_datasource`

dginty4 (approved):

PR #1176, opened 2024-07-19
Task/no jira remove test workflows

carnellj-genesys:
Removed the test workflows from github actions as all tests should not run in Jenkins


dginty4 (approved):

HemanthDogiparthi12 (approved):

charliecon:
I think we should consider leaving the unit tests in?

PR #1178, opened 2024-07-22
Inspired by DEVTOOLING-741 - Hardening E.164 handling

bbbco:
Hardening E.164 handling that may occur by ensuring all instances that we validate for an E.164 formatted number from user config (which are applied on create/update) are also returned (via read) as an E.164 formatted number to maintain consistency in number handling. This likely means we could remove the custom validate export handling for E164 that was inconsistently applied and used custom logic.

### DIFF
--- a/genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
+++ b/genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
@@ -84,7 +84,8 @@ func readIvrConfig(ctx context.Context, d *schema.ResourceData, meta interface{}
 		}
 
 		_ = d.Set("name", *ivrConfig.Name)
-		_ = d.Set("dnis", lists.StringListToSetOrNil(ivrConfig.Dnis))
+		dnis := lists.Map(*ivrConfig.Dnis, util.FormatAsCalculatedE164Number)
+		_ = d.Set("dnis", lists.StringListToSetOrNil(&dnis))
 
 		resourcedata.SetNillableValue(d, "description", ivrConfig.Description)
 		resourcedata.SetNillableReference(d, "open_hours_flow_id", ivrConfig.OpenHoursFlow)
--- a/genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
+++ b/genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
@@ -84,8 +84,12 @@ func readIvrConfig(ctx context.Context, d *schema.ResourceData, meta interface{}
 		}
 
 		_ = d.Set("name", *ivrConfig.Name)
-		dnis := lists.Map(*ivrConfig.Dnis, util.FormatAsCalculatedE164Number)
-		_ = d.Set("dnis", lists.StringListToSetOrNil(&dnis))
+		if ivrConfig.Dnis == nil || *ivrConfig.Dnis == nil {
+			_ = d.Set("dnis", nil)
+		} else {
+			dnis := lists.Map(*ivrConfig.Dnis, util.FormatAsCalculatedE164Number)
+			_ = d.Set("dnis", lists.StringListToSetOrNil(&dnis))
+		}
 
 		resourcedata.SetNillableValue(d, "description", ivrConfig.Description)
 		resourcedata.SetNillableReference(d, "open_hours_flow_id", ivrConfig.OpenHoursFlow)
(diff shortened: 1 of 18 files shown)


genesyscloud/architect_ivr/resource_genesyscloud_architect_ivr.go
```diff
@@ -84,7 +84,8 @@ func readIvrConfig(ctx context.Context, d *schema.ResourceData, meta interface{}
 		}
 
 		_ = d.Set("name", *ivrConfig.Name)
-		_ = d.Set("dnis", lists.StringListToSetOrNil(ivrConfig.Dnis))
+		dnis := lists.Map(*ivrConfig.Dnis, util.FormatAsCalculatedE164Number)
```
charliecon, line 87: Is there any risk of seeing a nil pointer dereference error here? Maybe we should check if `ivrConfig.Dnis` is nil first

genesyscloud/external_contacts/resource_genesyscloud_externalcontacts_contact_utils.go
```diff
@@ -89,7 +90,8 @@ func flattenPhoneNumber(phonenumber *platformclientv2.Phonenumber) []interface{}
 	resourcedata.SetMapValueIfNotNil(phonenumberInterface, "display", phonenumber.Display)
 	resourcedata.SetMapValueIfNotNil(phonenumberInterface, "extension", phonenumber.Extension)
 	resourcedata.SetMapValueIfNotNil(phonenumberInterface, "accepts_sms", phonenumber.AcceptsSMS)
-	resourcedata.SetMapValueIfNotNil(phonenumberInterface, "e164", phonenumber.E164)
+	phoneNumber := util.FormatAsCalculatedE164Number(*phonenumber.E164)
```
charliecon, line 93: Same here. The function `SetMapValueIfNotNil` would have been checking if `phonenumber.E164` was nil before dereferencing

genesyscloud/telephony_providers_edges_did/genesyscloud_telephony_providers_edges_did_schema.go
```diff
@@ -17,7 +18,7 @@ func SetRegistrar(l registrar.Registrar) {
 // DataSourceDid registers the genesyscloud_telephony_providers_edges_did data source
 func DataSourceDid() *schema.Resource {
 	return &schema.Resource{
-		Description: "Data source for Genesys Cloud DID. The identifier is the E-164 phone number.",
+		Description: "Data source for Genesys Cloud DID. The identifier is the E.164 phone number.",
```
charliecon, line 21: We should run `make docs` to render these changes

charliecon (approved):
Looks good! Just a few small comments

genesyscloud/util/util_e164.go
```diff
@@ -5,11 +5,21 @@ import (
...
 		return "", diag.Errorf("Failed to format phone number %s: %s", number, err)
 	}
 	formattedNum := phonenumbers.Format(phoneNumber, phonenumbers.E164)
 	return formattedNum, nil
 }
+
+// Formats string as a valid E.164 international standard telephone format.
+// Use this function when data is being returned back from the API already in expected format
+func FormatAsCalculatedE164Number(number string) string {
+	phoneNumber, _ := phonenumbers.Parse(number, "US")
```
HemanthDogiparthi12, line 22: can we have the region retrieved from env variable or aws region with default as "US" ?

HemanthDogiparthi12 (approved):

genesyscloud/util/util_e164.go
bbbco: Hey @HemanthDogiparthi12 , I implemented this by looking up the Organization's Default Country Code and passing that in to the phone numbers library. See commit https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1178/commits/1ba502536e117a81df0c9627932404c55c601d67
I have tested this commit out against the Japan org and it handles it correctly. I also had to use the dependency injection approach so that I could properly unit test the functions against both "US" and "JP" handling.

I do realize that using a global variable in the providers package is not exactly kosher, but I needed some way to store the Default Country Code. If you have any suggestions of alternative approaches, I'm all ears! Also, if we need to revert this commit in order to get the build out, I'm open to that and revisiting this later if need be.

cc @charliecon @carnellj

PR #1186, opened 2024-07-30
NO-JIRA: Fixing workitem test

dginty4:


genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_schema.go
```diff
@@ -92,7 +92,7 @@ func ResourceTaskManagementWorktypeStatus() *schema.Resource {
 				Type:        schema.TypeString,
 			},
 			`default`: {
-				Description: `This status is the default status for Workitems created from this Worktype. Only one status can be set as the default status at a time.`,
+				Description: `This status is the default status for Workitems created from this Worktype. Only one status can be set as the default status at a time. Once set there must always be a default status. The default can not be deleted.`,
```
charliecon, line 95: Can you update this field back to `false` ? If so, we should probably update it to false in the delete method

charliecon (approved):

carnellj-genesys (approved):

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_schema.go
dginty4: This property isn't actually on the `worktype_status` object. On the `worktype` object there's a field `default_status_id` but because of circular dependencies between `workype` and `worktype_status` I can't put `default_status_id` in the  `worktype` resource. As a workaround, I put this boolean `default` in the `worktype_status` resource. If true it will set `default_status_id` on the parent worktype. Setting the field to false doesn't trigger any code though since we can't unset `default_status_id` on the parent worktype

PR #1202, opened 2024-08-07
Feat/devtooling 654 Integration Instagram

shrutisuryawanshigenesys:


genesyscloud/conversations_messaging_integrations_instagram/genesyscloud_conversations_messaging_integrations_instagram_proxy.go
```diff
@@ -0,0 +1,168 @@
...
+			allInstagramIntegrationRequests = append(allInstagramIntegrationRequests, instagramIntegrationRequest)
+		}
+	}
+
+	return &allInstagramIntegrationRequests, resp, nil
+}
+
+// getConversationsMessagingIntegrationsInstagramIdByNameFn is an implementation of the function to get a Genesys Cloud conversations messaging integrations instagram by name
+func getConversationsMessagingIntegrationsInstagramIdByNameFn(ctx context.Context, p *conversationsMessagingIntegrationsInstagramProxy, name string) (id string, retryable bool, response *platformclientv2.APIResponse, err error) {
+	instagramIntegrationRequests, resp, err := getAllConversationsMessagingIntegrationsInstagramFn(ctx, p)
```
charliecon, line 136: I would use `p.getAllConversationsMessagingIntegrationsInstagram(ctx)` instead of the `Fn` function. It makes no difference now, but if we want to implement unit testing later it will make it easier to mock out this call.

genesyscloud/conversations_messaging_integrations_instagram/resource_genesyscloud_conversations_messaging_integrations_instagram.go
```diff
@@ -0,0 +1,150 @@
...
+
+	log.Printf("Reading conversations messaging integrations instagram %s", d.Id())
+
+	return util.WithRetriesForRead(ctx, d, func() *retry.RetryError {
+		instagramIntegrationRequest, resp, getErr := proxy.getConversationsMessagingIntegrationsInstagramById(ctx, d.Id())
+		if getErr != nil {
+			if util.IsStatus404(resp) {
+				return retry.RetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("Failed to read conversations messaging integrations instagram %s: %s", d.Id(), getErr), resp))
+			}
+			return retry.RetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("Failed to read conversations messaging integrations instagram %s: %s", d.Id(), getErr), resp))
```
charliecon, line 76: Retryable -> NonRetryable

genesyscloud/conversations_messaging_integrations_instagram/resource_genesyscloud_conversations_messaging_integrations_instagram_test.go
```diff
@@ -0,0 +1,171 @@
...
+func testVerifyConversationsMessagingIntegrationsInstagramDestroyed(state *terraform.State) error {
+	integrationApi := platformclientv2.NewConversationsApi()
+	for _, rs := range state.RootModule().Resources {
+		if rs.Type != "genesyscloud_conversations_messaging_integrations_instagram" {
+			continue
+		}
+
+		instagram, resp, err := integrationApi.GetConversationsMessagingIntegrationsInstagramIntegrationId(rs.Primary.ID, "")
+		if instagram != nil {
+			fmt.Errorf("Instagram still exists")
```
charliecon, line 162: `fmt.Errorf` -> `return fmt.Errorf`

genesyscloud/conversations_messaging_integrations_instagram/resource_genesyscloud_conversations_messaging_integrations_instagram.go
```diff
@@ -0,0 +1,150 @@
...
+	pageAccessToken := d.Get("page_access_token").(string)
+	userAccessToken := d.Get("user_access_token").(string)
+
+	conversationsMessagingIntegrationsInstagram := platformclientv2.Instagramintegrationupdaterequest{
+		Name:             platformclientv2.String(d.Get("name").(string)),
+		SupportedContent: &platformclientv2.Supportedcontentreference{Id: &supportedContentId},
+		MessagingSetting: &platformclientv2.Messagingsettingrequestreference{Id: &messagingContentId},
+		PageAccessToken:  &pageAccessToken,
+		UserAccessToken:  &userAccessToken,
+	}
```
charliecon, line 115: I would extend the function `getConversationsMessagingIntegrationsInstagramFromResourceData` to take a bool which indicates if it is for a create or update, and then return `Instagramintegrationupdaterequest` or `Instagramintegrationrequest` depending on that. At the moment, it's a bit inconsistent looking using it in the create but not for the update

genesyscloud/conversations_messaging_integrations_instagram/resource_genesyscloud_conversations_messaging_integrations_instagram_schema.go
```diff
@@ -0,0 +1,120 @@
...
+				Optional:    true,
+				Type:        schema.TypeString,
+				ForceNew:    true,
+			},
+			`app_secret`: {
+				Description: `The app Secret of Facebook app. The appSecret is required when appId is provided.`,
+				Optional:    true,
+				Type:        schema.TypeString,
+				ForceNew:    true,
+			},
```
charliecon, line 84: Are you sure all these fields need to be ForceNew? Did you see an error when you tried to update them ?

docs/resources/conversations_messaging_integrations_instagram.md
```diff
@@ -0,0 +1,54 @@
...
+## Example Usage
+
+```terraform
+resource "genesyscloud_conversations_messaging_integrations_instagram" "test_sample" {
+  name                 = "Test Integration Instagram"
+  user_access_token    = "1234567890"
+  page_id              = "1"
+  messaging_setting_id = "2c4e3b8e-3c9f-45c9-82cd-4bb54c8f18f0"
+  supported_content_id = "019c37a7-ccb4-4966-b1d7-ddb20399f7ab"
+}
```
charliecon, line 31: Could you change these GUIDs for resource references? e.g. `genesyscloud_conversations_messaging_settings.settings.id` and `genesyscloud_conversations_messaging_supportedcontent.content.id`

charliecon (approved):

genesyscloud/conversations_messaging_integrations_instagram/resource_genesyscloud_conversations_messaging_integrations_instagram_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+/*
+The resource_genesyscloud_conversations_messaging_integrations_instagram_test.go contains all of the test cases for running the resource
+tests for conversations_messaging_integrations_instagram.
+*/
+
+func TestAccResourceConversationsMessagingIntegrationsInstagram(t *testing.T) {
+	t.Skip("Skipping because it requires setting up a org as test account for the mocks to respond correctly.")
```
charliecon, line 25: I was able to run this test against our org inintesting. The instagram integration created but failed to update because it said that it wasn't finished creating. Is this what you're seeing? Seems like we could work around this test failure

carnellj-genesys: Any feedback on this @shrutisuryawanshigenesys .  Is this addressed?

carnellj-genesys (approved):

charliecon:
If I run

```no-highlight
go test ./genesyscloud/conversations_messaging_integrations_instagram/... -v
```

some tests fail and I think it is because we need to wait for a few seconds after creating the integration, before trying to delete it. You can find a few test cases that use the field `PreConfig` - you can use that to sleep before apply the config. I would add an initial TestStep to create the integration, then in the next use the PreConfig field to sleep for a few seconds before applying again and verifying fields etc. Then the delete should work at the end 

In the data source test, I think you need to create the resource and then let the sleep happen before creating the data source

charliecon (approved):

PR #1208, opened 2024-08-08
Bug/devtooling 642

HemanthDogiparthi12:


charliecon (approved):

bbbco:
@HemanthDogiparthi12 , as we discussed about on GC, I'm concerned that the extra code to handle DEVTOOLING-319 inside of `getAllAuthoOutboundRulesets()` doesn't really belong there. Instead, the `resourceExporter.RuleSetSkillPropertyResolver()` should be doing the heavy lifting.

If we look at things, the `doesRule{Actions,Conditions}RefDeletedSkill()` function is getting the actual values of resource objects when it is called by `filterOutboundRulesets()` inside of `getAllAuthoOutboundRulesets()` _BEFORE_ any other resource objects are actually retrieved. This is not adhering to the standard export strategy of the provider, where we use `getAllAuthOutboundRulesets()` to get the ids of all objects for that resource type. We retrieve all the objects individually using the `read` function. We then use the CustomAttributeResolver AFTER the full resource objects are fully retrieved and added to the config map. This is the proper place to be doing the filtering to handle ids that don't match, and we'll already have the skills resources retrieved, so there is no reason to go grab them in a separate call.

carnellj-genesys (approved):

dginty4:
@HemanthDogiparthi12 Is this done?

HemanthDogiparthi12:
this PR was meant to be discarded. But now there is an edge case identified with inindca.atlassian.net/browse/DEVTOOLING-817 , this need to go now to fix that problem.

carnellj-genesys (approved):

PR #1212, opened 2024-08-13
Fixes DEVTOOLING-786 and DEVTOOLING-740 as well as other resources that could be impacted

bbbco:
This PR fixes DEVTOOLING-786 and DEVTOOLING-740 by improving permission handling for single-instance resources

Background:
- Many single-instance resources (e.g., `genesyscloud_outbound_settings`) return a simple `ResourceMeta` object in their `getAll...()` function.
- These resources have API access permissions associated with them.
- The exporter checks for permission errors after the initial `getAll...()` call and handles them via logging if `log_permissions_errors` is configured.
- Constructing a simple `ResourceMeta` object bypasses this check, causing errors in the downstream `read` functions even when `log_permissions_errors` is set.

Fix:
- Update the `getAll...()` function for single-instance resources that return key-value objects.
- Implement an API call to confirm user permissions before returning data.
- This approach aligns with the existing convention used in `genesyscloud_idp_*` resources.

This change ensures proper permission checking and error handling for single-instance resources, resolving the issue with the `log_permissions_errors` configuration.


charliecon (approved):
Nice catch 👌

carnellj-genesys (approved):

PR #1218, opened 2024-08-19
Ruleset and Skill Resource

HemanthDogiparthi12:
This PR

Will remove the filter of rulesets based on skill set existence.
Instead it will filter the skills being referenced i they are not existent. 
The rulesets will still be present.

### DIFF
--- a/genesyscloud/provider/provider.go
+++ b/genesyscloud/provider/provider.go
@@ -135,6 +135,7 @@ func New(version string, providerResources map[string]*schema.Resource, provider
 								DefaultFunc: schema.EnvDefaultFunc("GENESYSCLOUD_PROXY_PROTOCOL", nil),
 								Description: "Protocol for the proxy can be set with the `GENESYSCLOUD_PROXY_PROTOCOL` environment variable.",
 							},
+
 							"auth": {
 								Type:     schema.TypeSet,
 								Optional: true,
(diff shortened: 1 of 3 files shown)


carnellj-genesys (approved):

genesyscloud/provider/provider.go
```diff
@@ -178,24 +179,9 @@ type ProviderMeta struct {
 
 func configure(version string) schema.ConfigureContextFunc {
 	return func(context context.Context, data *schema.ResourceData) (interface{}, diag.Diagnostics) {
-		// Initialize a single client if we have an access token
```
bbbco, line 181: What is this?

bbbco:
I'm not entirely convinced the issue is completely resolved by removing this code. The code that Will had added here is checking for and comparing the `skills` ids that return in this resource to existing `genesyscloud_routing_skill` resource types and removing/filtering any non-matching skills. 

I think the full fix to this is to use a `CustomAttributeResolver` hook in the exporter definition to do the comparisons between the `skill` references found in the outbound ruleset, and the actual skills found in the `genesyscloud_routing_skill` resource type and filter out / log any references that cannot be resolved

HemanthDogiparthi12:
> I'm not entirely convinced the issue is completely resolved by removing this code. I mean, it will fix the hanging issue, but the code that Will had added here for DEVTOOLING-319 is checking for and comparing the `skills` ids that return in this resource to existing `genesyscloud_routing_skill` resource types and removing/filtering any non-matching skills. We'll just get another report of DEVTOOLING-319 without handling this issue properly.
> 
> I think the full fix to this is to use a `CustomAttributeResolver` hook in the exporter definition to do the comparisons between the `skill` references found in the outbound ruleset, and the actual skills found in the `genesyscloud_routing_skill` resource type and filter out / log any references that cannot be resolved


Hi Brian..

The problem is CustomAttributeResolver does only have references for the attributes and  to filter them or modify them.

if we want to change the instances of exporter based on some condition , we need to overridde the GetResourceFunc of Exporter

In other words modify the GetResourceFunc method to give us the values we want after filter

genesyscloud/provider/provider.go
HemanthDogiparthi12: since we have a single client for access token, we are seeing issues when getResourceFunc context is called more than once anywhere in the already existing context.  This is forming a race condition where both the getResourceContexts are waiting for a config pool and we only have one, therefore making the export hung for infinite time.

this issue happens when a access token is provided instead of credential. and the exporter gets hung at the filter logic of ruleset ( using exporter instance of skillset to get to getllroutingskills)

root cause for both the issues
https://inindca.atlassian.net/browse/DEVTOOLING-750
https://inindca.atlassian.net/browse/DEVTOOLING-642

genesyscloud/resource_exporter/resource_exporter_custom.go
```diff
@@ -162,9 +162,6 @@ func RuleSetSkillPropertyResolver(configMap map[string]interface{}, exporters ma
 				if exists {
 					exportId = value.Name
 					sanitisedSkillIds = append(sanitisedSkillIds, fmt.Sprintf("${genesyscloud_routing_skill.%s.id}", exportId))
-				} else {
```
HemanthDogiparthi12, line 165: removing this , will make sure the import to new org will also work. the deleted skills are simply ignored in the skillset references.

genesyscloud/provider/provider.go
bbbco: Great find!

PR #1239, opened 2024-08-28
DEVTOOLING-814 - Fixed routing email route test

BrianMoyles:


dginty4 (approved):

genesyscloud/routing_email_route/resource_genesyscloud_routing_email_route_test.go
```diff
@@ -24,142 +22,32 @@ import (
...
+		fromName2     = "Jane Terraform"
+		priority1     = "1"
+		bccEmail1     = "test1@" + domainId
+		bccEmail2     = "test2@" + domainId
 	)
 
 	CleanupRoutingEmailDomains()
 
-	// Test error configs
-	resource.Test(t, resource.TestCase{
```
HemanthDogiparthi12, line 56: Hi, Didnt understand why this is removed.

BrianMoyles: reply_email_address and from_email no longer conflict with each other. The removed test cases expected a error when both are set, which now caused the test to fail.

carnellj-genesys (approved):

PR #1244, opened 2024-09-05
No jira devcontainers

carnellj-genesys:


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM mcr.microsoft.com/vscode/devcontainers/base:ubuntu
```
carnellj-genesys, line 1: Installingthe genesys cloud cli and archy.  Archy still needs a little work, but I am working on ot

GNUmakefile
```diff
@@ -54,6 +54,7 @@ clean:
 
 build:
 	mkdir -p ${DIST_DIR}
+	go mod tidy
```
carnellj-genesys, line 57: Added this to build part because if our go mod gets out of date I want to make sure go mod tidy runs

scripts/generate_env.py
```diff
@@ -0,0 +1,39 @@
+## This script is included in our .devcontainers setup.  We basically use this to generate our an environment variables file that can be used to set the Genesys Cloud environment variables.
```
carnellj-genesys, line 1: Simple python script I am including in the containers to help set environment variables

kavinbalagen (approved):

PR #1251, opened 2024-09-10
DEVTOOLING-737

sureshperiyappan:


charliecon (approved):

genesyscloud/resource_genesyscloud_knowledge_category_test.go
```diff
@@ -88,20 +173,36 @@ func generateKnowledgeCategoryResource(resourceName string, knowledgeBaseResourc
...
 }
 
-func generateKnowledgeCategoryRequestBody(categoryName string, categoryDescription string) string {
+func generateKnowledgeCategoryRequestBody(categoryName string, categoryDescription string, parrentCategoryName string) string {
 
 	return fmt.Sprintf(`
         knowledge_category {
             name = "%s"
             description = "%s"
+			%s
```
charliecon, line 201: I would leave this as 

```
parent_id = %s
```

and pass `genesyscloud_knowledge_category."+parentCategoryName+".id"` into the function. And if you do not want to provide a value, you can pass `util.NullValue` as a parameter. This is a constant value for "null" , and it is the same as omitting the field from the resource altogether.

If you want the resource to look like this

```
parent_id = "some hard-coded string"
```

You would pass `strconv.Quote("some hard-coded string")` into the function to add the quotation marks

genesyscloud/resource_genesyscloud_knowledge_category_test.go
```diff
@@ -88,20 +173,36 @@ func generateKnowledgeCategoryResource(resourceName string, knowledgeBaseResourc
...
+	`, resourceName,
+		knowledgeBaseResource,
+		generateKnowledgeCategoryRequestBody(categoryName, categoryDescription,
+			"parent_id = genesyscloud_knowledge_category."+parentCategoryName+".id"),
 	)
 	return category
 }
 
-func generateKnowledgeCategoryRequestBody(categoryName string, categoryDescription string) string {
+func generateKnowledgeCategoryRequestBody(categoryName string, categoryDescription string, parrentCategoryName string) string {
```
charliecon, line 195: `parrentCategoryName` -> `parentCategoryName`

charliecon (approved):

genesyscloud/resource_genesyscloud_knowledge_category_test.go
sureshperiyappan: thanks for the suggestions

charliecon (approved):

PR #1270, opened 2024-09-26
No jira: auth data source minor refactor

charliecon:


genesyscloud/auth_division/genesyscloud_auth_division_proxy.go
```diff
@@ -131,13 +131,15 @@ func getAuthDivisionByIdFn(ctx context.Context, p *authDivisionProxy, id string,
...
+	notFoundError := fmt.Errorf("unable to find auth division with name %s", name)
+
 	authzDivisions, resp, err := getAllAuthDivisionFn(ctx, p, name)
 	if err != nil {
 		return "", resp, false, err
 	}
 
 	if authzDivisions == nil || len(*authzDivisions) == 0 {
-		return "", resp, true, err
+		return "", resp, true, notFoundError
```
charliecon, line 142: Because we were returning nil as the error here, there was no error info displayed when no division was found with that name. It would just say `Error requesting division {{name}} | error: <nil>`

genesyscloud/auth_division/data_source_genesyscloud_auth_division.go
```diff
@@ -21,11 +21,12 @@ func dataSourceAuthDivisionRead(ctx context.Context, d *schema.ResourceData, m i
 	// Query division by name. Retry in case search has not yet indexed the division.
 	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
 		divisionId, resp, retryable, getErr := proxy.getAuthDivisionIdByName(ctx, name)
-		if getErr != nil && !retryable {
-			return retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("Error requesting division %s | error: %s", name, getErr), resp))
-		}
-		if retryable {
```
HemanthDogiparthi12, line 27: think this could be the problem. retryable infinite loop here. no getErr check ?

carnellj-genesys (approved):

PR #1271, opened 2024-09-26
No jira/fixauth

HemanthDogiparthi12:


genesyscloud/auth_division/genesyscloud_auth_division_proxy.go
```diff
@@ -137,7 +137,7 @@ func getAuthDivisionIdByNameFn(ctx context.Context, p *authDivisionProxy, name s
 	}
 
 	if authzDivisions == nil || len(*authzDivisions) == 0 {
-		return "", resp, true, err
+		return "", resp, false, err
 	}
```
charliecon, line 141: I think we should still retry here because we could be waiting for a division that was just created to be returned from [GET /api/v2/authorization/divisions](https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-authorization-divisions)

If it is retrying forever then that would be a problem with `util.WithRetries` because it should stop after 15 seconds

PR #1284, opened 2024-10-08
DEVTOOLING-872: Added some nil pointer checks around user groups

carnellj-genesys:


genesyscloud/group_roles/genesyscloud_group_roles_proxy.go
```diff
@@ -69,6 +70,11 @@ func getGroupRolesByIdFn(_ context.Context, p *groupRolesProxy, roleId string) (
 func updateGroupRolesFn(_ context.Context, p *groupRolesProxy, roleId string, rolesConfig *schema.Set, subjectType string) (*platformclientv2.APIResponse, error) {
 	// Get existing roles/divisions
 	subject, resp, err := p.authorizationApi.GetAuthorizationSubject(roleId, true)
+
```
carnellj-genesys, line 73: There was not a nil check here.  So if the user entered a role name that was not real, the subject would not be retrieved and the code would fail with a nil pointer

genesyscloud/user_roles/genesyscloud_user_roles_proxy.go
```diff
@@ -70,6 +72,11 @@ func getUserRolesByIdFn(_ context.Context, p *userRolesProxy, roleId string) (*[
 func updateUserRolesFn(_ context.Context, p *userRolesProxy, roleId string, rolesConfig *schema.Set, subjectType string) (*platformclientv2.APIResponse, error) {
 	// Get existing roles/divisions
 	subject, resp, err := p.authorizationApi.GetAuthorizationSubject(roleId, true)
+
+	if err != nil || resp.StatusCode == http.StatusNotFound || subject == nil {
```
carnellj-genesys, line 76: There was not a nil check here. So if the user entered a role name that was not real, the subject would not be retrieved and the code would fail with a nil pointer

dginty4 (approved):

PR #1291, opened 2024-10-11
Fix/devtooling 840

bbbco:
The hope is to be able to explicitly set contact ids will keep the state clean.


carnellj-genesys (approved):

genesyscloud/genesyscloud.tf
dginty4, line 1: What is this file? It seems extremely large

bbbco:
Yeah that was inadvertently committed. It was an export output during
debugging.

Brian Goad
@bbbco
Creative Analytic Convergence

On October 14, 2024, GitHub Notifications ***@***.***>
wrote:
> @dginty4 commented on this pull request.
>

> On genesyscloud/genesyscloud.tf
> <https://github.com/MyPureCloud/terraform-provider-
> genesyscloud/pull/1291#discussion_r1799461199>:
>
> What is this file? It seems extremely large
>

—
Reply to this email directly, view it on GitHub
<https://github.com/MyPureCloud/terraform-provider-
genesyscloud/pull/1291#pullrequestreview-2366561899>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAAHPWHCQVSZX73L4KFZO3DZ3O6CRAVCNFSM6AAAAABPZX2QN6VHI2DSMVQWIX3LMV43YUDVNRWFEZLROVSXG5CSMV3GSZLXHMZDGNRWGU3DCOBZHE>.
You are receiving this because you authored the thread.Message ID:
<MyPureCloud/terraform-provider-
***@***.***>

genesyscloud/outbound_contact_list_contact/resource_genesyscloud_outbound_contact_list_contact_utils.go
```diff
@@ -12,6 +12,7 @@ import (
 
 // buildWritableContactFromResourceData used to build the request body for contact creation
 func buildWritableContactFromResourceData(d *schema.ResourceData) platformclientv2.Writabledialercontact {
+	contactId := d.Get("contact_id").(string)
```
charliecon, line 15: if the user provides `null` for contact_id or completely omits it from the resource, I'm not sure that it will be an empty string by default. Either way, it definitely will be if you change this to 

```golang
contactId, _ := d.Get("contact_id").(string)
```
so it might be worth making that change just in case.

charliecon (approved):
Looks good. Just had one small comment. Also you'll need to remove the accidentally checked-in file

genesyscloud/outbound_contact_list_contact/resource_genesyscloud_outbound_contact_list_contact_utils.go
carnellj-genesys: @bbbco Can you make this small change.  Then we can merge

bbbco: Fixed

PR #1293, opened 2024-10-14
NO-JIRA: Modified the core HTTP provider to capture data in a JSON fo…

carnellj-genesys:
…rmat to make it easier to parse and do analysis on

### DIFF
--- a/genesyscloud/provider/provider.go
+++ b/genesyscloud/provider/provider.go
@@ -268,14 +268,23 @@ func InitClientConfig(data *schema.ResourceData, version string, config *platfor
 		RetryWaitMax: time.Second * 30,
 		RetryMax:     20,
 		RequestLogHook: func(request *http.Request, count int) {
-			if count > 0 && request != nil {
-				log.Printf("Retry #%d for %s %s", count, request.Method, request.URL)
+			sdkDebugRequest := newSDKDebugRequest(request, count)
+			request.Header.Set("TF-Correlation-Id", sdkDebugRequest.TransactionId)
+			err, jsonStr := sdkDebugRequest.ToJSON()
+
+			if err != nil {
+				log.Printf("WARNING: Unable to log RequestLogHook: %s", err)
 			}
+			log.Printf(jsonStr)
 		},
 		ResponseLogHook: func(response *http.Response) {
-			if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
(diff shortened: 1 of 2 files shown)


genesyscloud/provider/provider.go
```diff
@@ -268,14 +268,23 @@ func InitClientConfig(data *schema.ResourceData, version string, config *platfor
 		RetryWaitMax: time.Second * 30,
 		RetryMax:     20,
 		RequestLogHook: func(request *http.Request, count int) {
-			if count > 0 && request != nil {
-				log.Printf("Retry #%d for %s %s", count, request.Method, request.URL)
+			sdkDebugRequest := newSDKDebugRequest(request, count)
```
carnellj-genesys, line 271: So I restructured the logging in the core provider so it logs before and after each HTTP request.  After the AirBNB work, it did not make sense to just log information only when somethings fails.  These log messages only show up when the TF_LOG environment variable is set.

genesyscloud/provider/provider.go
```diff
@@ -268,14 +268,23 @@ func InitClientConfig(data *schema.ResourceData, version string, config *platfor
...
+			request.Header.Set("TF-Correlation-Id", sdkDebugRequest.TransactionId)
```
carnellj-genesys, line 272: I am setting this variable on the request so we can capture it on the response (even though the public API does not give it back).  By having a stable ID, I can then merge the request and response in my analysis on a single call.

genesyscloud/provider/provider_sdk_debug.go
```diff
@@ -0,0 +1,74 @@
...
+	TransactionId    string `json:"transaction_id,omitempty"`    //Unique id to link the request and response
+	InvocationCount  int    `json:"invocation_count,omitempty"`  //Number of times the URL has been invoked.  Will be greater then zero when it is a retry
+	InvocationMethod string `json:"invocation_method,omitempty"` //HTTP method that will be invoked
+	InvocationUrl    string `json:"invocation_url,omitempty"`    //HTTP URL
+}
+
+func (s *sdkDebugRequest) ToJSON() (err error, jsonStr string) {
+	jsonData, err := json.Marshal(s)
+	if err != nil {
+		fmt.Println(err)
```
charliecon, line 21: Should we be using `fmt.Println` here? `log` would probably be more appropriate

charliecon (approved):

genesyscloud/provider/provider_sdk_debug.go
carnellj-genesys: fixed

PR #1297, opened 2024-10-16
Devtooling 867

HemanthDogiparthi12:


genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -429,6 +422,65 @@ func (g *GenesysCloudResourceExporter) buildResourceConfigMap() diag.Diagnostics
...
+
+	if len(exporters[resource.Type].CustomFlowResolver) > 0 {
+		g.updateInstanceStateAttributes(jsonResult, resource)
+	}
+}
+
+func (g *GenesysCloudResourceExporter) updateInstanceStateAttributes(jsonResult util.JsonMap,
+	resource resourceExporter.ResourceInfo) {
+
+	for attr, val := range jsonResult {
```
carnellj-genesys, line 444: This seems like a really big convuluted if..for..if..if..for..if loop. We might want to reconsider refactoring this or at the least put a ticket out there to refactor this.  :)

carnellj-genesys (approved):

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
HemanthDogiparthi12: refactored this code logic a bit.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -429,6 +422,74 @@ func (g *GenesysCloudResourceExporter) buildResourceConfigMap() diag.Diagnostics
...
+			// Directly add string attribute for rest of the flows/scripts/assets
+			if _, exists := resource.State.Attributes[attr]; !exists {
+				resource.State.Attributes[attr] = valStr
+			}
+		}
+	}
+}
+
+// Find the resourceID, return early if found
+func findResourceID(resource resourceExporter.ResourceInfo, resMap map[string]interface{}) string {
```
bbbco, line 471: What is the ResourceID? Why can't you just do `resource.State.ID`?

charliecon (commented):
I think what Adam is asking for is that the hash of the file be included in the state file on export like below 

```json
"name": "a_test_prompt_2210",
 "resources": [
    {
      "file_content_hash": "431af9cf1a02ccd6edfe0b1db5225ac67b7a49b21114c6eee0727a660a2b3f45",
      "filename": "./test.wav",
      "language": "en-us",
      "text": "This is a test greeting!",
       "tts_string": ""
      }
   ]
```

the way this is built now, the filesha256 function will be in the string for `file_content_hash` in the state file, and this will just be overwritten by the actual hash when the user tries to apply the export

charliecon (approved):

HemanthDogiparthi12:
> I think what Adam is asking for is that the hash of the file be included in the state file on export like below
> 
> ```json
> "name": "a_test_prompt_2210",
>  "resources": [
>     {
>       "file_content_hash": "431af9cf1a02ccd6edfe0b1db5225ac67b7a49b21114c6eee0727a660a2b3f45",
>       "filename": "./test.wav",
>       "language": "en-us",
>       "text": "This is a test greeting!",
>        "tts_string": ""
>       }
>    ]
> ```
> 
> the way this is built now, the filesha256 function will be in the string for `file_content_hash` in the state file, and this will just be overwritten by the actual hash when the user tries to apply the export

resolved

PR #1309, opened 2024-10-22
DEVTOOLING-882: Put a check to make sure we dont get a nil pointer ex…

carnellj-genesys:
…ception

### DIFF
--- a/genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
+++ b/genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
@@ -272,9 +272,12 @@ func (p *architectUserPromptProxy) verifyPromptResourceFilesAreTranscoded(ctx co
 			response = resp
 			retry.NonRetryableError(fmt.Errorf("failed to read user prompt '%s': %v", promptId, err))
 		}
-		if userPrompt.Resources == nil {
+
+		if userPrompt != nil || userPrompt.Resources == nil {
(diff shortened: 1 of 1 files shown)


genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
```diff
@@ -272,9 +272,12 @@ func (p *architectUserPromptProxy) verifyPromptResourceFilesAreTranscoded(ctx co
 			response = resp
 			retry.NonRetryableError(fmt.Errorf("failed to read user prompt '%s': %v", promptId, err))
 		}
-		if userPrompt.Resources == nil {
+
+		if userPrompt != nil || userPrompt.Resources == nil {
+			log.Printf("WARNING: User prompt or userPrompt.Resources is nil in the call from getArchitectUserPrompt().  StatusCode returned by the call %d", resp.StatusCode)
```
charliecon, line 277: I think this should be `if userPrompt == nil || userPrompt.Resources == nil`

charliecon (changes requested):

genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
carnellj-genesys: Good catch.  fixed

charliecon (approved):

PR #1314, opened 2024-10-23
NOJIRA Fix/script filepath arg

bbbco:
I noticed this while working with some exports. The attribute is actually `filepath` not `filename` in the schema.


carnellj-genesys (approved):

charliecon:
Nice catch! I'm going to close this PR though because Hemanth just fixed this in PR #1316 , along with another similar bug

PR #1315, opened 2024-10-23
NOJIRA fix oauth client.authorized_grant_type to accept SAML2-BEARER value

bbbco:
Update `oauth_client`'s `authorized_grant_type` attribute to accept both `SAML2BEARER` and `SAML2-BEARER` values (the API allows both versions, but outputs as SAML2-BEARER in the API and export. It's important that the export be able to be reapplied right away, so this fix handles that).

Otherwise, this can happen when trying to apply an export of the oauth_client resource:

```
│ Error: expected authorized_grant_type to be one of ["CODE" "TOKEN" "SAML2BEARER" "PASSWORD" "CLIENT-CREDENTIALS"], got SAML2-BEARER
│
│   with genesyscloud_oauth_client.TT_SAML,
│   on genesyscloud.tf line 2799, in resource "genesyscloud_oauth_client" "TT_SAML":
│ 2799:   authorized_grant_type         = "SAML2-BEARER"
│
╵
```

### DIFF
--- a/genesyscloud/oauth_client/resource_genesyscloud_oauth_client_schema.go
+++ b/genesyscloud/oauth_client/resource_genesyscloud_oauth_client_schema.go
@@ -1,11 +1,12 @@
 package oauth_client
 
 import (
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"terraform-provider-genesyscloud/genesyscloud/provider"
 	resourceExporter "terraform-provider-genesyscloud/genesyscloud/resource_exporter"
 	registrar "terraform-provider-genesyscloud/genesyscloud/resource_register"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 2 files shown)


genesyscloud/oauth_client/resource_genesyscloud_oauth_client_schema.go
```diff
@@ -73,10 +74,10 @@ func ResourceOAuthClient() *schema.Resource {
 				Optional:    true,
 			},
 			"authorized_grant_type": {
-				Description:  "The OAuth Grant/Client type supported by this client (CODE | TOKEN | SAML2BEARER | PASSWORD | CLIENT-CREDENTIALS).",
+				Description:  "The OAuth Grant/Client type supported by this client (CODE | TOKEN | SAML2-BEARER | PASSWORD | CLIENT-CREDENTIALS).",
```
bbbco, line 77: `SAML2-BEARER` is what is output via the API, which is why I chose to use this in the description examples instead of `SAML2BEARER`. In reality, both values are perfectly acceptable.

charliecon (approved):

PR #1324, opened 2024-11-01
Fixes DEVTOOLING-906

bbbco:
Fixes DEVTOOLING-906 by calling the `/api/v2/telephony/providers/edges/phonebasesettings/templates` endpoint first before create/update to pull the `lineMetaBase` details to pass into the POST/PUT body.

I added unit tests to ensure that the template call is being made

I had to move the `customizePhoneBaseSettingsPropertiesDiff`  into the same package because I needed to update the calls to use the proxy, otherwise my unit tests were failing. I went ahead and updated the other resources that had proxy pattern already established, but we're still waiting for the trunks resource to be proxyed.


genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings_unit_test.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+		return &phoneBaseSettings, &apiResponse, nil
+	}
+
+	pbsProxy.postPhoneBaseSettingAttr = func(ctx context.Context, p *phoneBaseProxy, body platformclientv2.Phonebase) (*platformclientv2.Phonebase, *platformclientv2.APIResponse, error) {
+		functionCalls = append(functionCalls, "postPhoneBaseSettingAttr")
+		assert.Equal(t, tSettingsName, *body.Name)
+		assert.Equal(t, tSettingsMetaBaseId, *body.PhoneMetaBase.Id)
+
+		// TODO fill out rest of assert commands
```
carnellj-genesys, line 91: I would remove this TODO or fill in the rest of the assert commands :)

carnellj-genesys (approved):

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings.go
```diff
@@ -47,7 +41,24 @@ func createPhoneBaseSettings(ctx context.Context, d *schema.ResourceData, meta i
...
+	}
+
+	phoneBaseSettingTemplateLines := *phoneBaseSettingTemplate.Lines
+	if len(phoneBaseSettingTemplateLines) == 0 {
+		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to get phone base settings template lines for %s", phoneMetaBase), resp)
+	}
+	phoneBase.Lines = &[]platformclientv2.Linebase{
+		{
+			Name:         &name,
+			LineMetaBase: phoneBaseSettingTemplateLines[0].LineMetaBase,
```
HemanthDogiparthi12, line 57: can we have more than 1 phoneBaseSettingTemplateLines ? and are we sure it is always the 1st one from the api that need to be assigned for linemetabase ?

bbbco: the template always only returns a single instance of the line meta base.

HemanthDogiparthi12 (approved):

PR #1351, opened 2024-11-12
[DEVTOOLING-849] API Limitation on Prompt Export

BrianMoyles:
This PR changes the functionality of the getAllPrompts due to hitting the 10,000 limit.
To prevent this, the API will be called with different name filters based on starting letter of the prompt

### DIFF
--- a/genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
+++ b/genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
@@ -22,6 +22,7 @@ var internalProxy *architectUserPromptProxy
 
 type createArchitectUserPromptFunc func(ctx context.Context, p *architectUserPromptProxy, body platformclientv2.Prompt) (*platformclientv2.Prompt, *platformclientv2.APIResponse, error)
 type getArchitectUserPromptFunc func(ctx context.Context, p *architectUserPromptProxy, id string, includeMediaUris bool, includeResources bool, language []string, checkCache bool) (*platformclientv2.Prompt, *platformclientv2.APIResponse, error)
+type getArchitectUserPromptPageCountFunc func(ctx context.Context, p *architectUserPromptProxy, name string) (int, *platformclientv2.APIResponse, error)
 type getAllArchitectUserPromptsFunc func(ctx context.Context, p *architectUserPromptProxy, includeMediaUris bool, includeResources bool, name string) (*[]platformclientv2.Prompt, *platformclientv2.APIResponse, error)
 type updateArchitectUserPromptFunc func(ctx context.Context, p *architectUserPromptProxy, id string, body platformclientv2.Prompt) (*platformclientv2.Prompt, *platformclientv2.APIResponse, error)
 type deleteArchitectUserPromptFunc func(ctx context.Context, p *architectUserPromptProxy, id string, allResources bool) (*platformclientv2.APIResponse, error)
@@ -38,6 +39,7 @@ type architectUserPromptProxy struct {
 	architectApi                                   *platformclientv2.ArchitectApi
 	createArchitectUserPromptAttr                  createArchitectUserPromptFunc
 	getArchitectUserPromptAttr                     getArchitectUserPromptFunc
+	getArchitectUserPromptPageCountAttr            getArchitectUserPromptPageCountFunc
 	getAllArchitectUserPromptsAttr                 getAllArchitectUserPromptsFunc
 	updateArchitectUserPromptAttr                  updateArchitectUserPromptFunc
 	deleteArchitectUserPromptAttr                  deleteArchitectUserPromptFunc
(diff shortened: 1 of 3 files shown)


genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
```diff
@@ -155,11 +156,20 @@ func deleteArchitectUserPromptFn(_ context.Context, p *architectUserPromptProxy,
...
 }
 
-func getAllArchitectUserPromptsFn(_ context.Context, p *architectUserPromptProxy, includeMediaUris, includeResources bool, name string) (*[]platformclientv2.Prompt, *platformclientv2.APIResponse, error) {
+func getAllArchitectUserPromptsFn(_ context.Context, p *architectUserPromptProxy, includeMediaUris, includeResources bool, name string, exportNames []string) (*[]platformclientv2.Prompt, *platformclientv2.APIResponse, error) {
 	const pageSize = 100
 	var allPrompts []platformclientv2.Prompt
+	var nameFilter []string
 
-	userPrompts, response, err := p.architectApi.GetArchitectPrompts(1, pageSize, []string{name}, "", "", "", "", includeMediaUris, includeResources, nil)
+	// Determine whether to pass in a single name for getByName or export name filter
```
bbbco, line 164: Why not leave it as is and just pass in to the `name` argument the first character of each letter in the `getAllUserPrompts()`?

genesyscloud/architect_user_prompt/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -25,14 +26,17 @@ func getAllUserPrompts(ctx context.Context, clientConfig *platformclientv2.Confi
...
 	proxy := getArchitectUserPromptProxy(clientConfig)
 
-	userPrompts, resp, err := proxy.getAllArchitectUserPrompts(ctx, true, true, "")
-	if err != nil {
-		return nil, util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("failed to get user prompts: %s", err), resp)
-	}
-	for _, userPrompt := range *userPrompts {
-		resources[*userPrompt.Id] = &resourceExporter.ResourceMeta{Name: *userPrompt.Name}
-	}
+	exportNameFilter := "abcdefghijklmnopqrstuvwxyz"
```
HemanthDogiparthi12, line 29: there could be names starting with special chars or nums ? also check if we can leverage "name" argument instead of a new argument "exportNames"?

HemanthDogiparthi12, line 29: there could be names starting with special chars or nums ? also check if we can leverage "name" argument instead of a new argument "exportNames"?

BrianMoyles: Added Check for Names starting with a number. 
Not Necessary to check for special chars as api specifies:
`Prompt names can only contain letters, numbers, and the underscore, and must start with a letter or number.`

genesyscloud/architect_user_prompt/resource_genesyscloud_architect_user_prompt.go
```diff
@@ -25,14 +26,17 @@ func getAllUserPrompts(ctx context.Context, clientConfig *platformclientv2.Confi
...
-	if err != nil {
-		return nil, util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("failed to get user prompts: %s", err), resp)
-	}
-	for _, userPrompt := range *userPrompts {
-		resources[*userPrompt.Id] = &resourceExporter.ResourceMeta{Name: *userPrompt.Name}
-	}
+	exportNameFilter := "abcdefghijklmnopqrstuvwxyz1234567890"
 
+	for _, filter := range strings.Split(exportNameFilter, "") {
+		userPrompts, resp, err := proxy.getAllArchitectUserPrompts(ctx, true, true, filter+"*")
```
charliecon, line 32: Is this case insensitive? I mean will "a*" match any prompt that starts with "A"?

charliecon (commented):
Nice work 👌 I think we should have a check to see if this logic needs to be applied. I think that would look like an initial call to check the pageCount, and if it is less than 100, collect the prompts the normal way. This workaround will only be needed by a small number of users, and adding this check will mean if there is a bug in it, it won't affect everyone

genesyscloud/architect_user_prompt/resource_genesyscloud_architect_user_prompt.go
bbbco: Also, what about names that are not alphanumeric?

carnellj-genesys: Can we have names that are not alphanumeric?  Does prompts allow it?

carnellj-genesys (approved):

genesyscloud/architect_user_prompt/resource_genesyscloud_architect_user_prompt.go
BrianMoyles: According to the API, names must be alphanumeric. It did not allow me to create a prompt with a non-alphanumeric name. However, I have seen some resources with non-alphanumeric names so i will add them to the check

BrianMoyles: > Is this case insensitive? I mean will "a*" match any prompt that starts with "A"?

'a*' will return resources starting with both 'a' and 'A'. I will update the test case to confirm this

BrianMoyles: > Can we have names that are not alphanumeric? Does prompts allow it?

On Further investigation, non-alphanumeric names are not valid, not needed to add them to the filter

charliecon (approved):

PR #1356, opened 2024-11-14
DO NOT MERGE - DEVTOOLING-936 - Default Outbound Routes should always exported as data resources

bbbco:
Fixes an issue that @williamjeuniaux has seen when exporting `genesyscloud_telephony_providers_edges_site_outbound_route` resources so that Default Outbound Routes are always exported as data resources since they are always autocreated by the API.


carnellj-genesys (approved):

williamjeuniaux:
Hey @bbbco !

Thanks for everything you're doing I really appreciate it!

I understand a route named "Default Outbound Route" is automatically created by the API when a site is created. Trying to recreate the route will throw an exception because the route already exists, hence why having a data resource fixes that error. 

However, I believe it hides the real issue and could result in having inconsistent data between two orgs. Let's say we have a source organization with Site A that has a default outbound route configured with Emergency and Network classifications, Random Distribution Pattern & two External Trunks. Once we create that site A on a new organization through a terraform apply, site A on the target organization will have a new Default Outbound Route which will have the default configuration for an outbound route: different network classifications, a Sequential Distribution Pattern and no external trunks. 

I think default outbound routes should be exported as normal resources as it is important that they are in sync, like the other outbound routes. I believe there should be specific logic for default outbound routes, where it handles the HTTP 400 "route already exists" returned by the API when trying to create route. Then it would retrieve the route ID and update it via PUT.

charliecon (approved):
We are doing a release tomorrow (1.53.0). After that let's get this merged to dev so you don't have to keep fixing conflicts

charliecon (commented):
I thought I was approving the naming conventions PR. Ignore that. We should address William's comment first

charliecon (changes requested):
Address William's comment about updating instead of creating

willjeuniaux-genesys:
@bbbco https://inindca.atlassian.net/browse/DEVTOOLING-936

bbbco:
This should be fixed now, but is waiting on https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1373

bbbco:
I'm closing this PR. I will be reimplementing this in a different manner with a new PR.

PR #1368, opened 2024-11-19
Feature/devtooling 836

kavinbalagen:


carnellj-genesys (approved):

genesyscloud/external_contacts_organization/data_source_genesyscloud_externalcontacts_organization.go
```diff
@@ -0,0 +1,36 @@
...
+	name := d.Get("name").(string)
+
+	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
+		externalOrganizationId, retryable, response, err := proxy.getExternalContactsOrganizationIdByName(ctx, name)
+
+		if err != nil && !retryable {
+			return retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("Error searching exteral organization %s | error: %s", name, err), response))
+		}
+
+		if retryable {
```
charliecon, line 29: Could you move this check inside the error check? That way we will only return an error if err is not nil. If for some reason `getExternalContactsOrganizationIdByName` returns nil for the error but true for retryable, then we will fail for no reason. That might not be possible based on how you wrote the code, but at least there's no risk of it happening in the future if we refactor here

genesyscloud/external_contacts_organization/genesyscloud_externalcontacts_organization_proxy.go
```diff
@@ -0,0 +1,201 @@
...
+	var externalOrganization platformclientv2.Externalorganization
+	for _, externalOrganizationSdk := range *externalOrganizations {
+		if *externalOrganization.Name == name {
+			log.Printf("Retrieved the external contacts organization id %s by name %s", *externalOrganizationSdk.Id, name)
+			externalOrganization = externalOrganizationSdk
+			return *externalOrganization.Id, false, response, nil
+		}
+	}
+
+	return "", false, response, fmt.Errorf("unable to find external contacts organization with name %s", name)
```
charliecon, line 168: retryable should be true here i think

genesyscloud/external_contacts_organization/genesyscloud_externalcontacts_organization_proxy.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	return externalOrganization, response, nil
+}
+
+// updateExternalContactsOrganizationFn is an implementation of the function to update a Genesys Cloud external contacts organization
+func updateExternalContactsOrganizationFn(ctx context.Context, p *externalContactsOrganizationProxy, id string, externalContactsOrganization *platformclientv2.Externalorganization) (*platformclientv2.Externalorganization, *platformclientv2.APIResponse, error) {
+	externalOrganization, response, err := p.externalContactsApi.PutExternalcontactsOrganization(id, *externalContactsOrganization)
+	if err != nil {
+		return nil, response, fmt.Errorf("failed to update external contacts organization: %s", err)
```
charliecon, line 188: There is no need to add extra context to the error here since we will do that in the invoking functions. It can just be one line:

```go
return p.externalContactsApi.PutExternalcontactsOrganization(id, *externalContactsOrganization)
```


 Same with the other functions in this file (create, delete, getById, etc.)

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// createExternalContactsOrganization is used by the external_contacts_organization resource to create Genesys cloud external contacts organization
+func createExternalContactsOrganization(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := getExternalContactsOrganizationProxy(sdkConfig)
+
+	externalContactsOrganization, err := getExternalContactsOrganizationFromResourceData(d)
+	if err != nil {
+		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("failed to build external organization error: %s", err), nil)
```
charliecon, line 51: We should use `util.BuildDiagnosticError` if there is no API response object.

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization.go
```diff
@@ -0,0 +1,159 @@
...
+		resourcedata.SetNillableValueWithInterfaceArrayWithFunc(d, "external_data_sources", externalOrganization.ExternalDataSources, flattenExternalDataSources)
+		dataSchema, err := flattenDataSchema(externalOrganization.Schema)
+		if err != nil {
+			return retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("failed to flatten data schema %s | error: %s", d.Id(), getErr), response))
+		}
+		d.Set("schema", dataSchema)
+		cf, err := flattenCustomFields(externalOrganization.CustomFields)
+		if err != nil {
+			return retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("failed to flatten custom fields %s | error: %s", d.Id(), getErr), response))
+		}
```
charliecon, line 107: This response object is not related to the error. We should use `BuildDiagnosticError`. Also we should be returning `err` , not `getErr`. Same with the error check directly above

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization.go
```diff
@@ -0,0 +1,159 @@
...
+
+// updateExternalContactsOrganization is used by the external_contacts_organization resource to update an external contacts organization in Genesys Cloud
+func updateExternalContactsOrganization(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := getExternalContactsOrganizationProxy(sdkConfig)
+
+	externalContactsOrganization, err := getExternalContactsOrganizationFromResourceData(d)
+
+	if err != nil {
+		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("failed to build external organization error: %s", err), nil)
```
charliecon, line 123: Same here about what error building function we should use

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization.go
```diff
@@ -0,0 +1,159 @@
...
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := getExternalContactsOrganizationProxy(sdkConfig)
+
+	response, err := proxy.deleteExternalContactsOrganization(ctx, d.Id())
+	if err != nil {
+		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("failed to delete external organization %s: %s", d.Id(), err), response)
+	}
+
+	return util.WithRetries(ctx, 180*time.Second, func() *retry.RetryError {
+		_, respCode, err := proxy.getExternalContactsOrganizationById(ctx, d.Id())
```
charliecon, line 147: respCode -> resp

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization_schema.go
```diff
@@ -0,0 +1,366 @@
...
+	"terraform-provider-genesyscloud/genesyscloud/util"
+	"terraform-provider-genesyscloud/genesyscloud/validators"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	resourceExporter "terraform-provider-genesyscloud/genesyscloud/resource_exporter"
+	registrar "terraform-provider-genesyscloud/genesyscloud/resource_register"
+)
+
+const resourceName = `genesyscloud_externalcontacts_organization`
```
charliecon, line 14: We are changing our naming conventions. This should be `resourceType` now 

More details here - https://inindca.atlassian.net/browse/DEVTOOLING-924

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/mypurecloud/platform-client-sdk-go/v146/platformclientv2"
+)
+
+func TestAccResourceExternalContacts(t *testing.T) {
+	var (
+		organizationResource = "externalOrganizationAbc"
```
charliecon, line 18: This should be `resourceLabel`

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization_test.go
```diff
@@ -0,0 +1,176 @@
...
+				ResourceName:      "genesyscloud_externalcontacts_organization." + organizationResource,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+		CheckDestroy: testVerifyContactDestroyed,
+	})
+}
+
+func GenerateBasicExternalOrganizationResource(resourceID string, name string,
```
charliecon, line 114: resourceID -> resourceLabel

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization_utils.go
```diff
@@ -0,0 +1,407 @@
...
+
+	}
+	externalOrganization.CustomFields = customFields
+
+	return externalOrganization, nil
+}
+
+// buildPhonenumberFromData is a helper method to map phone data to the GenesysCloud platformclientv2.PhoneNumber
+func buildPhonenumberFromData(phoneData []interface{}) *platformclientv2.Phonenumber {
+	phoneMap := phoneData[0].(map[string]interface{})
```
charliecon, line 61: A type-check here could help avoid a panic error in the future 

```go
phoneMap, ok := ...
if !ok {
  return nil
}
```

kavinbalagen: It looks like the calling function does a nil check, do we still need to include it?
image

charliecon: I think that would be checking that the interface array is not nil. Here we're casting the item inside the array. So if phoneData has one nil item, phoneData is not nil but we will see a panic error if we try to cast the item inside phoneData to a map without the `, ok :=`

charliecon: This can happen if the user defines phone_number and doesn't specify any attributes 

```
phone_number {}
```

genesyscloud/external_contacts_organization/genesyscloud_externalcontacts_organization_proxy.go
```diff
@@ -0,0 +1,201 @@
...
+		rc.SetCache(p.externalOrganizationCache, *externalOrganization.Id, externalOrganization)
+	}
+
+	return &allExternalOrganizations, response, nil
+}
+
+// getExternalContactsOrganizationIdByNameFn is an implementation of the function to get a Genesys Cloud external contacts organization by name
+func getExternalContactsOrganizationIdByNameFn(ctx context.Context, p *externalContactsOrganizationProxy, name string) (id string, retryable bool, apiResponse *platformclientv2.APIResponse, err error) {
+
+	externalOrganizations, response, err := getAllExternalContactsOrganizationFn(ctx, p, name)
```
HemanthDogiparthi12, line 150: related to comment above. can we have cache lookup for datasource. before making actual api call

genesyscloud/external_contacts_organization/data_source_genesyscloud_externalcontacts_organization.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func dataSourceExternalContactsOrganizationRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := newExternalContactsOrganizationProxy(sdkConfig)
+
+	name := d.Get("name").(string)
+
+	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
+		externalOrganizationId, retryable, response, err := proxy.getExternalContactsOrganizationIdByName(ctx, name)
```
HemanthDogiparthi12, line 23: can you add datasource cache here ? evry datasource lookup would make an entire getall fn call if cache is not enabled.

genesyscloud/external_contacts_organization/resource_genesyscloud_externalcontacts_organization_utils.go
```diff
@@ -0,0 +1,407 @@
...
+
+func BuildOrganizationSchema(schemaList []interface{}) (*platformclientv2.Dataschema, error) {
+	if len(schemaList) > 0 {
+		schemaMap := schemaList[0].(map[string]interface{})
+		dataSchema := &platformclientv2.Dataschema{}
+		dataSchema.Name = platformclientv2.String(schemaMap["name"].(string))
+		dataSchema.Enabled = platformclientv2.Bool(schemaMap["enabled"].(bool))
+		version := platformclientv2.Int(schemaMap["version"].(int))
+		if *version != 0 {
+			dataSchema.Version = version
```
HemanthDogiparthi12, line 219: are we using any default schema ?

kavinbalagen: There is no default schema but the user is allowed to use a particular version of a schema by pointing to its version & Id.

charliecon (commented):
We are missing the new files inside `/examples/`

genesyscloud/external_contacts_organization/data_source_genesyscloud_externalcontacts_organization.go
kavinbalagen: done

kavinbalagen:
> We are missing the new files inside `/examples/`
added

charliecon (approved):
Looks good. Just need to run `make docs` one more time

PR #1374, opened 2024-11-19
Bug/devtooling 877 Airbnb - groupmembers division more than 50 with "*" issue

shrutisuryawanshigenesys:


charliecon (approved):

genesyscloud/routing_skill_group/resource_genesyscloud_routing_skill_group_utils.go
```diff
@@ -95,11 +94,9 @@ func createListsForSkillgroupsMembersDivisions(schemaMemberDivisionIds []string,
 	toRemove := make([]string, 0)
 
 	if allMemberDivisionsSpecified(schemaMemberDivisionIds) {
-		if len(schemaMemberDivisionIds) > 1 {
-			return nil, nil, util.BuildDiagnosticError(resourceName, fmt.Sprintf(`member_division_ids should not contain more than one item when the value of an item is "*"`), fmt.Errorf(`member_division_ids should not contain more than one item when the value of an item is "*"`))
-		}
-		toAdd, err := getAllAuthDivisionIds(meta)
-		return toAdd, nil, err
+		// member_division_ids should not contain more than one item when the value of an item is "*"
```
HemanthDogiparthi12, line 97: for a type list you can add in validatefunc in schema definition.

ValidateFunc: validateMemberIds,

and the method signature can look like this 


func validateMemberIds(val interface{}, key string) (err []error) {}


you can typecast val to 
list, ok := val.([]interface{}) 

then search for * and length for this list.

PR #1377, opened 2024-11-21
JM-1465: Last Update about Charts for Journey Management

nicob-29:


genesyscloud/journey_views/genesyscloud_journey_views_proxy.go
```diff
@@ -114,8 +115,9 @@ func createJourneyViewFn(_ context.Context, p *journeyViewsProxy, journeyView *p
 	return p.journeyViewsApi.PostJourneyViews(*journeyView)
 }
 
-func updateJourneyViewFn(_ context.Context, p *journeyViewsProxy, viewId string, journeyView *platformclientv2.Journeyview) (*platformclientv2.Journeyview, *platformclientv2.APIResponse, error) {
-	return p.journeyViewsApi.PostJourneyViewVersions(viewId, *journeyView)
+func updateJourneyViewFn(_ context.Context, p *journeyViewsProxy, viewId string, versionId int, journeyView *platformclientv2.Journeyview) (*platformclientv2.Journeyview, *platformclientv2.APIResponse, error) {
+	versionIdToString := strconv.Itoa(versionId)
+	return p.journeyViewsApi.PutJourneyViewVersion(viewId, versionIdToString, *journeyView)
```
charliecon, line 120: If the user does not provide a value for version, it will be the zero value for integers which is 0 

If you don't want 0 to be the specified version in this scenario, I would change this slightly.

```go
var versionIdToString string
if versionId != 0 {
  versionIdToString = strconv.Itoa(versionId)
}
```

This way it will only specify a version in the API call if the terraform user provided one

genesyscloud/journey_views/resource_genesyscloud_journey_views.go
```diff
@@ -57,12 +57,14 @@ func createJourneyView(ctx context.Context, d *schema.ResourceData, meta interfa
 
 func updateJourneyView(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	name := d.Get("name").(string)
+	versionId := d.Get("version").(int)
```
charliecon, line 60: It is possible that if the user does not specify a version, casting to int here *could* cause a conversion error to be thrown. I would change this slightly to make sure this does not happen 

```diff
-versionId := d.Get("version").(int)
+versionId, _ := d.Get("version").(int)
```

charliecon (approved):

genesyscloud/journey_views/resource_genesyscloud_journey_views.go
nicob-29: No risk because the version is coming from API. It is computed information. So not provided by user and API returns always a version

genesyscloud/journey_views/genesyscloud_journey_views_proxy.go
nicob-29: Terraform user can't not provide it . It is computed property on schema

charliecon: Computed means that if the user does not provide a value, then the API will decide what it should be, but the user can still choose to provide the value. 

If the user is not meant to set the value and you want the API to always take care of the version, then I would not include it in the schema. Better to use [GET /api/v2/journey/views/{viewId}](https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-journey-views--viewId-) to establish the latest version, save it to a variable, and then use that variable in the call to [PUT /api/v2/journey/views/{viewId}/versions/{versionId}](https://developer.genesys.cloud/devapps/api-explorer#put-api-v2-journey-views--viewId--versions--versionId-)

nicob-29: For it is the case, we force the version on state after each API calls 
See readJourneyView  https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/journey_views/resource_genesyscloud_journey_views.go#L55

charliecon: Ah I understand now. Because Computed is true and Optional is omitted, it is a read-only field. Since it is read-only, there is no need to read it from the schema during an update. You can remove the line 

```go
versionId := d.Get("version").(int)
```

To establish the version, use the proxy method `getJourneyViewById` , and then call update. The update function will look something like this:

```go
func updateJourneyViewFn(ctx context.Context, p *journeyViewsProxy, viewId string, journeyView *platformclientv2.Journeyview) (*platformclientv2.Journeyview, *platformclientv2.APIResponse, error) {
	viewRead, resp err := p.getJourneyViewById(ctx, viewId)
        if err != nil {
            return nil, resp, err
        }
        if viewRead.Version == nil {
            return nil, nil, fmt.Errorf("failed to establish view '%s' version for update", viewId)
        }
        
	return p.journeyViewsApi.PutJourneyViewVersion(viewId, *viewRead.Version, *journeyView)
}
```

charliecon: True. The create will call the read which will set the version in the schema, and then we can use that in the update. I didn't think through all that. You're right

nicob-29: Thanks for your support. I appreciated

PR #1373, opened 2024-11-19
DEVTOOLING-924 pt2 - Refactor/exporter resource types

bbbco:
Here is the second PR for DEVTOOLING-924 where we refactor inconsistent usages:
 - `resourceName` to `ResourceType` when referring to the type of resource. 
     - This also allows us to reference the ResourceType for importing in test initialization
 - `resourceName` to `resourcePath` when referring to the combined path of type + label
 - There are a few other variable name fixes along the way.

PR #1400, opened 2024-12-02
feature/DEVTOOLING-163

sureshperiyappan:
Code refactored for Journey Action Map


genesyscloud/journey_action_map/genesyscloud_journey_action_map_init_test.go
```diff
@@ -0,0 +1,82 @@
...
+	datasourceMapMutex sync.RWMutex
+}
+
+// registerTestResources registers all resources used in the tests
+func (r *registerTestInstance) registerTestResources() {
+	r.resourceMapMutex.Lock()
+	defer r.resourceMapMutex.Unlock()
+
+	providerResources[resourceName] = ResourceJourneyActionMap()
+	providerResources["genesyscloud_auth_division"] = authDivision.ResourceAuthDivision()
```
bbbco, line 45: You can now refer to the provider resource types like so: instead of `"genesyscloud_auth_division"` you can use `authDivision.ResourceType` and the same for all the other non-`gcloud` resources.

bbbco (changes requested):

sureshperiyappan:
> It looks like you need to merge some changes from dev that were pulled in from #1373 .

I think #1373 change was recently merged in to dev, I will have to update.

PR #1404, opened 2024-12-02
No-JIRA: Allow user to set number of times consistency checker runs

dginty4:
This PR adds another env variable that can be used to customise the consistency checker. `Consistency checks` can be used to allow the user to set how many times they want the consistency checker to retry. Setting it to 0 will disable the consistency checker and stop it from running. This env variable is only used if `BYPASS_CONSISTENCY_CHECKER` is set. 

I also added all this to the README.


HemanthDogiparthi12 (approved):
looks good to merged, once conflicts are resolved

PR #1413, opened 2024-12-05
No jira: fix attempt limit test

charliecon:
```
    data_source_genesyscloud_outbound_attempt_limit_test.go:19: Step 1/1 error: Error running pre-apply refresh: exit status 1
        
        Error: Invalid resource type
        
          on terraform_plugin_test.tf line 2, in resource "genesyscloud_outbound_attempt_limit" "attempt_limit":
           2: resource "genesyscloud_outbound_attempt_limit" "attempt_limit" {
        
        The provider hashicorp/genesyscloud does not support resource type
        "genesyscloud_outbound_attempt_limit". Did you mean
        "genesyscloud_outbound_attemptlimit"?
--- FAIL: TestAccDataSourceOutboundAttemptLimit (0.20s)
```

### DIFF
--- a/genesyscloud/outbound_attempt_limit/resource_genesyscloud_outbound_attempt_limit.go
+++ b/genesyscloud/outbound_attempt_limit/resource_genesyscloud_outbound_attempt_limit.go
@@ -182,7 +182,7 @@ func createOutboundAttemptLimit(ctx context.Context, d *schema.ResourceData, met
 	if resetPeriod != "" {
 		sdkAttemptLimits.ResetPeriod = &resetPeriod
 	}
-	if recallEntries != nil && len(recallEntries) > 0 {
+	if len(recallEntries) > 0 {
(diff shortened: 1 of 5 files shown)


genesyscloud/outbound_attempt_limit/resource_genesyscloud_outbound_attempt_limit.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	ResourceType = "genesyscloud_outbound_attemptlimit"
+	ResourceType = "genesyscloud_outbound_attempt_limit"
```
charliecon, line 26: This was what caused the tests to fail. It should be "attempt_limit" instead of "attemptlimit". 

The `SetRegistrar` function still used the correct string which is why it made no difference to the docs

HemanthDogiparthi12 (approved):

PR #1425, opened 2024-12-09
DEVTOOLING-966: Verifying schedule start date is a weekday

charliecon:


genesyscloud/architect_schedules/resource_genesyscloud_architect_schedules_test.go
```diff
@@ -105,23 +111,112 @@ func TestAccResourceArchitectSchedules(t *testing.T) {
 	})
 }
 
+func TestAccResourceArchitectSchedulesCreateFailsWhenStartDateIsNotWeekday(t *testing.T) {
```
carnellj-genesys, line 114: Is it possible for a start date in on a Saturday.  I just want to make sure we don't have problems with an export where  we export something that is on Saturday and then we cant reimport it.  Just want to make sure we are good here.

carnellj-genesys (approved):

kavinbalagen (approved):

genesyscloud/architect_schedules/resource_genesyscloud_architect_schedules_test.go
charliecon: I found schedules with weekend start dates already present in two different orgs. I posted in the Architect Support room for more info on this. I'm assuming they could have been created before this rule was enforced.

PR #1430, opened 2024-12-13
DEVTOOLING-966: Validating that the schedule start date matches the rrule

charliecon:
If an architect schedule defines its recurrence rule (`rrule`) as `"FREQ=WEEKLY;INTERVAL=1;BYDAY=MO,TU,WE"` , then the start date must be a Monday, Tuesday or Wednesday. If it is not, the API will automatically push the day back until it matches the rrule (in this case it will transform to the next Monday)  

This PR adds a check to the create and update context to establish the days specified in the rrule (if any), and then establish which day of the week the start date is. If the start date day is not in the rrule, fail. 

@carnellj-genesys After exporting all schedules in an org and applying them against another org, I found that there were schedules that did not conform to the rules (i.e. had start dates that were not in the rrule), so the deployment failed. I don't think this is an issue from a CX as Code perspective because there are three possibilities:

1. We leave the code as is and the consistency checker will retry for ages before outputting a vague error message:
    
Screenshot 2024-12-13 at 10 27 47

2. The user bypasses the consistency checker which will get them around the error, but every `terraform apply/plan` will produce changes because of the inconsistency between the config and API.

3. We enforce this validation which will cause there deployment to fail quickly, providing an explanation:

Screenshot 2024-12-13 at 10 28 10

### DIFF
--- a/genesyscloud/architect_schedules/resource_genesyscloud_architect_schedules_utils.go
+++ b/genesyscloud/architect_schedules/resource_genesyscloud_architect_schedules_utils.go
@@ -1,18 +1,60 @@
 package architect_schedules
 
 import (
-	"errors"
 	"fmt"
+	"regexp"
+	"strings"
+	"terraform-provider-genesyscloud/genesyscloud/util/lists"
 	"time"
 )
 
 const timeFormat = "2006-01-02T15:04:05.000000"
 
-func verifyDateTimeIsWeekday(dateTime time.Time) error {
-	if dateTime.Weekday() == time.Saturday || dateTime.Weekday() == time.Sunday {
-		return errors.New("schedule start date cannot be a Saturday or Sunday")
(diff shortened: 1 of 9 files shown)


genesyscloud/architect_schedules/resource_genesyscloud_architect_schedules_utils.go
```diff
@@ -0,0 +1,77 @@
...
+	if len(matches) < 2 {
+		// No BYDAY parameter found
+		return []string{}
+	}
+
+	// Split the matched days on commas
+	days := strings.Split(matches[1], ",")
+
+	// Map the abbreviations to full day names
+	abbreviationMap := map[string]string{
```
HemanthDogiparthi12, line 39: you can put this outside the function !!

PR #1433, opened 2024-12-16
line properties added

kavinbalagen:


genesyscloud/telephony_providers_edges_phonebasesettings/genesyscloud_telephony_providers_edges_phonebasesettings_schema.go
```diff
@@ -68,9 +68,28 @@ var (
...
 	}
-)
 
-func ResourcePhoneBaseSettings() *schema.Resource {
+	lineBase := &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"line_meta_base_id": {
+				Description: "lineMetaBaseId is computed by the provider based on the phoneBaseSettings using the phoneBaseSetting's template",
+				Type:        schema.TypeString,
+				Optional:    true,
```
bbbco, line 77: Optional and Required should be set to `false` if this is truly a computed only attribute.

bbbco (approved):

genesyscloud/telephony_providers_edges_phonebasesettings/genesyscloud_telephony_providers_edges_phonebasesettings_schema.go
kavinbalagen: it cannot be read only, because we have some users who export the resources from one org and install them to another, in those cases the provider will force them to remove that field.

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings.go
```diff
@@ -118,6 +121,9 @@ func updatePhoneBaseSettings(ctx context.Context, d *schema.ResourceData, meta i
 			State:        (*phoneBaseSettings.Lines)[0].State,
 		},
 	}
+	if lineProperties := BuildTelephonyLineBaseProperties(d); lineProperties != nil {
+		(*phoneBase.Lines)[0].Properties = lineProperties
+	}
```
charliecon, line 126: We should be careful here to not overwrite all the data in Properties that is generated by the API after creation. The team that owns the API might be able to advise more on this, but I think it would be better to read the current state of `Properties` for this line and update just the bool and int field.

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings_test.go
```diff
@@ -23,6 +23,8 @@ func TestAccResourcePhoneBaseSettings(t *testing.T) {
 		description1                   = "test description 1"
 		description2                   = "test description 2"
 		phoneMetaBaseId                = "generic_sip.json"
+		stationPersistTimeout          = "2000"
+		stationPersistTimeoutUpdate    = "3000"
```
charliecon, line 27: With `line_meta_base_id` as "generic_sip.json" , I noticed that the bool and int fields do not appear in the Admin UI under General for a phone base setting. But they do for "inin_webrtc_softphone.json". It might be worth investigating it these new fields are only relevant for a particular phone_meta_base_id type

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings_utils.go
```diff
@@ -130,3 +130,52 @@ func customizePhoneBaseSettingsPropertiesDiff(ctx context.Context, diff *schema.
 
 	return util.ApplyPropertyDefaults(diff, phoneBaseSetting.Properties)
 }
+
+func BuildTelephonyLineBaseProperties(d *schema.ResourceData) *map[string]interface{} {
+
+	if lineBase := d.Get("line_base").([]interface{}); len(lineBase) > 0 {
+
+		lineBaseMap := lineBase[0].(map[string]interface{})
```
charliecon, line 138: Since there are no required fields in the `line_base` schema, we should be careful when casting the item to a map as it might be nil.

```go
lineBaseMap := lineBase[0].(map[string]interface{})
``` 

⬇️ 

```go
lineBaseMap, ok := lineBase[0].(map[string]interface{})
if !ok {
  return nil
}
```

I guess we could either make this change or make `station_persistent_enabled` Required.

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings_utils.go
```diff
@@ -130,3 +130,52 @@ func customizePhoneBaseSettingsPropertiesDiff(ctx context.Context, diff *schema.
...
+	}
+
+	lineBaseMap := make(map[string]interface{})
+	resourcedata.SetMapValueIfNotNil(lineBaseMap, "line_meta_base_id", (*lineBase)[0].LineMetaBase.Id)
+	propertiesObject := (*lineBase)[0].Properties
+	if propertiesObject == nil {
+		return []interface{}{lineBaseMap}
+	}
+	if enabledKey := (*propertiesObject)["station_persistent_enabled"]; enabledKey != nil {
+		enabledValue := enabledKey.(map[string]interface{})["value"].(map[string]interface{})["instance"]
```
charliecon, line 169: We should be able to cast to map on line 168 instead. I think it would make the code a bit cleaner 

```go
if enabledKeyMap, ok := (*propertiesObject)["station_persistent_enabled"].(map[string]interface{}); ok && enabledKeyMap != nil {
  // the added _ below, along with the cast to bool at the end,  should also mean that the value will default to the zero value "false" for bool 
  enabledValue, _ := enabledKeyMap["value"].(map[string]interface{})["instance"].(bool)
```

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings_utils.go
```diff
@@ -130,3 +130,52 @@ func customizePhoneBaseSettingsPropertiesDiff(ctx context.Context, diff *schema.
...
+		if enabledValue != nil {
+			resourcedata.SetMapValueIfNotNil(lineBaseMap, "station_persistent_enabled", &enabledValue)
+		}
+	}
+	if timeOutKey := (*propertiesObject)["station_persistent_timeout"]; timeOutKey != nil {
+		timeOutKey := timeOutKey.(map[string]interface{})["value"].(map[string]interface{})["instance"]
```
charliecon, line 175: Same here

charliecon (changes requested):

docs/resources/telephony_providers_edges_phonebasesettings.md
```diff
@@ -71,7 +71,7 @@ resource "genesyscloud_telephony_providers_edges_phonebasesettings" "examplePhon
 
 - `capabilities` (Block List, Max: 1) Phone Capabilities. (see [below for nested schema](#nestedblock--capabilities))
 - `description` (String) The resource's description.
-- `line_base_settings_id` (String) Computed line base settings id
```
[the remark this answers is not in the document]
kavinbalagen: will revert to its original position and name.

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings.go
kavinbalagen: for the properties attributes across telephony, the backend only mutates the fields that we send, the default fields are not overwritten.

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings_test.go
kavinbalagen: We see the attributes under the general tab for generic_sip, also there are several types of telephony that aren't seen in hollywoo, so it may not be easy to add a validation

genesyscloud/telephony_providers_edges_phonebasesettings/resource_genesyscloud_telephony_providers_edges_phonebasesettings_utils.go
kavinbalagen: done

charliecon (approved):

PR #1450, opened 2025-01-06
GPE-14846: Add External Sources resource

michalk-genesys:
Adding  External Sources resource to manage resources with /api/v2/externalcontacts/externalsources API


docs/resources/externalcontacts_external_source.md
```diff
@@ -0,0 +1,54 @@
...
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `name` (String) The name of the external source.
+
+### Optional
+
+- `active` (Boolean) Whether the external source is active. Defaults to `true`.
```
michalk-genesys, line 41: When creating new External Sources, the backing API will ignore the value from the request and always create resource with `active` set to `true`.
I thought I'd call this out, in case this is a problem for TF provider or our users.

docs/resources/externalcontacts_external_source.md
```diff
@@ -0,0 +1,54 @@
...
+### Read-Only
+
+- `id` (String) The ID of this resource.
+
+<a id="nestedblock--link_configuration"></a>
+### Nested Schema for `link_configuration`
+
+Optional:
+
+- `uri_template` (String)
```
michalk-genesys, line 53: This string allows templating, with `{{}}`. The only value supported for now is {{externalId.value}}. Should this be called out in the field description?
This is not mentioned in Public API docs.

genesyscloud/external_contacts_external_source/genesyscloud_externalcontacts_external_source_proxy.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// getAllExternalContactsExternalSourceFn is the implementation for retrieving all external contacts external source in Genesys Cloud
+func getAllExternalContactsExternalSourceFn(ctx context.Context, p *externalContactsExternalSourceProxy, query string) (*[]platformclientv2.Externalsource, *platformclientv2.APIResponse, error) {
+	var allExternalSources []platformclientv2.Externalsource
+	var response *platformclientv2.APIResponse
+
+	cursor := ""
+	for {
+		externalSources, resp, err := p.externalContactsApi.GetExternalcontactsExternalsources(cursor, 100, query, true) // workaround for active being a required arg on Platform Client SDK Go GetExternalcontactsExternalsources
```
michalk-genesys, line 105: I left a comment in the code to make this more visible. The method in our SDK requires passing `active` as an arg and this is not optional, so I ended up making two requests: first one for all active:true and the second for all active:false.

genesyscloud/external_contacts_external_source/resource_genesyscloud_externalcontacts_external_source.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v149/platformclientv2"
+
+	"terraform-provider-genesyscloud/genesyscloud/consistency_checker"
+	"terraform-provider-genesyscloud/genesyscloud/provider"
+	"terraform-provider-genesyscloud/genesyscloud/util/resourcedata"
+)
+
+// getAllAuthExternalContactsExternalSources retrieves all of the external sources via Terraform in the Genesys Cloud and is used for the exporter
+func getAllAuthExternalContactsExternalSources(ctx context.Context, clientConfig *platformclientv2.Configuration) (resourceExporter.ResourceIDMetaMap, diag.Diagnostics) {
```
michalk-genesys, line 23: This is similar resource to External Organizations and can have same problem as the one dealt with in DEVTOOLING-977.
The API can return more than 1000 resources, as there is no upper limit on how many External Sources we can have in total. The limit is only applied on the number of active ones.

docs/resources/externalcontacts_external_source.md
michalk-genesys: A change is being proposed to address this in the backend, so that the `active` value from the request is honoured.

docs/resources/externalcontacts_external_source.md
charliecon: If more values are supported in the future, we'll have to come back and modify the docs here, so it might be easier to be generic and let the response tell them what they did wrong.

genesyscloud/external_contacts_external_source/resource_genesyscloud_externalcontacts_external_source_schema.go
```diff
@@ -0,0 +1,86 @@
...
+
+		CreateContext: provider.CreateWithPooledClient(createExternalContactsExternalSource),
+		ReadContext:   provider.ReadWithPooledClient(readExternalContactsExternalSource),
+		UpdateContext: provider.UpdateWithPooledClient(updateExternalContactsExternalSource),
+		DeleteContext: provider.DeleteWithPooledClient(deleteExternalContactsExternalSource),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
```
charliecon, line 41: I noticed `id` is a field in the request body for the POST. Can the client choose the ID? If they can, we should probably add it here just so the end user can do anything from CX as Code that they can do via the API.

michalk-genesys: It seems the `id` from POST request body is ignored and new one is generated on the backend. Not sure if this happens in Public API or in the backend service, but that seems to be the end result.

genesyscloud/external_contacts_external_source/resource_genesyscloud_externalcontacts_external_source.go
michalk-genesys: With DEVTOOLING-977 being marked as done, I assume we're handling cases when there is more than 1000 external sources to retrieve.

docs/resources/externalcontacts_external_source.md
michalk-genesys: FYI, this fix has been deployed to prod for some time now. Creating inactive external sources is now working as expected.

genesyscloud/external_contacts_external_source/resource_genesyscloud_externalcontacts_external_source_schema.go
charliecon: Understood. That's common across the API

charliecon (approved):

PR #1463, opened 2025-01-10
Fixes DEVTOOLING-1029

bbbco:
Fixed DEVTOOLING-1029 by adding support for detecting the binary executing the provider and including logical support for determining Terraform vs OpenTofu

Also happens to fix https://github.com/MyPureCloud/terraform-provider-genesyscloud/issues/1475 when `terraform` or `tofu` is not in the `PATH`

### DIFF
--- a/genesyscloud/platform/platform.go
+++ b/genesyscloud/platform/platform.go
@@ -1,4 +1,4 @@
-package util
+package platform
 
 import (
 	"bytes"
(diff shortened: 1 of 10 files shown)


bbbco:
I had Amazon Q review this PR and check it for issues with the following prompt:
```
Check this library for any potential issues, including but not limited to nil pointer errors, potential areas that could panic, errors that are not handled properly, misused routines and/or channels, incorrect interface usage, forgetting to use `defer`, race conditions and concurrency primitives, and any other concerns that might impact the stability of the application.
```
I addressed all of the items it brought up. I built and tested the provider and ensured that it worked as expected.

carnellj-genesys (approved):

charliecon (approved):

PR #1478, opened 2025-01-14
Fix DEVTOOLING-1031 to ensure passwords will be updated

bbbco:
Fix DEVTOOLING-1031 to ensure passwords will be updated when changed or when reactivating inactive user

### DIFF
--- a/genesyscloud/user/genesyscloud_user_proxy.go
+++ b/genesyscloud/user/genesyscloud_user_proxy.go
@@ -33,6 +33,7 @@ type deleteUserFunc func(ctx context.Context, p *userProxy, id string) (*interfa
 type patchUserWithStateFunc func(ctx context.Context, p *userProxy, id string, updateUser *platformclientv2.Updateuser) (*platformclientv2.User, *platformclientv2.APIResponse, error)
 type hydrateUserCacheFunc func(ctx context.Context, p *userProxy, pageSize int, pageNum int) (*platformclientv2.Userentitylisting, *platformclientv2.APIResponse, error)
 type getUserByNameFunc func(ctx context.Context, p *userProxy, searchUser platformclientv2.Usersearchrequest) (*platformclientv2.Userssearchresponse, *platformclientv2.APIResponse, error)
+type updatePasswordFunc func(ctx context.Context, p *userProxy, id string, password string) (bool, *platformclientv2.APIResponse, error)
 
 /*
 The userProxy struct holds all the methods responsible for making calls to
@@ -54,6 +55,7 @@ type userProxy struct {
 	patchUserWithStateAttr patchUserWithStateFunc
 	hydrateUserCacheAttr   hydrateUserCacheFunc
 	getUserByNameAttr      getUserByNameFunc
+	updatePasswordAttr     updatePasswordFunc
 	userCache              rc.CacheInterface[platformclientv2.User] //Define the cache for user resource
 }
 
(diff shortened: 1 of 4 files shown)


genesyscloud/user/resource_genesyscloud_user_utils.go
```diff
@@ -115,6 +115,11 @@ func executeAllUpdates(d *schema.ResourceData, proxy *userProxy, sdkConfig *plat
 		return diagErr
 	}
 
+	diagErr = updatePassword(ctx, d, proxy)
```
bbbco, line 118: @carnellj Here's the `updatePassword` call

genesyscloud/user/resource_genesyscloud_user.go
```diff
@@ -123,7 +123,7 @@ func createUser(ctx context.Context, d *schema.ResourceData, meta interface{}) d
 		}
 	}
 
-	diagErr := executeAllUpdates(d, proxy, sdkConfig, false)
+	diagErr := executeAllUpdates(ctx, d, proxy, sdkConfig, false)
```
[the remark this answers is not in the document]
bbbco: https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1478/files#r1920632820

carnellj-genesys (approved):

genesyscloud/user/resource_genesyscloud_user_utils.go
```diff
@@ -287,6 +292,28 @@ func updateUserRoutingUtilization(d *schema.ResourceData, proxy *userProxy) diag
...
 
+func updatePassword(ctx context.Context, d *schema.ResourceData, proxy *userProxy) diag.Diagnostics {
+	if !d.HasChange("password") {
+		return nil
+	}
+
+	password, ok := d.Get("password").(string)
+	if !ok {
+		return util.BuildDiagnosticError(ResourceType, "Invalid password type", fmt.Errorf("password must be a string"))
+	}
```
charliecon, line 303: I think ok will be false here if password is an empty string. Terraform will fail right away if the wrong type is assigned to a field.

genesyscloud/user/genesyscloud_user_proxy.go
```diff
@@ -272,3 +280,14 @@ func getUserIdByNameFn(ctx context.Context, p *userProxy, name string) (id strin
...
+
+func updatePasswordFn(ctx context.Context, p *userProxy, userId string, newPassword string) (bool, *platformclientv2.APIResponse, error) {
+	// Get the user's current password
+	resp, err := p.userApi.PostUserPassword(userId, platformclientv2.Changepasswordrequest{
+		NewPassword: &newPassword,
+	})
+	if err != nil {
+		return false, resp, err
+	}
+	return true, resp, nil
```
charliecon, line 292: Is there any need for the bool here? I don't see it being used anywhere

charliecon (approved):

genesyscloud/user/resource_genesyscloud_user_utils.go
charliecon: If password is not provided by the user, then requesting the second return value (`ok`) will ensure that `password` is the zero value for whatever type we are casting to right of `:=`

I know this is how it works when casting map items... pretty sure its the same casting things returned from `d.Get`

bbbco: Ah. Are you thinking we just remove these extra checks then and let the API handle the password update straight away?

PR #1483, opened 2025-01-16
Update empty resources section for architect prompt

HemanthDogiparthi12:


genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
```diff
@@ -496,6 +502,23 @@ func (p *architectUserPromptProxy) buildUserPromptResourcesForCreateAndUpdate(ct
...
+	if resources != nil && len(resources.List()) == 1 {
+		for _, promptResource := range resources.List() {
+			promptResourceMap, ok := promptResource.(map[string]interface{})
+			if !ok {
+				continue
+			}
+
+			resourceLanguage := promptResourceMap["language"].(string)
+			if resourceLanguage == "" {
+				return true
```
charliecon, line 515: Is there a chance one resource in the list will have `language = ""` but the others dont'? It looks like in that case true will be returned and resources will be nil.

genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
```diff
@@ -496,6 +502,23 @@ func (p *architectUserPromptProxy) buildUserPromptResourcesForCreateAndUpdate(ct
 	return toCreate, toUpdate, toDelete, nil, nil
 }
 
+func checkEmptyResource(resources *schema.Set) bool {
```
charliecon, line 505: A comment above this function to explain its purpose would be helpful

genesyscloud/architect_user_prompt/resource_architect_user_prompt_schema.go
```diff
@@ -49,7 +49,7 @@ var userPromptResource = &schema.Resource{
 		"language": {
 			Description:  "Language for the prompt resource. (eg. en-us)",
 			Type:         schema.TypeString,
-			Required:     true,
+			Optional:     true,
```
charliecon, line 52: If we're making this optional, we might have to change the functions `buildUserPromptResourceForCreate` & `buildUserPromptResourceForUpdate` slightly since they're assuming it is required. Also will need to run `make docs`

genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
HemanthDogiparthi12: this is only applicable for one resource list. you can look at line 506 above.

genesyscloud/architect_user_prompt/resource_architect_user_prompt_schema.go
HemanthDogiparthi12: yes. this is taken care in "buildUserPromptResourcesForCreateAndUpdate"

charliecon (approved):

PR #1489, opened 2025-01-20
[Devtooling 959] Added 'allowSessionUpgrade' to webdeployment_configurations

fergalmcl:


dginty4 (approved):

charliecon (commented):
Can we incorporate this field into the tests? At the moment it doesn't look like anything inside `authentication_settings` is being tested. If it's too much trouble to test these fields with acceptance tests, we can do unit tests

charliecon (approved):

PR #1499, opened 2025-01-23
JM-2290 NumberPredicates

nicob-29:


examples/resources/genesyscloud_journey_views/resource.tf
```diff
@@ -1,22 +1,114 @@
...
-  name     = "Sample Journey 1"
   elements {
-    id   = "ac6c61b5-1cd4-4c6e-a8a5-edb74d9117eb"
-    name = "Wrap Up"
+    id   = var.element_first_id
+    name = "Bot Start"
     attributes {
       type   = "Event"
-      id     = "a416328b-167c-0365-d0e1-f072cd5d4ded"
+      id     = "94677135-0e97-1a87-44f0-ee7f5506bd09"
```
charliecon, line 21: Could any of these ids (metrics.id or attribute.id) be references to other existing resources? If yes, it is better to have a resource reference in the example instead of a hard-coded ID

charliecon (approved):
Looks good 👌 Just one comment

examples/resources/genesyscloud_journey_views/resource.tf
nicob-29: metrics.id: it is just an id.
attribute.id: it is hardcoded value. This value is provided by other service Trinity. 
https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-journey-views-eventdefinitions

PR #1506, opened 2025-01-25
Feature/devtooling-1019 Bulk Contact Management within the Contact Lists resource

bbbco:
This PR resolves DEVTOOLING-1019 and makes a major shift away from managing individual contacts as individual resources to consolidating all the contacts in a contact list into a single CSV file that is user managed. These changes use the GC API's export and upload endpoints to enable bulk within the existing `genesyscloud_outbound_contact_list` resource and provides a seamless migration path with documentation.

Please take extra special note of the schema file, as it includes a number of strategies that I am proposing we continue in future work:
- Bumping schema version
- Using CustomizeDiff function for extra validations
- Setting the `hash` attribute to computed and managing the computations ourselves (instead of depending on the user to set the sha calculations)
- When managing bulk items, add a computed record count attribute to track the number of records

Also, due to the nature of this PR and to prevent confusion and/or duplication when exporting contact resources, I took the liberty of removing the `contact_list_contact` resource exporter in favor of this bulk exporter built into `contact_list` resource. I added deprecation and migration notices, and added logic to ensure the resource gets tagged in the `Deprecated` subcategory in the rendered docs.

I have checked this work against Amazon Q. I also used Amazon Q to help me develop extra tests across a number of different functions and libraries and ensured the tests are passing.


templates/resources.md.tmpl
```diff
@@ -1,6 +1,10 @@
 ---
 page_title: "{{.Name}} {{.Type}} - {{.ProviderName}}"
+{{ if gt (len (split .Description "[DEPRECATED]")) 1 -}}
```
bbbco, line 3: See https://developer.hashicorp.com/terraform/registry/providers/docs#subcategories for details about subcategory, and I got the logic from https://github.com/hashicorp/terraform-plugin-docs/issues/156#issuecomment-1600427216

examples/resources/genesyscloud_outbound_contact_list_contact/apis.md
```diff
@@ -1,4 +1,31 @@
...
+The `genesyscloud_outbound_contact_list_contact` resource is deprecated and will be removed in a future version. Instead, use the `contacts_filepath` and `contacts_id_name` attributes in the `genesyscloud_outbound_contact_list` resource.
+
+### Note About Exporter
+
+The exporter functionality has been removed from this resource in favor of the `genesyscloud_outbound_contact_list` resource's built-in bulk handling of contacts via CSV exports. Contacts will now be exported within the CSV file output from the `genesyscloud_outbound_contact_list` and not be exported via this resource due to performance and scalability limitations with this resource.
+
+### Migration Steps
+
+1. Remove any `genesyscloud_outbound_contact_list_contact` resources from your Terraform configuration and add them to a `contacts.csv` file
+2. Update your `genesyscloud_outbound_contact_list` resource to include:
```
HemanthDogiparthi12, line 19: you can add the previous resource example too, just to have 'from' and 'to' at a single place.

genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_utils.go
```diff
@@ -168,6 +187,85 @@ func flattenSdkOutboundContactListColumnDataTypeSpecifications(columnDataTypeSpe
...
+	if err := os.MkdirAll(fullDirectoryPath, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create directory %s: %w", fullDirectoryPath, err)
+	}
+
+	ctx := context.Background()
+	var exportUrl string
+	diagErr := util.RetryWhen(util.IsStatus404, func() (*platformclientv2.APIResponse, diag.Diagnostics) {
+		resp, err := cp.initiateContactListContactsExport(ctx, contactListId)
+		// Sleep one second before attempting to retrieve export url to give the system time to be able to generate the URL
+		time.Sleep(time.Second)
```
HemanthDogiparthi12, line 208: will this delay be sufficient ?

genesyscloud/validators/validators.go
```diff
@@ -207,6 +210,114 @@ func ValidatePath(i interface{}, k string) (warnings []string, errors []error) {
...
+
+func ValidateCSVFormatWithConfig(filepath string, opts ValidateCSVOptions) error {
+
+	const (
+		maxSkipInterval     = 1000000 // Maximum allowed skip interval
+		defaultSkipInterval = 1000    // Default skip interval
+		maxSampleSize       = 100000  // Maximum allowed sample size
+	)
+	// Validate configuration
+	if opts.SkipInterval < 0 {
```
HemanthDogiparthi12, line 228: can you add a nil check here for opts. just to be sure. we have control of the method at the caller side.

genesyscloud/outbound_contact_list_contact/resource_genesyscloud_outbound_contact_list_contact.go
```diff
@@ -18,30 +17,6 @@ import (
 	"github.com/mypurecloud/platform-client-sdk-go/v150/platformclientv2"
 )
 
-func getAllContacts(ctx context.Context, clientConfig *platformclientv2.Configuration) (resourceExporter.ResourceIDMetaMap, diag.Diagnostics) {
```
HemanthDogiparthi12, line 21: the regression would fail for these , for someone who has resource_genesyscloud_outbound_contact_list_contact  in their export scripts. which forces them to move to the new resource.

genesyscloud/outbound_contact_list_contact/resource_genesyscloud_outbound_contact_list_contact_schema.go
```diff
@@ -66,24 +64,14 @@ var (
 	}
 )
 
-func ContactExporter() *resourceExporter.ResourceExporter {
```
HemanthDogiparthi12, line 69: the regression would fail for these , for someone who has resource_genesyscloud_outbound_contact_list_contact in their export scripts. which forces them to move to the new resource.

genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_schema.go
```diff
@@ -101,7 +106,37 @@ func ResourceOutboundContactList() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
-		SchemaVersion: 1,
+		SchemaVersion: 2,
+		CustomizeDiff: customdiff.All(
+			customdiff.ComputedIf("contacts_file_content_hash", files.FileContentHashChanged("contacts_filepath", "contacts_file_content_hash")),
+			// This function ensures that the contacts file is a CSV file and that it includes the columns defined on the resource
+			func(ctx context.Context, d *schema.ResourceDiff, _ interface{}) error {
```
HemanthDogiparthi12, line 113: can we include the function definition to a different util class, instead of having it in the schema section.

examples/resources/genesyscloud_outbound_contact_list_contact/apis.md
bbbco: That is handled by step 1.

genesyscloud/validators/validators.go
bbbco: `ValidateCSVOptions` is not a pointer, it is a value type. A nil check is unneccesary and actually won't compile because you can't check for nil on a non-pointer struct type. Since `opts` is passed by value, it will always have the default zero values for the struct if nothing is passed in, but it can never be nil.

genesyscloud/outbound_contact_list_contact/resource_genesyscloud_outbound_contact_list_contact_schema.go
bbbco: I tested, and if someone has a `genesyscloud_outbound_contact_list_contact` resource in the filters, it is just ignored.

genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_schema.go
bbbco: Done

genesyscloud/util/files/util_files.go
```diff
@@ -179,49 +182,76 @@ func DownloadOrOpenFile(path string) (io.Reader, *os.File, error) {
 	var reader io.Reader
 	var file *os.File
 
-	_, err := os.Stat(path)
-	if err != nil {
-		_, err = url.ParseRequestURI(path)
-		if err == nil {
-			resp, err := http.Get(path)
```
bbbco, line 186: Had to refactor this function because a non-existent filepath would get to this line and barf with an error complaining about not having a protocol, which was a red herring. The real issue is that the local file didn't exist. I refactored this function so that we check if the path has a protocol and make the call first, then if not, check to ensure the filepath exists. Also added test cases.

genesyscloud/outbound_contact_list_contact/resource_genesyscloud_outbound_contact_list_contact.go
bbbco: I'm not quite sure what you mean here.

bbbco:
@monishapadmavathi confirmed that tests on CI/CD builds are passing for this branch

HemanthDogiparthi12 (approved):

HemanthDogiparthi12 (commented):
approved. please resolve conflicts for merge to dev

bbbco:
Resolved merge conflicts

PR #1510, opened 2025-01-28
[Devtooling-991] Added voicemail timeout to user

fergalmcl:
exports correctly


docs/resources/user.md
```diff
@@ -132,6 +132,7 @@ resource "genesyscloud_user" "example_user" {
 - `routing_utilization` (List of Object) The routing utilization settings for this user. If empty list, the org default settings are used. If not set, this resource will not manage the users's utilization settings. (see [below for nested schema](#nestedatt--routing_utilization))
 - `state` (String) User's state (active | inactive). Default is 'active'. Defaults to `active`.
 - `title` (String) User's title.
+- `voicemail_userpolicies` (Block List, Max: 1) User's voicemail policies. (see [below for nested schema](#nestedblock--voicemail_userpolicies))
```
carnellj-genesys, line 135: Should we name this attribute voicemail_user_policies to be consistent with other attributes.  @HemanthDogiparthi12 @charliecon Looking for your thoughts

genesyscloud/user/resource_genesyscloud_user_utils.go
```diff
@@ -236,6 +241,32 @@ func updateUserProfileSkills(d *schema.ResourceData, proxy *userProxy) diag.Diag
 	return nil
 }
 
+func updateUserVoicemailPolicies(d *schema.ResourceData, proxy *userProxy) diag.Diagnostics {
+	if d.HasChange("voicemail_userpolicies") {
+		if voicemailUserpolicies := d.Get("voicemail_userpolicies").([]interface{}); voicemailUserpolicies != nil {
```
carnellj-genesys, line 246: This is a really deep method with multiple levels of if.

Go developer usually do a straight line where it does an if and then returns and then breaks.  Multiple-level if can be problematic and make the code harder to read

genesyscloud/user/resource_genesyscloud_user_utils.go
```diff
@@ -871,6 +902,27 @@ func getSdkUtilizationTypes() []string {
...
+func buildVoicemailUserpoliciesRequest(voicemailUserpolicyMap map[string]interface{}) platformclientv2.Voicemailuserpolicy {
+	alertTimeoutSeconds := voicemailUserpolicyMap["alert_timeout_seconds"].(int)
+
+	return platformclientv2.Voicemailuserpolicy{
+		AlertTimeoutSeconds: &alertTimeoutSeconds,
+	}
+}
+
+func flattenVoicemailUserpolicies(d *schema.ResourceData, voicemail *platformclientv2.Voicemailuserpolicy) []interface{} {
+	if voicemail == nil {
```
carnellj-genesys, line 914: Nice

genesyscloud/user/resource_genesyscloud_user_schema.go
```diff
@@ -422,9 +439,10 @@ func UserExporter() *resourceExporter.ResourceExporter {
 			"locations.location_id":         {RefType: "genesyscloud_location"},
 		},
 		RemoveIfMissing: map[string][]string{
-			"routing_skills":    {"skill_id"},
-			"routing_languages": {"language_id"},
-			"locations":         {"location_id"},
+			"routing_skills":         {"skill_id"},
+			"routing_languages":      {"language_id"},
+			"locations":              {"location_id"},
+			"voicemail_userpolicies": {"alert_timeout_seconds"},
```
carnellj-genesys, line 445: See my earlier comments about naming conventions

docs/resources/user.md
HemanthDogiparthi12: voicemail_user_policies or voicemail_policies should be good !!

genesyscloud/user/resource_genesyscloud_user_schema.go
```diff
@@ -159,6 +159,15 @@ var (
 			},
 		},
 	}
+	voicemailUserpoliciesResource = &schema.Resource{
```
HemanthDogiparthi12, line 162: there is another property sendEmailNotifications in voicemailUserpoliciesResource. can this also be added to the schema. will probably reduce us a jiira from PS in the future :)

docs/resources/user.md
HemanthDogiparthi12: correcting my statement.
may be /voicemail_userpolicies is approapriate. since the api is "/v2/voicemail/userpolicies/{userId}"

carnellj-genesys (approved):

genesyscloud/user/resource_genesyscloud_user_utils.go
```diff
@@ -871,6 +897,36 @@ func getSdkUtilizationTypes() []string {
 	return types
 }
 
+func buildVoicemailUserpoliciesRequest(voicemailUserpolicies []interface{}) platformclientv2.Voicemailuserpolicy {
+	var request platformclientv2.Voicemailuserpolicy
+	if extractMap, ok := voicemailUserpolicies[0].(map[string]interface{}); ok {
+		alertTimeoutSeconds := extractMap["alert_timeout_seconds"].(int)
```
charliecon, line 903: If you omit just `alert_timeout_seconds` from `voicemail_userpolicies`, you'll see the error "alert timeout too small". This is probably because the zero value for integers is being stored in `alertTimeoutSeconds`, which is zero. If you change this to `alertTimeoutSeconds, ok := ...` , and then only set the value of  `AlertTimeoutSeconds` if ok is true, then that should fix the problem

charliecon (approved):
Just have one new comment. Also there are some merge conflicts that need to be fixed

PR #1512, opened 2025-01-30
NO-JIRA Tech Debt sweep of items that came up during development on other PRs

bbbco:
NO-JIRA Tech Debt sweep
- Replace "path" package with "filepath" package, which should be preferred for several reasons:
  1. filepath automatically uses the correct path separator for the current O/S (Windows, Linux, Mac)
  2. filepath automatically normalizes the paths, handling multiple separators, resolving relative paths, etc
  3. filepath includes some extra helper functions

- Ensures consistency in organizing files in the test directory

- Adds helper functions to retrieve certain directories to make this task more consistent (and removes cruft):
  - Root level dir
  - Test data dir
  - Test temp dir

- Adds some test cases

- Adds the golang platform, which can be used to call the provider during tests. Also added IsDevelopmentPlatform() check

- Improves some test cases


charliecon (approved):
Very nice 👌 The complete move to using filepath instead of hard-coded slashes was much needed

PR #1516, opened 2025-01-31
DEVTOOLING-926: Optional stack trace recovery & logging

charliecon:


bbbco (approved):
Nifty

genesyscloud/provider/provider.go
```diff
@@ -125,9 +125,10 @@ func New(version string, providerResources map[string]*schema.Resource, provider
 					Type:        schema.TypeBool,
 					Optional:    true,
 					DefaultFunc: schema.EnvDefaultFunc(logStackTracesEnvVar, false),
-					Description: fmt.Sprintf(`If set to true the provider will log stack traces to a file instead of crashing, where possible. 
+					Description: fmt.Sprintf(`If true, stack traces will be logged to a file instead of crashing the provider, whenever possible. 
 If the stack trace occurs within the create context and before the ID is set in the schema object, then the command will fail with the message 
-"Root object was present, but now absent." Can be set with the %s environment variable.`, logStackTracesEnvVar),
+"Root object was present, but now absent." Can be set with the %s environment variable. **WARNING**: This is a debugging feature that may cause your Terraform state to become out of sync with the API.
```
bbbco, line 130: Is the risk of becoming out of sync because when a stack trace occurs, an update could be coming through on the API that is different in the state, or vice versa, and the two are not coalesced due to the stack?

charliecon: One example is if a stack trace occurs in the delete context before the actual delete API call happens; the traces will be written to the log file, the resource will be removed from the state file, the apply will complete successfully, but the resource will still exist in the API. 

Also, if the stack trace happens before every resource field is read back into the state during the read context, then some resource fields will be nil locally even though they have a value in the API.

carnellj-genesys:
@charliecon 

1.  First of all nice job.
2.  Did we run Q against these changes to search for Nil pointer exceptions.
3. Need to get the file conflicts resolved.  :)

charliecon:
> @charliecon
> 
> 1. First of all nice job.
> 2. Did we run Q against these changes to search for Nil pointer exceptions.
> 3. Need to get the file conflicts resolved.  :)

@carnellj-genesys Amazon Q scan done & conflicts resolved 👍

PR #1526, opened 2025-02-05
Feat/devtooling 823 - Add description to the Genesys Cloud routing wrapup code resource

shrutisuryawanshigenesys:


genesyscloud/routing_wrapupcode/resource_genesyscloud_routing_wrapupcode_test.go
```diff
@@ -54,6 +57,7 @@ func TestAccResourceRoutingWrapupcode(t *testing.T) {
 					codeResourceLabel1,
 					codeName2,
 					"genesyscloud_auth_division."+divResourceLabel+".id",
+					description,
 				),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(ResourceType+"."+codeResourceLabel1, "name", codeName2),
```
charliecon, line 63: We should check the value we here also after create and updating description

charliecon (approved):

charliecon (approved):
Based on the workflow failure, you need to run `make docs`. Also there is one merge conflict to fix. Then we're good to merge

carnellj-genesys (approved):

PR #1529, opened 2025-02-06
Fixing DEVTOOLING-1070 where destroy can fail on read timeout

bbbco:
Fixes DEVTOOLING-1070 where a destroy can fail on read timeout

### DIFF
--- a/genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone.go
+++ b/genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone.go
@@ -90,6 +90,7 @@ func readPhone(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 		}
 
 		if currentPhone.Site != nil && currentPhone.Site.Id != nil {
+			_ = d.Set("site_id", *currentPhone.Site.Id)
 			log.Printf("Phone ID = %s and the site_id = %s", d.Id(), *currentPhone.Site.Id)
 		} else {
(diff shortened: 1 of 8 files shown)


genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone.go
```diff
@@ -91,6 +91,7 @@ func readPhone(ctx context.Context, d *schema.ResourceData, meta interface{}) di
 
 		if currentPhone.Site != nil && currentPhone.Site.Id != nil {
 			log.Printf("Phone ID = %s and the site_id = %s", d.Id(), *currentPhone.Site.Id)
+			_ = d.Set("site_id", *currentPhone.Site.Id)
```
bbbco, line 94: This was inadvertently removed in a prior PR. cc @sureshperiyappan

genesyscloud/util/util_retries.go
```diff
@@ -37,6 +37,7 @@ func WithRetriesForReadCustomTimeout(ctx context.Context, timeout time.Duration,
 		if strings.Contains(fmt.Sprintf("%v", err), "API Error: 404") {
 			// Set ID empty if the object isn't found after the specified timeout
 			d.SetId("")
+			return nil
```
bbbco, line 40: Ultimately, this is where the true fix for the issue lives. We don't throw an error if the resource cannot be found. Instead, we just set the ID to empty after the timeout and immediately return.

carnellj-genesys (approved):

charliecon (approved):

PR #1530, opened 2025-02-06
DEVTOOLING-1058: Resolving references whose key in its own resource map is a concatenation of IDs

charliecon:
I left comments to highlight the relevant changes. Everything else is just refactoring based on suggestions from the linter

### DIFF
--- a/genesyscloud/knowledge_document_variation/genesyscloud_knowledge_document_variation_proxy.go
+++ b/genesyscloud/knowledge_document_variation/genesyscloud_knowledge_document_variation_proxy.go
@@ -180,7 +180,7 @@ func getVariationRequestIdByNameFn(ctx context.Context, p *variationRequestProxy
 		allVariations = append(allVariations, *allDraftVariations...)
 	}
 
-	if allVariations == nil || len(allVariations) == 0 {
+	if len(allVariations) == 0 {
 		return "", resp, true, err
 	}
 
--- a/genesyscloud/knowledge_document_variation/genesyscloud_knowledge_document_variation_proxy.go
+++ b/genesyscloud/knowledge_document_variation/genesyscloud_knowledge_document_variation_proxy.go
@@ -21,6 +21,7 @@ type deleteVariationRequestFunc func(ctx context.Context, p *variationRequestPro
 type getVariationRequestIdByNameFunc func(ctx context.Context, p *variationRequestProxy, name, knowledgeBaseID, knowledgeDocumentID string) (string, *platformclientv2.APIResponse, bool, error)
 type createKnowledgeKnowledgebaseDocumentVersionsFunc func(ctx context.Context, p *variationRequestProxy, knowledgeDocumentId, knowledgeBaseId string, version *platformclientv2.Knowledgedocumentversion) (*platformclientv2.Knowledgedocumentversion, *platformclientv2.APIResponse, error)
 type GetAllKnowledgebaseEntitiesFunc func(ctx context.Context, p *variationRequestProxy, published bool) (*[]platformclientv2.Knowledgebase, *platformclientv2.APIResponse, error)
+type getLatestPublishedOrDraftVariationFunc func(ctx context.Context, p *variationRequestProxy, ids *resourceIDs) (variation *platformclientv2.Documentvariationresponse, response *platformclientv2.APIResponse, err error)
 
 // variationRequestProxy contains all of the methods that call genesys cloud APIs.
 type variationRequestProxy struct {
@@ -34,6 +35,7 @@ type variationRequestProxy struct {
 	deleteVariationRequestAttr                       deleteVariationRequestFunc
 	createKnowledgeKnowledgebaseDocumentVersionsAttr createKnowledgeKnowledgebaseDocumentVersionsFunc
 	GetAllKnowledgebaseEntitiesAttr                  GetAllKnowledgebaseEntitiesFunc
+	getLatestPublishedOrDraftVariationAttr           getLatestPublishedOrDraftVariationFunc
 	variationCache                                   rc.CacheInterface[platformclientv2.Documentvariationresponse]
 }
 
@@ -52,6 +54,7 @@ func newVariationRequestProxy(clientConfig *platformclientv2.Configuration) *var
 		getVariationRequestIdByNameAttr:                  getVariationRequestIdByNameFn,
 		createKnowledgeKnowledgebaseDocumentVersionsAttr: createKnowledgeKnowledgebaseDocumentVersionsFn,
 		GetAllKnowledgebaseEntitiesAttr:                  getAllKnowledgebaseEntitiesFn,
+		getLatestPublishedOrDraftVariationAttr:           getLatestPublishedOrDraftVariationFn,
 		variationCache:                                   variationCache,
 	}
 }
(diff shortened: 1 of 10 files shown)


genesyscloud/resource_exporter/resource_exporter.go
```diff
@@ -45,6 +45,14 @@ type RefAttrSettings struct {
...
 	AltValues []string
+
+	// If the keys in ResourceIDMetaMap of RefType are concatenations of IDs
+	// If true, we will validate that the GUID exists within this key, rather than perform a direct comparison.
+	// E.g.genesyscloud_knowledge_document_variation.knowledge_document_id is a GUID, but the keys in the document exporter
+	// are combinations of the document and parent knowledge base IDs, so to resolve we must see if the key contains the document ID.
+	// Note: This field should only be used if the GUID of the resource we want to resolve is reliably unique i.e. cannot be configured
+	// by the user.
+	IsReferenceIDConcatenation bool
 }
```
charliecon, line 57: ⬆️

genesyscloud/knowledge_document_variation/resource_genesyscloud_knowledge_document_variation_schema.go
```diff
@@ -533,8 +533,11 @@ func KnowledgeDocumentVariationExporter() *resourceExporter.ResourceExporter {
...
 		GetResourcesFunc: provider.GetAllWithPooledClient(getAllKnowledgeDocumentVariations),
 		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
-			"knowledge_base_id":     {RefType: "genesyscloud_knowledge_knowledgebase"},
-			"knowledge_document_id": {RefType: "genesyscloud_knowledge_document"},
+			"knowledge_base_id": {RefType: "genesyscloud_knowledge_knowledgebase"},
+			"knowledge_document_id": {
+				RefType:                    "genesyscloud_knowledge_document",
+				IsReferenceIDConcatenation: true,
+			},
 		},
```
charliecon, line 541: ⬆️

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -1589,8 +1589,17 @@ func (g *GenesysCloudResourceExporter) resolveReference(refSettings *resourceExp
...
+				for k, m := range idMetaMap {
+					if strings.Contains(k, refID) {
+						meta = m
+					}
+				}
+			} else {
+				meta = idMetaMap[refID]
+			}
+			if meta != nil && meta.BlockLabel != "" {
 				if g.isDataSource(refSettings.RefType, meta.BlockLabel, meta.OriginalLabel) && g.resourceIdExists(refID, nil) {
```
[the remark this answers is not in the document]
bbbco: What happens with multiple layers of ids?
For example, knowlege document variation is really a knowlege base + knowledge document + knowledge variation, right? So, let's say the exporter comes across a document id. This logic will iterate over the id map and come to the first instance of a document id matching the guid, BUT, it could also return the first instance of the document variation because the guid would be inside that id.

I think a better option is to have a function on the exporter that allows each resource to determine how to return the GUID (if necessary). In this case, the function can be as simple as:

```
"knowledge_document_id": {
				RefType:                    "genesyscloud_knowledge_document",
				ReferenceIDFunc: func(id) string {
                                     return parseResourceIDs(id).variationID
                                 }
			},
```

genesyscloud/knowledge_document_variation/resource_genesyscloud_knowledge_document_variation.go
```diff
@@ -232,7 +232,7 @@ func readKnowledgeDocumentVariation(ctx context.Context, d *schema.ResourceData,
 		d.SetId(newId)
 
 		_ = d.Set("knowledge_base_id", *knowledgeDocumentVariation.Document.KnowledgeBase.Id)
-		_ = d.Set("knowledge_document_id", ids.documentID)
+		_ = d.Set("knowledge_document_id", *knowledgeDocumentVariation.Document.Id)
```
carnellj-genesys, line 235: Do we need to do some nil checks here.  Are we at risk of of ever having a nil pointer exception

genesyscloud/resource_exporter/resource_exporter.go
carnellj-genesys: Nice.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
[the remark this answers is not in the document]
carnellj-genesys: @charliecon This actually looks a pretty good idea.  That we if we have to deal with multiple ids we could override the function.  In the mean-time for a standard mechanism is id a + id b we could have a default function.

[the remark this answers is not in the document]
charliecon: @bbbco I don't think that can occur because this field is referencing a document, so we're searching through the document exporter

```go
if idMetaMap := exporters[refSettings.RefType].SanitizedResourceMap; idMetaMap != nil {
```

And if you are searching through the ID map from the documents exporter, a document ID will only appear once 

Also, this logic is only being invoked if the bool `IsReferenceIDConcatenation` is true i.e. only to resolve the field `knowledge_document_id`. Otherwise the previous logic is invoked. 

Regardless, I prefer your suggestion of providing a custom function — it definitely simplifies things. Will implement now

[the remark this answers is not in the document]
bbbco: Oh, right. We're only checking the one specific resource type. Good deal.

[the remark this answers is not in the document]
charliecon: @bbbco @carnellj-genesys turns out none of this was necessary and was actually an attempt to work around a bug. 

If anyone is curious, this was the actual bug: 

Variation exporter map is built this way:

```go
id := buildVariationId(*knowledgeBase.Id, *knowledgeDoc.Id, *knowledgeDocumentVariation.Id)
// ...
resources[id] = &resourceExporter.ResourceMeta{BlockLabel: blockLabel}
```

Document exporter map was built this way:

```go
id := fmt.Sprintf("%s,%s", *knowledgeDocument.Id, *knowledgeDocument.KnowledgeBase.Id)
resources[id] = &resourceExporter.ResourceMeta{BlockLabel: *knowledgeBase.Name + "_" + *knowledgeDocument.Title}
```

So during export when we tried to parse apart the id in the variation read context, we weren't setting `knowledge_document_id` to the concatenation of strings; we were setting it to the document ID alone (which is how it is stored in the document export resource map.) And then of course the exporter couldn't resolve the reference to another resource.

genesyscloud/knowledge_document_variation/resource_genesyscloud_knowledge_document_variation.go
charliecon: done

bbbco (approved):

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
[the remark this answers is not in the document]
bbbco: Nice find. Good job digging into the issue!

genesyscloud/knowledge_document_variation/genesyscloud_knowledge_document_variation_proxy.go
```diff
@@ -101,13 +105,62 @@ func (p *variationRequestProxy) GetAllKnowledgebaseEntities(ctx context.Context,
...
+}
+
+func getLatestPublishedOrDraftVariationFn(ctx context.Context, p *variationRequestProxy, ids *resourceIDs) (_ *platformclientv2.Documentvariationresponse, _ *platformclientv2.APIResponse, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("getLatestPublishedOrDraftVariationFn: %w", err)
+		}
+	}()
+
+	time.Sleep(2 * time.Second) // wait for the variation to be published/draft
```
carnellj-genesys, line 128: Rather then just doing a flat out sleep.  Should this be a retry with multiple retries before it dies.

charliecon: done

carnellj-genesys (approved):

PR #1534, opened 2025-02-10
[Devtooling-960] outbound_contact_list exporting incorrectly

fergalmcl:
In the ticket for this [960], I think the problem is caused by a difference with the terraform schema and the api requirements. The 'max_length' field is required in terraform, but when calling the API its only required when 'TEXT' is set as the data type. For 'NUMERIC' types, it is 'max' and 'min' that are required. Since terraform requires 'max_length', exporting a NUMERIC type will cause an error as 'max_length' might not have been set. Setting 'max_length' to optional fixes the issue, but the other option is some validation logic when creating and reading to reflect the API.

### DIFF
--- a/docs/resources/outbound_contact_list.md
+++ b/docs/resources/outbound_contact_list.md
@@ -65,12 +65,12 @@ resource "genesyscloud_outbound_contact_list" "contact-list" {
 Required:
 
 - `column_name` (String) The column name of a column selected for dynamic queueing.
-- `max_length` (Number) The maximum length of the text column selected for dynamic queueing.
 
 Optional:
 
 - `column_data_type` (String) The data type of the column selected for dynamic queueing (TEXT, NUMERIC or TIMESTAMP)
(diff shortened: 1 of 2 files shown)


carnellj-genesys (approved):

bbbco (approved):
I bet you could add some extra validation with the CustomizeDiff validation to ensure that `max` and `min` are set when the type is NUMERIC and `max_length` when type is TEXT to align with the API requirements.

charliecon (approved):

charliecon:
> I bet you could add some extra validation with the CustomizeDiff validation to ensure that `max` and `min` are set when the type is NUMERIC and `max_length` when type is TEXT to align with the API requirements.

@bbbco I'm going to go ahead and merge, and we can come back to that in the future if necessary. But I do find that reflecting API rules in CX as Code can result in forum posts & issues later on if the API rules change

PR #1542, opened 2025-02-10
fix platform dependency for regression

HemanthDogiparthi12:


kavinbalagen (approved):

carnellj-genesys (approved):

charliecon (approved):

genesyscloud/platform/platform.go
```diff
@@ -123,7 +123,7 @@ func init() {
 	path, err := detectExecutingBinary()
 	if err != nil {
 		log.Printf(`Error detecting binary: %v`, err)
-		platformConfigSingleton.platform = PlatformUnknown
+		platformConfigSingleton.platform = PlatformTerraform // use default terraform binary. so it wont break regression
```
bbbco, line 126: This is a sanity check to ensure that a valid platform is in use. I agree we need to default to the Terraform platform binary, which we can do in `Binary()`

genesyscloud/tfexporter/tfstate_exporter.go
```diff
@@ -132,6 +139,57 @@ func (t *TFStateFileWriter) writeTfState() diag.Diagnostics {
 			%s`, err, cliErrorPostscript)
 		log.Print(cliErrorPostscript)
 		// Don't fail everything even if this errors.
+		err := t.writeTfStateLegacy(stateFilePath)
```
bbbco, line 142: There is already handling to output the command that the user should run manually if there is an issue with the platform detection or the command running. We don't fail, we just log the error message and what the user should do. There is no need for the extra `writeTfStateLegacy()`

bbbco:
Hey @HemanthDogiparthi12 thanks for working on this while I was out. After discussing with you, I agree we need to add some more graceful handling if platform detection doesn't work. I've pushed a fix for the platform detection (`gopsutils` library needed updated) and incorporated your suggested logs and graceful handling in this PR: https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1545

bbbco (changes requested):
Hey @HemanthDogiparthi12 thanks for working on this while I was out. After discussing with you, I agree we need to add some more graceful handling if platform detection doesn't work. I've pushed a fix for the platform detection (`gopsutils` library needed updated) and incorporated your suggested logs and graceful handling in this PR: https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1545

genesyscloud/tfexporter/tfstate_exporter.go
HemanthDogiparthi12: even in the future, when we intend to have exporter as a service, it would be nice to have tf state dumped with little/no manual intervention.

HemanthDogiparthi12: I mean , if platform validate fails. we can run back up to prev way of writing down the tfstate.

HemanthDogiparthi12:
closing this PR.

genesyscloud/tfexporter/tfstate_exporter.go
HemanthDogiparthi12: image

@bbbco this is the problem I am trying to mention.  https://inindca.atlassian.net/browse/DEVTOOLING-1072 just saw a ticket is already created for the same reason. tfstate will not be in v3 if terraform path is not identified. and customer has to do a manual intervention trying to convert it to desired tfstate format.

bbbco: @HemanthDogiparthi12 yes, the full message you highlighted already indicates what the user has to do:
```
2025/02/07 15:18:06 Failed to run the terraform CLI to upgrade the generated state file.\n\tThe generated tfstate file will need to be upgraded manually by running the\n\tfollowing in the state file's directory:\n\t'terraform state replace-provider registry.terraform.io/-/genesyscloud registry.terraform.io/mypurecloud/genesyscloud'"
```
If the output is not in JSON then it should render nicer for the user to see.

HemanthDogiparthi12: this would just create one more manual intervention for the customer to  troubleshoot the logs and do the required state upgrade. would likely anticipate a bunch of tickets on similar lines. I would still feel it is better we write it out ourselves in the code during a error scenario.

HemanthDogiparthi12: i think the problem is not running terraform command, but identifying terraform path via the platform . as the platform command shows "failed to run terraform path". which is similar to provider crash. the regression which runs the "terraform" (writeTfStateLegacy) without fetching from platform os exec, works for all the existing users incase if it fails to identify the path in an errored scenario. we can well put it in error log level even if the writeTfStateLegacy too fails.

bbbco: Ok, how about a slightly more robust way of detecting the terraform executable that uses your original legacy method? https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1567

charliecon (approved):

genesyscloud/tfexporter/tfstate_exporter.go
charliecon: I agree with Hemanth that its better to try run the command ourselves this way before failing. Let's merge these changes for now to get them out the door and look into #1567 after

bbbco: I agree with the principle that we should try to run the command ourselves.

However, the code that is in this PR duplicates the same actions that are already being taken. The reason the failures occur is because the platform exe can't always be found in the initial platform implementation. https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1567 hardens this so that the platform will (almost) always be returned so that the command can be run. This way the platform is detected or assumed and the command is always run, and we're not duplicating code or actions.

PR #1551, opened 2025-02-12
Feature/devtooling 931

kavinbalagen:


genesyscloud/organization_authentication_settings/genesyscloud_organization_authentication_settings_proxy.go
```diff
@@ -58,6 +75,22 @@ func (p *orgAuthSettingsProxy) updateOrgAuthSettings(ctx context.Context, orgAut
 	return p.updateOrgAuthSettingsAttr(ctx, p, orgAuthSettings)
 }
 
+func getTokensTimeOutSettingsFn(ctx context.Context, p *orgAuthSettingsProxy) (*platformclientv2.Idletokentimeout, *platformclientv2.APIResponse, error) {
```
carnellj-genesys, line 78: Did you run Q against this?  Also we should doc it and if there is no unit tests include a unit test for these two functions

kavinbalagen: image
yes john i have ran Q over these changes and have added more unit tests and documentation

PR #1565, opened 2025-02-21
Feature/devtooling 985 1

HemanthDogiparthi12:


genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -173,6 +173,12 @@ func NewGenesysCloudResourceExporter(ctx context.Context, d *schema.ResourceData
 	return gre, nil
 }
 
+func identifyExportFormat(d *schema.ResourceData) string {
+	if d.Get("export_as_hcl").(bool) {
+		return formatHCL
```
bbbco, line 178: This could be set to true or false. You should return as `formatHCL` if true, `formatJson` if not.

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -99,6 +99,13 @@ func ResourceTfExport() *schema.Resource {
 				Default:     false,
 				ForceNew:    true,
 			},
+			"export_as_hcl": {
+				Description: "Export the config as HCL. Deprecated use export_format instead ",
```
bbbco, line 103: Description update suggested: `... Deprecated. Please use the export_format attribute instead`

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -99,6 +99,13 @@ func ResourceTfExport() *schema.Resource {
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
```
bbbco, line 106: Maybe remove the default in deference to the default set by `export_format`. We only really care if this attribute is actually set.

bbbco (changes requested):
I identified a few items @HemanthDogiparthi12

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
HemanthDogiparthi12: d.Get("export_as_hcl").(bool) returns true for hcl and false else.
the default is json for export_format. we should be good here.

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
HemanthDogiparthi12: done

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
bbbco: We're not accounting for the case where a user has configured `export_as_hcl = false` (to export as JSON).

I think this should be updated:
```
exportAsHcl := d.Get("export_as_hcl")
if exportAsHcl != nil {
  if exportAsHcl.(bool) == true {
    return formatHCL
  } 
  return formatJson
}
```

bbbco (approved):

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
HemanthDogiparthi12: strings.ToLower(d.Get("export_format").(string)) returns "json" since the default is json for export_format when it is not used.  will add a conflicts_with just to be sure only one of these is configured.

docs/resources/tf_export.md
```diff
@@ -65,6 +65,7 @@ resource "genesyscloud_tf_export" "export" {
 - `enable_dependency_resolution` (Boolean) Adds a "depends_on" attribute to genesyscloud_flow resources with a list of resources that are referenced inside the flow configuration . This also resolves and exports all the dependent resources for any given resource. Resources mentioned in exclude_attributes will not be exported. Defaults to `false`.
 - `exclude_attributes` (List of String) Attributes to exclude from the config when exporting resources. Each value should be of the form {resource_type}.{attribute}, e.g. 'genesyscloud_user.skills'. Excluded attributes must be optional.
 - `exclude_filter_resources` (List of String) Exclude resources that match either a resource type or a resource type::regular expression.  See export guide for additional information.
+- `export_as_hcl` (Boolean) Export the config as HCL. Deprecated. Please use the export_format attribute instead Defaults to `false`.
```
bbbco, line 68: Missing a period after instead.

kavinbalagen (approved):

PR #1566, opened 2025-02-21
Fix routing problem

HemanthDogiparthi12:


genesyscloud/routing_email_route/genesyscloud_routing_email_route_proxy.go
```diff
@@ -207,8 +207,8 @@ func getRoutingEmailRouteIdByPatternFn(ctx context.Context, p *routingEmailRoute
 
 	for _, inboundRoutes := range *inboundRoutesMap {
 		for _, inboundRoute := range inboundRoutes {
-			if *inboundRoute.Pattern == pattern {
-				log.Printf("Retrieved the routing email route id %s by pattern %s", *inboundRoute.Id, pattern)
+			if *inboundRoute.Pattern == pattern && *inboundRoute.FromName == fromName {
```
sureshperiyappan, line 210: I tested the fix you provided in the PR, but it’s not working. The issue is that fromName can be duplicated across multiple domains. We have to handle this at domain level we have to filter domain specific.

PR #1572, opened 2025-02-25
No jira: fix email route test case

charliecon:


genesyscloud/routing_email_route/data_source_genesyscloud_routing_email_route_test.go
```diff
@@ -68,7 +68,7 @@ func generateRoutingEmailRouteDataSource(
 	return fmt.Sprintf(`
 		data "genesyscloud_routing_email_route" "%s" {
 			pattern = "%s"
-			domain_id = "%s"
+			domain_id = %s
```
charliecon, line 71: This is what fixed the data source test case

genesyscloud/routing_email_route/genesyscloud_routing_email_route_proxy.go
```diff
@@ -92,73 +92,65 @@ func (p *routingEmailRouteProxy) deleteRoutingEmailRoute(ctx context.Context, do
...
-				return nil, apiResponse, fmt.Errorf("Failed to get routing email route: %s", err)
+				return nil, apiResponse, fmt.Errorf("failed to get routing email route: %s", err.Error())
 			}
 			if routes.Entities == nil || len(*routes.Entities) == 0 {
 				break
 			}
 			allDomainRoutes = append(allDomainRoutes, *routes.Entities...)
-			allInboundRoutes[*domain.Id] = allDomainRoutes
 		}
+		allInboundRoutes[*domain.Id] = allDomainRoutes
```
charliecon, line 114: Moving this to here means the value at key `*domain.Id` won't be overwritten on each page of results

genesyscloud/routing_email_route/genesyscloud_routing_email_route_proxy.go
```diff
@@ -92,73 +92,65 @@ func (p *routingEmailRouteProxy) deleteRoutingEmailRoute(ctx context.Context, do
...
-		}
-
-		routes, _, err := getAllRoutingEmailRouteByDomainIdFn(ctx, p, domains, name)
-		if err != nil {
-			return nil, resp, fmt.Errorf("Failed to get routing email domains: %s", err)
-		}
-		mergeIntoExistingMap(allInboundRoutes, *routes)
-	}
-	log.Printf("Returning routes for domains: %v", reflect.ValueOf(allInboundRoutes).MapKeys())
-	return &allInboundRoutes, apiResponse, nil
```
charliecon, line 161: This was unnecessary because it just did the same thing accept with `""` instead of `domainId` , and only if domainId was "". But if domainId was "", then we were doing the same thing twice; once with the empty string variable and again with a hard-coded empty string.

HemanthDogiparthi12 (approved):

genesyscloud/routing_email_route/genesyscloud_routing_email_route_proxy.go
```diff
@@ -92,73 +92,65 @@ func (p *routingEmailRouteProxy) deleteRoutingEmailRoute(ctx context.Context, do
...
+		if domains.Entities == nil || len(*domains.Entities) == 0 {
+			break
+		}
+		allDomains = append(allDomains, *domains.Entities...)
 	}
-	if domains.Entities == nil || len(*domains.Entities) == 0 {
-		return &allInboundRoutes, resp, nil
+
+	if len(allDomains) == 0 {
+		return nil, apiResponse, nil
```
bbbco, line 138: Because we now return `nil` here, that means any calling functions need to also check that this is not nil before processing the records. The `readRoutingEmailRoute()` and `getAllRoutingEmailRoutes()` functions both make calls to this function without checking that the inboundRoutes result is null.

Thus we get https://inindca.atlassian.net/browse/DEVTOOLING-1099

PR #1577, opened 2025-02-27
Feat/gpe 14858 update segment contract

matthewhallinan-gen:
As part of the in-progress Customer Segmentation feature, Journey Segments can be directly assigned to an external contact (representing the end user), rather than the user's session. This allows Segments to live beyond the lifetime of the session. A new field is being added to the Segment, for admins to configure how long these Segment assignments to a contact should last. This PR adds this new field to the CX as Code repo.

### DIFF
--- a/genesyscloud/journey_segment/resource_genesyscloud_journey_segment_schema.go
+++ b/genesyscloud/journey_segment/resource_genesyscloud_journey_segment_schema.go
@@ -72,7 +72,7 @@ var (
 			MaxItems:    1,
 			Elem:        journeyResource,
 		},
-		"assignmentExpirationDays": {
+		"assignment_expiration_days": {
 			Description:  "Time, in days, from when the segment is assigned until it is automatically unassigned.",
 			Type:         schema.TypeInt,
 			Optional:     true,
--- a/genesyscloud/journey_segment/resource_genesyscloud_journey_segment_schema.go
+++ b/genesyscloud/journey_segment/resource_genesyscloud_journey_segment_schema.go
@@ -52,6 +52,7 @@ var (
 			Required:     true,
 			ForceNew:     true, // scope can be only set during creation
 			ValidateFunc: validation.StringInSlice([]string{"Session"}, false),
+			Deprecated:   "The scope field is redundant and will be removed in a future release.",
 		},
 		"should_display_to_agent": {
 			Description: "Whether or not the segment should be displayed to agent/supervisor users.",
--- a/genesyscloud/journey_segment/resource_genesyscloud_journey_segment_schema.go
+++ b/genesyscloud/journey_segment/resource_genesyscloud_journey_segment_schema.go
@@ -72,6 +72,12 @@ var (
 			MaxItems:    1,
 			Elem:        journeyResource,
 		},
+		"assignmentExpirationDays": {
+			Description:  "Time, in days, from when the segment is assigned until it is automatically unassigned.",
+			Type:         schema.TypeInt,
+			Optional:     true,
+			ValidateFunc: validation.IntAtLeast(0),
+		},
 	}
 
 	contextResource = &schema.Resource{
--- a/docs/resources/journey_segment.md
+++ b/docs/resources/journey_segment.md
@@ -64,6 +64,7 @@ resource "genesyscloud_journey_segment" "example_journey_segment_resource" {
 
 ### Optional
 
+- `assignment_expiration_days` (Number) Time, in days, from when the segment is assigned until it is automatically unassigned.
 - `context` (Block Set, Max: 1) The context of the segment. (see [below for nested schema](#nestedblock--context))
 - `description` (String) A description of the segment.
 - `is_active` (Boolean) Whether or not the segment is active. Defaults to `true`.
--- a/docs/resources/journey_segment.md
+++ b/docs/resources/journey_segment.md
@@ -60,7 +60,7 @@ resource "genesyscloud_journey_segment" "example_journey_segment_resource" {
 
 - `color` (String) The hexadecimal color value of the segment.
 - `display_name` (String) The display name of the segment.
-- `scope` (String) The target entity that a segment applies to. Valid values: Session
+- `scope` (String, Deprecated) The target entity that a segment applies to. Valid values: Session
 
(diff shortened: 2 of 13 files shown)


matthewhallinan-gen:
Few things I noticed while working on the PR:

1. Scope is being removed from the public API as a request body field (see [this ticket](https://inindca.atlassian.net/browse/GPE-15227)). It's already removed from the backend service as of [this ticket](https://inindca.atlassian.net/browse/GPE-15226). I can block this ticket with the linked Pub API ticket and remove it while in the neighbourhood, but it’s probably better to make a separate ticket for it.

Should removing it here block its removal from Pub API?

2. The test structure confused me. 2 of the test cases are basic_attributes and optional_attributes - see [here](https://github.com/MyPureCloud/terraform-provider-genesyscloud/tree/main/test/data/resource/genesyscloud_journey_segment). The only thing being tested under optional attributes is description. However, under [the Segment docs](https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/docs/resources/journey_segment.md), colour, display_name and scope are the only ones listed as required; everything else is optional. That in turn doesn’t align with Pub API, where there are 4 required fields: colour, displayName, journey and context (see pic below). 
Considering Scope and shouldDisplayToAgent are optional and have defaults (and therefore should probably be tested under the optional tests in CX as Code), I’ve added assignmentExpirationDays into all tests where they’re added. So, the optional attributes test is still only testing description. But maybe all of this should be tidied up, so that the optional tests only test the fields which are optional as per Pub API?
Again, probably something for a separate ticket?

Screenshot 2025-02-27 at 15 26 05

3. Confusing point 2 further, isActive is present on the Segment schema but it doesn’t seem to be tested at all. Maybe also a separate ticket? Does it need adding to the testing? I guess it does.

charliecon:
@matthewhallinan-gen Apologies about the delay. We will get to this ASAP. Also thanks for all the detail 👌

matthewhallinan-gen:
> @matthewhallinan-gen Apologies about the delay. We will get to this ASAP. Also thanks for all the detail 👌

Thanks a million @charliecon! The main thing is point 1, whether I can go ahead and merge the Pub API PR. This PR is effectively soft blocking that one till I know. Even if the review & points 2 & 3 are addressed later, it would be great to get that part cleared up :)

charliecon:
Hi @matthewhallinan-gen 

1. In this case, the process can just include adding a Deprecation notice to the schema definition for this field and letting the user know it is redundant and will be removed in a future CX as Code release. That way, they should remove it manually themselves instead of their Terraform config suddenly breaking on version upgrade. Once the deprecation notice is added, you can remove any logic that handles the field in the CRUD functions. And no — changes in CX as Code should not block Public API changes. 
2.  The tests definitely made not be perfect. The important thing is that they confirm the Create context sets those values properly, and then that update does also (for fields that can be updated), and that the delete works. And that all fields are covered. 
3. If it's not tested, then we should fix that. 

I agree that points 2 & 3 can be separate tickets. You can create the ticket under `DEVTOOLING` with the component `CX as Code` and leave it in our backlog. You can also set it to low priority since we don't suspect that there is necessarily anything wrong with the fields.  Thanks!

charliecon (approved):
Looks good! Is there a feature toggle we need to enable to get these tests passing? At the moment I'm seeing `API Error: 501 - Assignment expiration is not currently supported.`

matthewhallinan-gen:
Cheers @charliecon!

I unblocked the Pub API PR. And when you say adding a deprecation notice to the schema: that's just adding 
			`Deprecated:   "The scope field is redundant and will be removed in a future release.",`
to scope, in `resource_genesyscloud_journey_segment_schema.go`, right? I've done this locally, but I also see [a past PR where it's added somewhere else](https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1460/files), so I wanted to double check before pushing my commit.

I made a ticket here for that testing: https://inindca.atlassian.net/browse/DEVTOOLING-1153 Since it's all related to the structure of testing Segment optional vs mandatory fields I put it in one ticket.

Yep, there is an FT, it's PURE-5834. I don't need to reference that in the repo though right? It's just for your test org?

Also - who merges, you or me? If it's me, do I need to wait for any other approvers?

charliecon:
> Cheers @charliecon!
> 
> I unblocked the Pub API PR. And when you say adding a deprecation notice to the schema: that's just adding `Deprecated: "The scope field is redundant and will be removed in a future release.",` to scope, in `resource_genesyscloud_journey_segment_schema.go`, right? I've done this locally, but I also see [a past PR where it's added somewhere else](https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1460/files), so I wanted to double check before pushing my commit.
> 
> I made a ticket here for that testing: https://inindca.atlassian.net/browse/DEVTOOLING-1153 Since it's all related to the structure of testing Segment optional vs mandatory fields I put it in one ticket.
> 
> Yep, there is an FT, it's PURE-5834. I don't need to reference that in the repo though right? It's just for your test org?
> 
> Also - who merges, you or me? If it's me, do I need to wait for any other approvers?

Sounds good. You can go ahead and push that change 

We just need the feature toggle to get the tests passing in jenkins. No need to make any more changes here 

I'll merge after you push the deprecation stuff 👍

genesyscloud/journey_segment/resource_genesyscloud_journey_segment_schema.go
```diff
@@ -52,6 +52,7 @@ var (
 			Required:     true,
 			ForceNew:     true, // scope can be only set during creation
 			ValidateFunc: validation.StringInSlice([]string{"Session"}, false),
+			Deprecated:   "The scope field is redundant and will be removed in a future release.",
```
charliecon, line 55: One more small thing - I'm pretty sure you'll have to re-run `make docs` to render this change

charliecon:
@matthewhallinan-gen Merged 👍 You'll see these changes released next week with v1.62.0. Thanks for the contribution!

matthewhallinan-gen:
> @matthewhallinan-gen Merged 👍 You'll see these changes released next week with v1.62.0. Thanks for the contribution!

Thanks a million for the help Charlie 👍

PR #1585, opened 2025-03-03
DEVTOOLING-1099: fix email_route nil pointer dereference + refactorings

charliecon:


bbbco (approved):
ty @charliecon !

bbbco:
@charliecon It does seem like we should probably update a test case to account for this particular edge case

charliecon:
> @charliecon It does seem like we should probably update a test case to account for this particular edge case

done 👍

kavinbalagen (approved):

PR #1607, opened 2025-03-13
Using archy export service in architect flow exporter 

charliecon:


docs/resources/tf_export.md
```diff
@@ -75,6 +75,7 @@ resource "genesyscloud_tf_export" "export" {
 - `replace_with_datasource` (List of String) Include only resources that match either a resource type or a resource type::regular expression.  See export guide for additional information.
 - `resource_types` (List of String, Deprecated) Resource types to export, e.g. 'genesyscloud_user'. Defaults to all exportable types. NOTE: This field is deprecated and will be removed in future release.  Please use the include_filter_resources or exclude_filter_resources attribute.
 - `split_files_by_resource` (Boolean) Split export files by resource type. This will also split the terraform provider and variable declarations into their own files. Defaults to `false`.
+- `use_legacy_architect_flow_exporter` (Boolean) When set to `false`, architect flow configuration files will be downloaded as part of the flow export process. Defaults to `true`.
```
bbbco, line 78: So we're planning to retain the legacy implementation as default?

charliecon: Yes, that's the plan for now to avoid breaking existing systems with the version upgrade. Later we can emit a notice and then make `false` the default or even remove this field entirely and ditch the legacy method. 

cc @carnellj-genesys in case I'm misremembering anything about our discussions

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_schema.go
```diff
@@ -34,6 +37,20 @@ func ArchitectFlowExporter() *resourceExporter.ResourceExporter {
...
+			SubDirectory:              exportSubDirectoryName,
+		},
+	}
+
+	resourceExporter.SetNewFlowResourceExporter(newExporter)
+
+	return legacyExporter
 }
 
 func ResourceArchitectFlow() *schema.Resource {
```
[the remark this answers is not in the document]
charliecon: Good call. I'll take a look

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
```diff
@@ -84,7 +100,21 @@ func architectFlowResolver(flowId, exportDirectory, subDirectory string, configM
 
 	log.Printf("Updating resource config and state file for flow '%s'", flowId)
 	updateResourceConfigAndState(configMap, resource, exportDirectory, subDirectory, filename)
-	return nil
+	return err
+}
+
+// sanitizeFlowName will replace all forward slashes, backslashes and white spaces with an underscore
+func sanitizeFlowName(s string) string {
+	// First replace empty strings (multiple spaces) with a single underscore
```
bbbco, line 108: How about just replacing all non-alphanumeric characters with underscore?

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_schema.go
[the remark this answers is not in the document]
charliecon: I made a ticket for that work (DEVTOOLING-1136). I got started on it in this PR but there was a few challenges that popped up that I couldn't sort out right away, so I thought better to leave that complexity for another PR.

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
charliecon: I was going to go with that, but I thought leaving special characters in will make it easier to identify which flows are in which files. If a flow name is mostly special characters, it will come out like `___Foo_1___2___-{ID}.yaml`. I could see that being a bit of a inconvenience for users

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
```diff
@@ -35,8 +41,109 @@ func GenerateFlowResource(resourceLabel, srcFile, fileContent string, forceUnloc
...
+	var (
+		sdkConfig = meta.(*provider.ProviderMeta).ClientConfig
+		proxy     = newArchitectFlowProxy(sdkConfig)
+		ctx       = context.Background()
+		flowName  string
+	)
+
+	flow, resp, err := proxy.GetFlow(ctx, flowId)
+	if err != nil {
+		log.Printf("Failed to establish flow name. Error: %s. API Response: %s", err.Error(), resp)
```
bbbco, line 73: Is "establish" the correct word here?

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
```diff
@@ -35,8 +41,109 @@ func GenerateFlowResource(resourceLabel, srcFile, fileContent string, forceUnloc
...
+	}
+
+	log.Printf("Creating subfolder '%s' inside '%s'", subDirectory, exportDirectory)
+	fullPath := filepath.Join(exportDirectory, subDirectory)
+	if err := os.MkdirAll(fullPath, os.ModePerm); err != nil {
+		return err
+	}
+	log.Printf("Successfully created subfolder '%s' inside '%s'", subDirectory, exportDirectory)
+
+	filename := fmt.Sprintf("%s-%s.yaml", flowName, flowId)
```
bbbco, line 90: We should also include flow type on the filename construction, as flows can have the same name across different flow types.

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
```diff
@@ -35,8 +41,109 @@ func GenerateFlowResource(resourceLabel, srcFile, fileContent string, forceUnloc
...
+	log.Printf("Downloading export flow '%s' to '%s' from download URL", flowId, fullPath)
```
bbbco, line 91: Include flowType in log as well

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
```diff
@@ -35,8 +41,109 @@ func GenerateFlowResource(resourceLabel, srcFile, fileContent string, forceUnloc
...
+	}
+	log.Printf("Successfully downloaded export flow '%s' to '%s'", flowId, fullPath)
+
+	log.Printf("Updating resource config and state file for flow '%s'", flowId)
+	updateResourceConfigAndState(configMap, resource, exportDirectory, subDirectory, filename)
+	return err
+}
+
+// sanitizeFlowName will replace all forward slashes, backslashes and white spaces with an underscore
+func sanitizeFlowName(s string) string {
```
bbbco, line 107: This is to ensure a valid filename is generated right? Should we sanitize all invalid filename characters then? https://stackoverflow.com/a/31976060
Should we pull this into a helper utility?

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -169,6 +170,13 @@ func ResourceTfExport() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"use_legacy_architect_flow_exporter": {
```
bbbco, line 173: Should this be an attribute? Or should this be a feature flag env variable?
If it's kept as an attribute and we eventually want to remove it, won't that impact state?

charliecon: Yes we agreed an attribute on the exporter is the way to go for now. In the future, we can warn the user that the default value will change to `false` and eventually that we're going to remove the field altogether

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
charliecon: I used that phrasing because the api call is being made to establish the flow name and use it in the file name. I updated the log to be clearer

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
charliecon: The intent was to make sure no directory traversal stuff can be done. Also escaped whitespaces in the CLI are an eye sore, so I removed them also. Otherwise, I don't think there is any risk since the `.yaml` extension is hard-coded at the end

kavinbalagen (approved):

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
```diff
@@ -83,46 +98,143 @@ func (a *architectFlowProxy) GetAllFlows(ctx context.Context, name string, varTy
...
+// Parameters:
+//   - a: *architectFlowProxy - The architect flow proxy instance
+//   - jobId: string - The ID of the export job to poll
+//
+// Returns:
+//   - string: The download URL for the exported flow
+//   - error: An error if the polling times out or if there's an issue getting the job status
+//
+// The function will timeout after 90 seconds if the job doesn't complete.
+// It polls the job status every 1 second.
```
bbbco, line 192: Thanks for putting all these docs in for these functions. It will help with previewing the functions in the IDE. Good work @charliecon !

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
```diff
@@ -83,46 +98,143 @@ func (a *architectFlowProxy) GetAllFlows(ctx context.Context, name string, varTy
...
 	for _, flow := range *flows {
 		if *flow.Name == name {
-			matchedFlowIds = append(matchedFlowIds, *flow.Id)
+			matchedFlows = append(matchedFlows, flow)
 		}
-		if len(matchedFlowIds) > 1 {
-			return "", nil, true, fmt.Errorf("found more than one flow that matched the name '%s'", name)
+	}
+
+	// if more than one flow is found, return the one that matches the type also
```
bbbco, line 224: Shouldn't we also ensure the VarType _always_ matches, and not just if there is more than one?

charliecon:
```bash
╷
│ Warning: 'type' should be provided to the genesyscloud_flow data source
│ 
│   with data.genesyscloud_flow.flow,
│   on main.tf line 1, in data "genesyscloud_flow" "flow":
│    1: data "genesyscloud_flow" "flow" {
│ 
│ Please provide a value for the field 'type' in the genesyscloud_flow data source as it will become required in a later version.
╵

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

flow_id = "0387e182-5c0c-44c2-9f34-66c168ebda3e"
```

bbbco (approved):

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
```diff
@@ -185,3 +297,124 @@ func getAllArchitectFlowsFn(ctx context.Context, p *architectFlowProxy, name str
 
 	return &totalFlows, nil, nil
 }
+
+func generateDownloadUrlFn(a *architectFlowProxy, flowId string) (string, error) {
+	log.Printf("Creating export job for flow %s", flowId)
+	jobId, resp, err := a.createExportJob(flowId)
+	if err != nil {
+		log.Println(err)
```
carnellj-genesys, line 305: Should we add some more context to this log message like "Error in generateDownLoadUrlFN.  Otherwise the error message might not give an indication what is wrong.

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
```diff
@@ -185,3 +297,124 @@ func getAllArchitectFlowsFn(ctx context.Context, p *architectFlowProxy, name str
...
+	downloadUrl, err := a.pollExportJobForDownloadUrl(jobId)
+	if err != nil {
+		return "", err
+	}
+	log.Printf("Successfully read download URL. Export job: %s", jobId)
+
+	return downloadUrl, nil
+}
+
+func createExportJobFn(a *architectFlowProxy, flowId string) (string, *platformclientv2.APIResponse, error) {
```
carnellj-genesys, line 323: I would have Q generate documentation and unit tests if you have not already written them else where

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
```diff
@@ -185,3 +297,124 @@ func getAllArchitectFlowsFn(ctx context.Context, p *architectFlowProxy, name str
 
 	return &totalFlows, nil, nil
 }
+
```
carnellj-genesys, line 300: I would have Q documented these methods and also generate unit tests if you do not have them

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
```diff
@@ -185,3 +297,124 @@ func getAllArchitectFlowsFn(ctx context.Context, p *architectFlowProxy, name str
...
+	}
+
+	if createJob == nil || createJob.Id == nil {
+		return "", resp, fmt.Errorf("no export job flow ID returned for flow %s", flowId)
+	}
+
+	return *createJob.Id, nil, nil
+}
+
+func getExportJobStatusByIdFn(a *architectFlowProxy, jobId string) (*platformclientv2.Architectexportjobstateresponse, *platformclientv2.APIResponse, error) {
```
carnellj-genesys, line 346: See earlier comments

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
```diff
@@ -185,3 +297,124 @@ func getAllArchitectFlowsFn(ctx context.Context, p *architectFlowProxy, name str
...
+			log.Println(err)
+			if resp != nil {
+				log.Printf("API Response: %s", resp.String())
+			}
+			return "", err
+		}
+
+		status := *exportJob.Status
+
+		if status == "Started" {
```
carnellj-genesys, line 383: I would break these out into there own if check rather then do if..then else.

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_utils.go
```diff
@@ -35,8 +41,112 @@ func GenerateFlowResource(resourceLabel, srcFile, fileContent string, forceUnloc
...
+// Returns:
+//   - error: Returns an error if any operation fails, nil otherwise
+func architectFlowResolver(flowId, exportDirectory, subDirectory string, configMap map[string]any, meta any, resource resourceExporter.ResourceInfo) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("caught in architectFlowResolver: %w", err)
+		}
+	}()
+
+	var (
```
carnellj-genesys, line 64: Why use a var block here.  Usually a var block is only used when you are not initialziing a value.  In each of these cases an := would be sufficient.

I think you have been hanging out in Typescript land to long Charlie :)

genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go
charliecon: The main method  has all the docs I added a simple comment here also to say that is the implementation function for the method

carnellj-genesys (approved):

PR #1621, opened 2025-03-21
Fixes #1618

### LINKED ISSUE #1618: Unexpected Terraform Plan Changes to Canned Response Libraries in Routing Queues
Terraform detects unexpected changes in canned_response_libraries.library_ids when reapplying a Genesys Cloud Routing Queue, even when no actual modifications are made. This leads to unnecessary updates and deployment inconsistencies.

1. Initial terraform plan / apply (Successful Deployment)
+ canned_response_libraries {
    + library_ids = [
        + "51d22d29-dd6d-4a8c-8c04-627f9e3c8750",
        + "52e49e0f-ea68-4aee-9363-6209d6191855",
        + "694afa2e-709f-433b-b43a-40e0c199ab06",
        + "fd276c7f-274a-4c70-887f-c61cdab66e07",
      ]
    + mode = "SelectedOnly"
}

2. Subsequent terraform plan (No Changes to Canned Response Libraries, but still diff detected)
~ canned_response_libraries {
    ~ library_ids = [
        - "52e49e0f-ea68-4aee-9363-6209d6191855",
        - "fd276c7f-274a-4c70-887f-c61cdab66e07",
          "51d22d29-dd6d-4a8c-8c04-627f9e3c8750",
        + "52e49e0f-ea68-4aee-9363-6209d6191855",
          "694afa2e-709f-433b-b43a-40e0c199ab06",
        + "fd276c7f-274a-4c70-887f-c61cdab66e07",
      ]
}

- Expected Behavior
  Terraform should only detect actual additions/removals, not reordering.

- Impact
   Causes unnecessary Terraform updates on every apply.
   Creates deployment inconsistencies despite no real changes in Genesys Cloud.

While looking into the code I found that,
https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go

"library_ids": {
    Description: "Set of canned response library IDs associated with the queue. Populate this field only when the mode is set to SelectedOnly.",
    Optional: true,
    Type: schema.TypeList,
    Elem: &schema.Schema{Type: schema.TypeString},
}

**Type: schema.TypeList**
The issue here is that TypeList enforces an ordered list, meaning Terraform detects changes even when the order of IDs is different but the contents remain the same. This is causing unnecessary updates when no real changes have been made.
It might help by changing TypeList to TypeSet, as TypeSet treats values as an unordered collection.

charliecon:


genesyscloud/responsemanagement_response/resource_genesyscloud_responsemanagement_response_schema.go
```diff
@@ -143,7 +143,7 @@ func ResourceResponsemanagementResponse() *schema.Resource {
 				Description: `One or more libraries response is associated with. Changing the library IDs will result in the resource being recreated`,
 				Required:    true,
 				ForceNew:    true,
-				Type:        schema.TypeList,
+				Type:        schema.TypeSet,
```
bbbco, line 146: I think this requires a state migration, otherwise Terraform will force recreate any of this resource type on the next apply because the data is stored differently on the state file.

bbbco (changes requested):

genesyscloud/responsemanagement_response/resource_genesyscloud_responsemanagement_response_schema.go
charliecon: I deployed a response with the latest published version and then upgraded to the locally compiled and saw no issues (`No changes.`)

I don't think this is a problem because in the state file `library_ids` is just a JSON string array with no mention to TypeList or Set

bbbco (approved):

PR #1624, opened 2025-03-24
Fixing some vetting errors

charliecon:


bbbco (approved):
My bad for not taking care of these tests. Thanks for doing this @charliecon

charliecon:
> My bad for not taking care of these tests. Thanks for doing this @charliecon

No problem 👍

PR #1655, opened 2025-04-09
Adding LastAgentRoutingMode field to Routing Queue resource

tamasgyenese-genesys:
Adding LastAgentRoutingMode field to Routing Queue resource


charliecon (approved):
Beautiful

PR #1659, opened 2025-04-10
Feat/devtooling 1149 - Added team_ids to genesyscloud_recording_media_retention_policy

shrutisuryawanshigenesys:


genesyscloud/recording_media_retention_policy/resource_genesyscloud_recording_media_retention_policy_schema.go
```diff
@@ -144,6 +144,12 @@ func ResourceMediaRetentionPolicy() *schema.Resource {
 				Default:     nil,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
+			"team_ids": {
+				Description: "Teams to match conversations against",
+				Type:        schema.TypeList,
```
charliecon, line 149: TypeSet is probably better to use here. We've had to change a few fields from TypeList to TypeSet recently because if the API returns the list in a different order, we'll see a plan not empty error. This doesn't happen with TypeSet though because it doesn't care about the ordering of items.

genesyscloud/recording_media_retention_policy/resource_genesyscloud_recording_media_retention_policy_test.go
```diff
@@ -1372,7 +1480,120 @@ func TestAccResourceMediaRetentionPolicyBasic(t *testing.T) {
...
+						}),
+					) +
+					team.GenerateTeamResource(
+						teamIdResourceLabel1,
+						teamIdName1,
+						"genesyscloud_auth_division."+divResourceLabel+".id",
+						teamDescription1,
+					) +
+					generateMediaRetentionPolicyResource(policyResourceLabel1, &mediaRetentionMessagePolicy2),
+				Check: resource.ComposeTestCheckFunc(
```
charliecon, line 1530: The function `ValidateResourceAttributeInArray` is useful here for testing that the id of another resource is present in a list

PR #1662, opened 2025-04-14
Oath secret creds

HemanthDogiparthi12:


genesyscloud/integration_credential/resource_genesyscloud_integration_credential_utils.go
```diff
@@ -2,31 +2,88 @@ package integration_credential
...
+
 	if fields, ok := d.GetOk("fields"); ok {
-		fieldMap := fields.(map[string]interface{})
-		for k, v := range fieldMap {
-			results[k] = v.(string)
+		if fieldMap, ok := fields.(map[string]interface{}); ok {
+			for k, v := range fieldMap {
+				results[k] = v.(string)
+			}
+			fmt.Printf("%v fields", results)
```
charliecon, line 46: left over print statement

genesyscloud/oauth_client/resource_genesyscloud_oauth_client.go
```diff
@@ -283,6 +285,11 @@ func cascadeUpdateOAuthClient(ctx context.Context, d *schema.ResourceData, meta
 	// check if there is a integration credential to update/create
 	if dealIntegrationFlag {
 		credentialId := resourcedata.GetNillableValue[string](d, "integration_credential_id")
+
+		if client.Secret == nil || *client.Secret != "" {
```
charliecon, line 289: Should this be `if secret != nil && *secret != ""` ?

genesyscloud/oauth_client/resource_genesyscloud_oauth_client.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/mypurecloud/platform-client-sdk-go/v154/platformclientv2"
 )
 
+var CacheFile = "/tmp/shared-cache.json"
```
charliecon, line 25: The [os.TempDir](https://pkg.go.dev/os#TempDir) function would be useful here. Also better to use `filepath.Join` to make sure the path built is compatible with different operating systems

genesyscloud/oauth_client/resource_genesyscloud_oauth_client_schema.go
```diff
@@ -111,10 +110,27 @@ func ResourceOAuthClient() *schema.Resource {
...
+				StateFunc:   ModifyStatusIdStateValue,
+			},
 		},
 	}
 }
 
+func ModifyStatusIdStateValue(id interface{}) string {
+	return ""
+}
+
```
charliecon, line 133: Can this be removed or does it serve a purpose?

genesyscloud/oauth_client/resource_genesyscloude_oauth_client_unit_test.go
```diff
@@ -13,6 +16,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var testCacheFile = "/tmp/test-cache.json"
```
charliecon, line 19: Same here

goreleaser
charliecon, line 1: Should this be here? looks like an empty dir

genesyscloud/oauth_client/resource_genesyscloud_oauth_client_schema.go
HemanthDogiparthi12: just to make sure the state is empty even if some one tries set it in the context methods.

genesyscloud/oauth_client/resource_genesyscloud_oauth_client.go
HemanthDogiparthi12: its correct.

genesyscloud/integration_credential/resource_genesyscloud_integration_credential_utils.go
HemanthDogiparthi12: removed.

charliecon (approved):

PR #1679, opened 2025-04-25
(DEVTOOLING-1146) genesyscloud_integration_action: Making sure exporter BlockLabel is unique for every resource

charliecon:


genesyscloud/integration_action/resource_genesyscloud_integration_action.go
```diff
@@ -61,7 +64,16 @@ func getAllIntegrationActions(ctx context.Context, clientConfig *platformclientv
...
 		}
-		resources[*action.Id] = &resourceExporter.ResourceMeta{BlockLabel: *action.Name, BlockHash: blockHash}
+
+		// It is more convenient for labels to not be org-specific, so if possible we will use the name alone as the label.
+		// If any actions share the same name, we will use the ID in the label.
+		var blockLabel string
+		if allNamesAreUnique {
+			blockLabel = *action.Name
+		} else {
+			blockLabel = *action.Name + "_" + *action.Id
```
bbbco, line 74: @charliecon I think we can already guarantee uniqueness to the BCP team by enabling the BCP sanitizer (in this upcoming release). 

I do have a bit of concern adding the action ID to the block label, namely:
- the BCP Wizard does extra processing to try to find the integration actions and do some extra work with them. Adding an inconsistent format that might get exported one way or another depending on whether there are duplicates or not is not idempotent, which we require in our BCP workflows.
- Adding the id to the block label prevents the resources from being able to get matched up when syncing between two orgs. The block label is what we use to match two resources up between the two orgs (regardless of ID). It's important that the block label is able to be consistently generated without attributes that can be different from one org to the next. See the documentation notes on the BlockHash attribute for more details.
-

bbbco (changes requested):
Please reconsider how the block label is being generated

kavinbalagen (approved):

PR #1681, opened 2025-04-28
Feature/devtooling 1162

kavinbalagen:


genesyscloud/scripts/genesyscloud_scripts_proxy.go
```diff
@@ -522,3 +530,31 @@ func scriptExistsWithName(ctx context.Context, scriptsProxy *scriptsProxy, scrip
 	}
 	return true, nil
 }
+
+func setScriptDivision(scriptId, divisionId string, p *scriptsProxy) error {
```
charliecon, line 534: Did you try adding `division_id` to the form data in the function [createScriptFormData](https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/scripts/genesyscloud_scripts_proxy.go#L249) ? I didn't think we would have to use a different endpoint

PR #1682, opened 2025-04-28
DEVTOOLING-1189 Fix/examples by promoting them to full fledged test cases that must be passed 

bbbco:
DRAFT STATE: Would appreciate feedback on this as I'm working on it @HemanthDogiparthi12 @charliecon @carnellj 
---
The approach I'm taking to resolving https://inindca.atlassian.net/browse/DEVTOOLING-1189 is to promote the content in the examples to full fledged test cases that must pass. This will help ensure that the examples docs are consistent and accurate. 

My implementation allows us to run a simple `plan` sanity check on examples whenever `make docs` is run, as well as a full implementation `apply` whenever the full suite of tests are run to ensure examples are always consistent in the provider.

This also enables me to do a re-runnable audit against each resource when running against Cloud CX1 orgs (which is the real goal of this task).


bbbco:
@HemanthDogiparthi12 @carnellj @charliecon This PR is ready for review.

charliecon (approved):

PR #1684, opened 2025-04-29
[Devtooling-1181] add 'filter' to outbound_campaign

fergalmcl:


genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_utils.go
```diff
@@ -182,6 +185,7 @@ func buildLineBalancingSettings(settings []interface{}) *platformclientv2.Dynami
 func flattenSettings(settings *platformclientv2.Dynamiccontactqueueingsettings) []interface{} {
 	settingsMap := make(map[string]interface{}, 0)
 	settingsMap["sort"] = *settings.Sort
+	settingsMap["filter"] = *settings.Filter
```
charliecon, line 188: `SetMapValueIfNotNil` should be used here for both of these to be safe

genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_schema.go
```diff
@@ -248,6 +248,12 @@ func ResourceOutboundCampaign() *schema.Resource {
 							Required:    true,
 							ForceNew:    true,
 						},
+						"filter": {
+							Description: "Whether to filter contacts dynamically.",
+							Type:        schema.TypeBool,
+							Required:    true,
+							ForceNew:    true,
+						},
```
charliecon, line 256: Do you mind removing the ForceNew field from both of these and seeing if you can update them? I feel like it might be unnecessary. If you can't update them and the ForceNew is appropriate, then I would add a note to the end of description stating that updating the field will cause the campaign to be dropped and recreated with a new GUID.

genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_schema.go
```diff
@@ -248,6 +248,12 @@ func ResourceOutboundCampaign() *schema.Resource {
 							Required:    true,
 							ForceNew:    true,
 						},
+						"filter": {
+							Description: "Whether to filter contacts dynamically.",
+							Type:        schema.TypeBool,
+							Required:    true,
```
bbbco, line 254: If this is now a required field, we need to add a schema migration that automatically updates existing orgs with the proper value in the state, along with instructions on how to update the config.

charliecon: I don't think it should be required if it didn't exist before

genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_schema.go
fergalmcl: Updating the attribute after creation causes an api error so ForceNew is needed, updated description

charliecon (approved):

PR #1687, opened 2025-05-01
[Devtooling-1139] User phone number extension not showing in extension assignments

fergalmcl:
Adds a reference attribute for the extension pool in user to ensure the creation order, while also exporting reference so apply happens in the correct order.


genesyscloud/user/genesyscloud_user_proxy.go
```diff
@@ -314,3 +325,43 @@ func updatePasswordFn(ctx context.Context, p *userProxy, userId string, newPassw
...
+func getTelephonyExtensionPoolByExtensionFn(ctx context.Context, p *userProxy, extNum string) (*platformclientv2.Extensionpool, *platformclientv2.APIResponse, error) {
+	var allPools []platformclientv2.Extensionpool
+	extensionPoolList, apiResponse, err := p.extensionPoolApi.GetTelephonyProvidersEdgesExtensionpools(100, 1, "", "")
+	if err != nil {
+		return nil, apiResponse, err
+	}
+	allPools = append(allPools, *extensionPoolList.Entities...)
+
+	for pageNumber := 2; pageNumber <= *extensionPoolList.PageCount; pageNumber++ {
+		extensionPoolList, apiResponse, err := p.extensionPoolApi.GetTelephonyProvidersEdgesExtensionpools(100, pageNumber, "", "")
```
HemanthDogiparthi12, line 338: is 100 page size here. move it to a constant

genesyscloud/user/genesyscloud_user_proxy.go
```diff
@@ -314,3 +325,43 @@ func updatePasswordFn(ctx context.Context, p *userProxy, userId string, newPassw
...
+		if err != nil {
+			return nil, apiResponse, err
+		}
+		allPools = append(allPools, *extensionPoolList.Entities...)
```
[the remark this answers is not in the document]
fergalmcl: added

HemanthDogiparthi12:
@fergalmcl seems the api params changed after the go sdk update. can you correct this PR.

genesyscloud/user/resource_genesyscloud_user_schema.go
```diff
@@ -51,6 +51,11 @@ var (
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"extension_pool_id": {
+				Description: "Id of the extension pool which contains this extension.",
+				Type:        schema.TypeString,
+				Optional:    true,
```
bbbco, line 57: Shouldn't this be Computed instead of Optional? We don't necessarily want users to set this.

fergalmcl: The attribute is Optional because it's only use is to control the dependancy for creation/deletion order, the idea is users can set the ID manually when trying to create an extension assignment from terraform, or when exporting existing users and extension pools the dependancy ensures the apply then creates the assignments properly.

PR #1695, opened 2025-05-07
(DEVTOOLING-1146) genesyscloud_tf_export: Duplicate data actions were not all appearing in exported state file

charliecon:
The original sanitizer will now append a hash to the second instance of a sanitized label, and append the same hash again for each subsequent occurrence. 

* First occurrence: `example_san_label`
* Second: `example_san_label_{hash}`
* Third: `example_san_label_{hash}_{hash}`


genesyscloud/resource_exporter/resource_name_sanitizer.go
```diff
@@ -62,29 +62,31 @@ func (sod *sanitizerOriginal) Sanitize(idMetaMap ResourceIDMetaMap) {
...
+		numOfOccurrences := labelOccurrences[sanitizedLabel]
+		for i := 0; i < numOfOccurrences; i++ {
+			sanitizedLabel = sanitizedLabel + "_" + sod.SanitizeResourceHash(meta.BlockLabel)
+		}
+
+		if meta.OriginalLabel == "" {
+			meta.OriginalLabel = meta.BlockLabel
+		}
+
+		meta.BlockLabel = sanitizedLabel
```
bbbco, line 89: So this implementation is very much in line with the Optimized Sanitizer by removing the double for loop quadratic time complexity bug. If you want to go with this as the default approach, we should probably remove the Optimized Sanitizer.
We do need to retain the BCP Sanitizer as that constructs the block labels with hash values across the board in order to guarantee uniqueness and matching across orgs. This unfortunately is not very human readable, so not advised to be used by default.

charliecon: I removed the optimised sanitiser. Also pinged Cesar so he can take a glance before moving forward. It's worth noting I also made a small change to this approach — the first label will not have any hash, only the second occurrence and so on

genesyscloud/resource_exporter/resource_name_sanitizer.go
```diff
@@ -112,63 +99,6 @@ func (sod *sanitizerOriginal) SanitizeResourceHash(originalBlockLabel string) st
...
-			} else {
-				sanitizedLabels[sanitizedLabel] = 1
-				meta.BlockLabel = sanitizedLabel
-			}
-		}
-	}
-}
-
-// SanitizeResourceBlockLabel sanitizes a single resource label
-func (sod *sanitizerOptimized) SanitizeResourceBlockLabel(inputLabel string) string {
```
bbbco, line 147: Please pull this implementation in to the original sanitizer. It adds useful transliteration of the block labels (so Japanese characters become ASCII character equivalents).

genesyscloud/tfexporter/tfstate_exporter.go
```diff
@@ -34,22 +34,22 @@ func NewTFStateWriter(ctx context.Context, resources []resourceExporter.Resource
...
 
-	return tfwriter, nil
+	return tfWriter, nil
 }
 
 func (t *TFStateFileWriter) writeTfState() diag.Diagnostics {
 
-	platform := platform.GetPlatform()
-	platformErr := platform.Validate()
+	platformInstance := platform.GetPlatform()
```
bbbco, line 51: Nice catch

bbbco (changes requested):
Please pull back in the transliteration of the optimized sanitizer

genesyscloud/resource_exporter/resource_name_sanitizer.go
charliecon: Done. I pulled the optimized sanitizer's SanitizeResourceHash method back in also

bbbco (approved):
Please be sure to call this out in the Release Notes that the Block Labels will look different from this release onward by default.

PR #1698, opened 2025-05-07
[Devtooling-1202] Added live_speaker_detection_mode, amd_speech_distinguish_enabled to outbound_callanalysisresponseset

fergalmcl:


genesyscloud/outbound_callanalysisresponseset/resource_genesyscloud_outbound_callanalysisresponseset_utils.go
```diff
@@ -15,6 +15,16 @@ func getResponseSetFromResourceData(d *schema.ResourceData) platformclientv2.Res
 		BeepDetectionEnabled: platformclientv2.Bool(d.Get("beep_detection_enabled").(bool)),
 	}
 
+	amdSpeechDistinguishEnabled := d.Get("amd_speech_distinguish_enabled").(bool)
+	if amdSpeechDistinguishEnabled {
+		sdkResponseSet.AmdSpeechDistinguishEnabled = &amdSpeechDistinguishEnabled
```
charliecon, line 20: Probably we shouldn't only set this if it is true. That way we won't be able to update it to false in another apply. Is there any reason why we can't just set `AmdSpeechDistinguishEnabled = d.Get("amd_speech_distinguish_enabled").(bool)` directly ?

genesyscloud/outbound_callanalysisresponseset/resource_genesyscloud_outbound_callanalysisresponseset_utils.go
```diff
@@ -94,14 +104,16 @@ func flattenSdkReaction(sdkReaction platformclientv2.Reaction) *schema.Set {
...
 }
 
-func GenerateOutboundCallAnalysisResponseSetResource(resourceLabel string, name string, beepDetectionEnabled string, responsesBlock string) string {
+func GenerateOutboundCallAnalysisResponseSetResource(resourceLabel string, name string, beepDetectionEnabled string, amdSpeechDistinguishEnabled string, liveSpeakerDetectionMode string, responsesBlock string) string {
 	return fmt.Sprintf(`
 resource "genesyscloud_outbound_callanalysisresponseset" "%s" {
 	name                   = "%s"
 	beep_detection_enabled = %s
+	amd_speech_distinguish_enabled = %s
+	live_speaker_detection_mode = "%s"
```
charliecon, line 113: Could you change this to `live_speaker_detection_mode = %s` and where ever you call `GenerateOutboundCallAnalysisResponseSetResource` , provide `"null"` or `strconv.Quote("the value")` ? 

In the first test step in TestAccResourceOutboundCallAnalysisResponseSet , it's better to keep both these new fields `null` , verify they're `null`, and then update them in the next step. This way we know existing configs won't be broken when this change is released

genesyscloud/outbound_callanalysisresponseset/resource_genesyscloud_outbound_callanalysisresponseset_utils.go
fergalmcl: Not sure why I did that, fixed now

charliecon (approved):

PR #1702, opened 2025-05-08
NO-JIRA fix issue with flow outcomes where a creation fails if it already exists.

bbbco:
Found while working through https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1682

 NO-JIRA fix issue with flow outcomes where a creation fails if it already exists. Because these resources cannot be deleted, we should be able to attach to the existing one and update it


genesyscloud/flow_outcome/resource_genesyscloud_flow_outcome.go
```diff
@@ -45,6 +45,15 @@ func createFlowOutcome(ctx context.Context, d *schema.ResourceData, meta interfa
 	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
 	proxy := getFlowOutcomeProxy(sdkConfig)
 
+	flowOutcomeName := d.Get("name").(string)
+
+	// Since Flow Outcomes cannot be deleted, we need to be able to check if there is an existing Flow Outcome and update it
+	flowOutcomeId, _, existingFlowOutcomeApiResp, _ := proxy.getFlowOutcomeIdByName(ctx, flowOutcomeName)
+	if existingFlowOutcomeApiResp.StatusCode != 404 {
```
charliecon, line 52: I had a look at the function `getFlowOutcomeIdByNameFn` and it looks like the endpoint [GET /api/v2/flows/outcomes](https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-flows-outcomes) will return a 200 if none exist with that name. I think this approach is more appropriate 

```go
if flowOutcomeId != "" {
  d.SetId(flowOutcomeId)
  return update
}

if err != nil && !retrayble {
  log.Println("error occurred: ", err)
  log.Println("Attempting to create...")
}
```

bbbco: Thanks for catching this @charliecon ! For some reason I was initially getting a response code that indicated it was not being found, but on revisiting, I see that your suggestion is correct.

charliecon (approved):

PR #1710, opened 2025-05-15
Fix issues with dangling resources on export due to not cleaning up the resource cache on delete item

bbbco:
As I was working on https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1682, specifically when testing the tf_exporter examples, it kept failing due to dangling resources.

Digging in, I discovered that the reason came back to the fact that many resources were using `get{Resource}ById()` after deletion to confirm the delete. They typically first check the cache and returns items. The reason they were returning items is because the cache item hadn't been removed yet.

This PR is the result of an audit I did to ensure that all resources that exercised a cache ensured the cache item was deleted at the time the resource instance was deleted.


charliecon (approved):
Awesome 👌

PR #1715, opened 2025-05-17
NO-JIRA Fixes for quality forms

bbbco:


charliecon (approved):
You'll need to run `make docs` to render the schema changes. Also there is one merge conflict. Otherwise looks good

Issue #1705, opened 2025-05-14
Version 1.63 removes members from routing queue


stevendewaele:
By migrating the provider version to the latest 1.63.0 we noticed that the members are removed from the queues.
However in our code we are not using the members property

charliecon:
Hi @stevendewaele 

Thank you for raising this issue. Let me explain the recent changes:

* In v1.63.0, we removed the Computed flag from the members field. This change means Terraform now expects to be the sole manager of queue members, allowing empty members sets to clear all queue members as many users requested.
* To address backward compatibility concerns, we're introducing a new field `ignore_members` in the upcoming v1.64.0 release (scheduled for this week). When set to true, this field will:
  * Allow you to omit the members field entirely
  * Prevent Terraform from managing queue members
  * Maintain compatibility with existing workflows

This solution should provide flexibility for both use cases - full Terraform management of queue members and external management as needed. 

FYI: long-term, we are looking at moving queue member management into its own resource for a more explicit and understandable implementation.

Hope this helps and thanks for raising the issue!

-Charlie

PR #1720, opened 2025-05-19
(NO-JIRA) Reintroducing Computed attribute to routing_queue.members field

charliecon:
@HemanthDogiparthi12 

I reinstated the Computed field on members to address issue #1705, but Fergal pointed out to be me that it was removed for DEVTOOLING-1150 

Basically some users have become used to members being computed i.e. they can add members to a queue outside of Terraform without the next terraform apply deleting all members (assuming this queue is being managed in the tf config and that `members` is empty.) Other users want Computed to be false / `members` to be controlled via Terraform only.

I think the best approach now is to remove the Computed value and to create a new queue_members resource.

### DIFF
--- a/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
+++ b/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
@@ -603,7 +603,6 @@ func ResourceRoutingQueue() *schema.Resource {
 				Type:        schema.TypeSet,
 				Optional:    true,
 				ConfigMode:  schema.SchemaConfigModeAttr,
-				Computed:    true,
 				Elem:        queueMemberResource,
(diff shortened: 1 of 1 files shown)


HemanthDogiparthi12:
> @HemanthDogiparthi12
> 
> I reinstated the Computed field on members to address issue #1705, but Fergal pointed out to be me that it was removed for DEVTOOLING-1150
> 
> Basically some users have become used to members being computed i.e. they can add members to a queue outside of Terraform without the next terraform apply deleting all members (assuming this queue is being managed in the tf config and that `members` is empty.) Other users want Computed to be false / `members` to be controlled via Terraform only.
> 
> I think the best approach now is to remove the Computed value and to create a new queue_members resource.

lets create a ticket for this.

PR #1725, opened 2025-05-20
(NO-JIRA) genesyscloud_organization_presence_definition: Removing validators to fix export

charliecon:
While trying to recreate another export bug, I caught a few instances where the validator functions in the new presence definition resource was producing plan errors e.g. `expected system_presence to be one of [...] got "abc"`

To fix this I removed the validators. I think it is better to list the known valid options in the descriptions and leave the validation up to the API.

### DIFF
--- a/docs/resources/organization_presence_definition.md
+++ b/docs/resources/organization_presence_definition.md
@@ -36,7 +36,7 @@ resource "genesyscloud_organization_presence_definition" "Away_From_Keyboard" {
 
 ### Required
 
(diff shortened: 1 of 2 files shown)


bbbco:
It seems to me that having schema validation is more preferred as you know right off the bat if you are providing a valid config (can also use `terraform validate` to confirm). Can we not update the validation values to match the API?

charliecon:
> It seems to me that having schema validation is more preferred as you know right off the bat if you are providing a valid config (can also use `terraform validate` to confirm). Can we not update the validation values to match the API?

@bbbco I brought back the validation function for `system_presence` because the API actually returns a 400 error if you provide an invalid string. The slice of potential values was missing a few ("Idle" for example), but I didn't need to remove the validation function in that case. 

For the language labels, I think validation will only cause issues down the road. I was able to use "random" as a key in the map without any problems. Unless the team that owns the API implements stricter rules, I bet we'll see TF users come back to us saying "I can do xyz over here but CX as Code won't let me"

bbbco (approved):

PR #1761, opened 2025-06-17
[Devtooling 1245] AI Guide Resource

BrianMoyles:


genesyscloud/guide/resource_genesyscloud_guide.go
```diff
@@ -0,0 +1,151 @@
...
+		guide, resp, err := proxy.getGuideById(ctx, d.Id())
+		if err != nil {
+			if util.IsStatus404(resp) {
+				log.Printf("Deleted Guide: %s", d.Id())
+				return nil
+			}
+			return retry.RetryableError(util.BuildWithRetriesApiDiagnosticError(ResourceType, fmt.Sprintf("error retrieving guide %s | error: %s", d.Id(), err), resp))
+		}
+
+		if guide != nil {
```
charliecon, line 134: Do we need to check if guide is nil here? doesn't seem to make a difference. Also, if it is nil then we go right to the retryable error "guide still exists" which doesn't really make sense. Maybe that error should be `"Unexpected job status: %s"`

genesyscloud/guide/resource_genesyscloud_guide_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"context"
+	"fmt"
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/provider"
+	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util"
+	"testing"
+)
+
```
charliecon, line 13: Not important but we could be using the ResourceType const here in a few places

genesyscloud/guide/resource_genesyscloud_guide_test.go
```diff
@@ -0,0 +1,76 @@
...
+func TestAccResourceGuide(t *testing.T) {
```
charliecon, line 14: I think we should check that we're running in tca. That would look like this:

```go
if v := os.Getenv("GENESYSCLOUD_REGION"); v == "tca" {
```

Also Hakan may need some feature toggles etc. I'm not sure if the jenkins tca tests run against the same org you've been using

genesyscloud/guide/resource_genesyscloud_guide_utils.go
```diff
@@ -0,0 +1,48 @@
...
+import "net/http"
+
+func setRequestHeader(r *http.Request, p *guideProxy) *http.Request {
+	r.Header.Set("Content-Type", "application/json")
+	r.Header.Set("Accept", "application/json")
+	r.Header.Set("Authorization", "Bearer "+p.clientConfig.AccessToken)
+	return r
+}
+
+type DeleteObject struct {
```
charliecon, line 12: I think `DeleteObjectJob` is a better name. Makes it easier to understand what `Id` is referring to

genesyscloud/guide/resource_genesyscloud_guide_test.go
BrianMoyles: Added and confirmed with Hakan that this should pass when tested

genesyscloud/guide/resource_genesyscloud_guide_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util"
+	"os"
+	"testing"
+)
+
+func TestAccResourceGuide(t *testing.T) {
+	if v := os.Getenv("GENESYSCLOUD_REGION"); v != "tca" {
+		t.Skipf("Skipping test for region %s. genesyscloud_guide is currently only supported in tca", v)
+		return
+	}
```
charliecon, line 19: We need to do this for the data source test also

charliecon (approved):

PR #1765, opened 2025-06-20
(DEVTOOLING-1230) Addressed the issue of increased export times 

charliecon:
* DEVTOOLING-991 introduced an extra API call to the export of users; [GET /api/v2/routing/users/{userId}/utilization](https://developer.genesys.cloud/devapps/api-explorer#get-api-v2-routing-users--userId--utilization). This almost doubles the export time for users because the endpoint is sensitive to rate-limiting. Removing the singleton pattern in the proxy file addressed this. 
* The main cause of the export time increase was the work done for DEVTOOLING-1191. Sometimes the download URL for a contact list export would take some time generate (if the list is huge), so a sleep was introduced. The mistake was sleeping for 30 seconds even before making the API call. So a 30 second sleep occurred for each contact list and 600 * 30 seconds is 5 hours.

### DIFF
--- a/genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_utils.go
+++ b/genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_utils.go
@@ -235,8 +235,8 @@ func ContactsExporterResolver(resourceId, exportDirectory, subDirectory string, 
 		}
 
 		// Give the system time to generate the URL
-		// Exponential backoff - 1st sleep: 1 second, 10th sleep: 200 seconds
-		waitTime := time.Duration(math.Pow(200, float64(retryAttempt)/10)) * time.Second
+		// Exponential backoff - 1st sleep: 2 seconds, 10th sleep: 1024 seconds (total: 2046 seconds/34 minutes)
+		waitTime := time.Duration(2*math.Pow(2, float64(retryAttempt)-1)) * time.Second
 		log.Printf("Sleeping for %f seconds before retrying", waitTime.Seconds())
 		time.Sleep(waitTime)
 
--- a/genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_utils.go
+++ b/genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_utils.go
@@ -8,6 +8,8 @@ import (
 	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util"
 	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util/files"
 	"log"
+	"math"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


genesyscloud/outbound_contact_list/resource_genesyscloud_outbound_contact_list_utils.go
```diff
@@ -219,13 +218,21 @@ func ContactsExporterResolver(resourceId, exportDirectory, subDirectory string,
...
-		// Sleep at least 30 secs before attempting to retrieve export url to give the system time to be able to generate the URL
-		// This has an exponential backoff to permit large lists the time to process and serve up the export signed url
-		time.Sleep(time.Duration(30*retryAttempt) * time.Second)
 		var err error
 		var resp *platformclientv2.APIResponse
 		exportUrl, resp, err = cp.getContactListContactsExportUrl(ctx, contactListId)
 		if err != nil {
+			// Sleep at least 30 secs before attempting to retrieve export url to give the system time to be able to generate the URL
+			// This has an exponential backoff to permit large lists the time to process and serve up the export signed url
+			waitTime := time.Duration(30*retryAttempt) * time.Second
```
bbbco, line 227: It might be better to revamp the backoff logic:

We could update the algorithm similar to `W(n) = 200^(n/10)` seconds where `n` is the number of tries (the util.RetryWhen retries up to 10 times) so that the most quickest exports would take a min of ~ 2 seconds and longest could take up to ~34 minutes

W(1) = 200^(1/10) ≈ 2.15 seconds
W(2) = 200^(2/10) ≈ 4.64 seconds
W(3) = 200^(3/10) ≈ 10.00 seconds
W(4) = 200^(4/10) ≈ 21.54 seconds
W(5) = 200^(5/10) ≈ 46.42 seconds
W(6) = 200^(6/10) ≈ 100.00 seconds
W(7) = 200^(7/10) ≈ 215.44 seconds
W(8) = 200^(8/10) ≈ 464.16 seconds
W(9) = 200^(9/10) ≈ 1000.00 seconds
W(10) = 200^(10/10) = 200.00 seconds

Total max wait time: ≈ 2064.35 seconds (about 34.41 minutes)

charliecon: Agreed. On it 👍

bbbco: @charliecon Well, this is a flaw in asking an LLM to output the first ten iterations of an algorithm and total it. Reviewing these values again it is completely wrong.

Working back through the algorithm with values from a spreadsheet, the algorithm should be `W(n) = 2*2^(i-1)`:

image

2046 seconds is around 34 minutes

charliecon: No problem. I'll make that change

HemanthDogiparthi12 (approved):

PR #1770, opened 2025-07-01
[Devtooling 1247] AI Guide Version Resource 

BrianMoyles:


examples/resources/genesyscloud_guide_version/locals.tf
```diff
@@ -0,0 +1,7 @@
+locals {
+  dependencies = {
+    resource = [
+      "../../data-sources/genesyscloud_guide/data-source.tf",
```
bbbco, line 4: Is the guide defined in the `data-source.tf` file referenced guaranteed to always be on the system? Otherwise you should probably use just a reference to the resource guide you have defined in the `resources` directory. You would also need to update the reference in this directory's `resource.tf` file as defined below

examples/resources/genesyscloud_guide_version/resource.tf
```diff
@@ -0,0 +1,23 @@
+resource "genesyscloud_guide_version" "sample-guide" {
+  guide_id    = data.genesyscloud_guide.id
```
bbbco, line 2: Whether you are using the reference to `examples/data-sources/genesyscloud_guide/data-source.tf` or `examples/resources/genesyscloud_guide/resource.tf` in your `locals.tf`, your resource path needs to match (`data.genesyscloud_guide.sample_guide.id` or `genesyscloud_guide.sample_guide.id` respectively).

examples/resources/genesyscloud_guide_version/resource.tf
```diff
@@ -0,0 +1,23 @@
...
+  instruction = "This is a test Instruction"
+  variables {
+    name        = "TestVariable"
+    type        = "String"
+    scope       = "InputAndOutput"
+    description = "This is a test Description"
+  }
+  resources {
+    data_action {
+      data_action_id = "DataActionId"
```
bbbco, line 12: Is this supposed to be a reference to an actual GC Data Action? If so, you should pull in those references from existing examples.

examples/resources/genesyscloud_guide_version/resource.tf
bbbco: It might be good for you to check out `examples/README.md` if you need more info

genesyscloud/util/feature_toggles/guides.go
```diff
@@ -0,0 +1,15 @@
...
+
+func GuideToggleName() string {
+	return guideEnvToggle
+}
+
+func GuideToggleExists() bool {
+	var exists bool
+	_, exists = os.LookupEnv(guideEnvToggle)
+	return exists
+}
```
charliecon, line 15: you might have misunderstood what I was saying about checking the toggle. I meant make an API request to [GET /api/v2/guides](https://developer.genesys.cloud/platform/preview-apis#get-api-v2-guides) to check if the feature toggle was enabled in GrandCentral. If the request returns 5xx error, skip the tests. I don't think there's any need for an env var feature toggle

charliecon (approved):

PR #1772, opened 2025-07-01
changed to warn

kavinbalagen:


carnellj-genesys (approved):

kavinbalagen:
> Why are we just warning here? If the user doesn't have permissions required, will they be able to use the resource properly?

this is not checking the user's permissions or oauth's permissions. that will be done by backend, earlier the code was put to validate if a permission that is being created existed in the org or not.

PR #1783, opened 2025-07-07
Fix DEVTOOLING-1286 - export timeouts/errors explicitly spelled out at end

bbbco:
Fix DEVTOOLING-1286 - ensure export errors/timeouts to be explicitly called out at end of run with extra errors JSON file written

I'm pulling this into the existing work of the #1779 with enhanced logging, fixes for ordering of structs, and outputting export errors report.


HemanthDogiparthi12:
the problem is with 
resourceCtx, resourceCancel := context.WithTimeout(context.Background(), 5*time.Minute)

even if we increase token_acquire_timeout , it would still fail , if the context reaches 5 mins. 

would propose to have an indefinite context  here.

bbbco:
Merged latest dev in and fixed merge errors @HemanthDogiparthi12 . Should be good!

PR #1798, opened 2025-07-16
feat: CPR-941: Deprecate WebDeployment Configuration fields

jshankarc:
These deprecated fields will be replaced with new fields post GA release of the AI-1102 feature.


charliecon (commented):
I don't think we should deprecate fields until we can tell the user "use xyz instead. Is it the case that these fields don't have any effect already? If they still serve a purpose until these new fields go GA, better to hold off on the deprecation

charliecon:
> > I don't think we should deprecate fields until we can tell the user "use xyz instead. Is it the case that these fields don't have any effect already? If they still serve a purpose until these new fields go GA, better to hold off on the deprecation
> 
> Yes, these fields are exposed, but they have no effect. We wanted to mention that they are deprecated, to ensure no one uses these fields anymore.

Understood. Two things:

1. Please change the target branch of the PR to `dev`
2. A deprecation message like "This field has no effect and will be removed in a later version." would be better here

jshankarc:
> These deprecated fields will be replaced with new fields post GA release of the AI-1102 feature.

Changing to dev has a problem with rebasing, so I created a new PR - https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1799

PR #1803, opened 2025-07-23
RULES-974: Add support for Rules Schemas Resources and Datasource

brlenn:
Hi folks, 

Adding a new resource and Datasource for Rules Schema, this is the first of a few resources that is being added for the Rule-Based Decisions feature.


brlenn:
I didn't see any docs around acceptance tests and adding products/FTs for test organizations, I can provide these if they are required.

BrianMoyles (approved):

PR #1807, opened 2025-07-24
[Devtooling 1261] - Quality Forms Exported without QuestionGroups

BrianMoyles:
After speaking to a dev on the Quality Management team, they advised me that the getAll endpoint does not return the questionGroups attributes at all, and that a call to the getById endpoint is necessary to get these attributes.
The solution is to remove the caching functionality, so that the getById endpoint is called.
This issue was also found in the quality_forms_evaluation resource, and resolved.
Comparing Caching vs no Caching export showed minimal change in export time


carnellj-genesys (approved):

HemanthDogiparthi12:
@BrianMoyles can you resolve merge conflicts.

PR #1872, opened 2025-09-04
[Devtooling-1368] 'outbound_digitalruleset' fix missing properties on export

fergalmcl:


genesyscloud/outbound_digitalruleset/resource_genesyscloud_outbound_digitalruleset_schema.go
```diff
@@ -41,14 +41,15 @@ var (
 				Type:        schema.TypeString,
 			},
 			`value`: {
-				Description: `The value to compare against the contact's data.`,
-				Required:    true,
+				Description: `The value to compare against the contact's data. If 'value_type' is Numeric or Period, value is required`,
```
bbbco, line 44: Is this logic being enforced by the API? Should you also add the logic to the schema to enable fast fail if it is not configured properly (i.e., value_type = 'Numeric' but value not set)?

fergalmcl: Yes the logic is being enforced by the API, thats a good idea I added validation logic now so it will fail with a config error.

genesyscloud/outbound_digitalruleset/resource_genesyscloud_outbound_digitalruleset.go
```diff
@@ -48,6 +48,10 @@ func createOutboundDigitalruleset(ctx context.Context, d *schema.ResourceData, m
 	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
 	proxy := getOutboundDigitalrulesetProxy(sdkConfig)
 
+	if err := validateDigitalRulesetData(d); err != nil {
```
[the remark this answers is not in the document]
fergalmcl: With ValidateDiagFunc i'm not able to access other attributes so when checking 'value_type', I can't also check 'value'. I have changed the validation to be a CustomizeDiff function which will fail fast on plan

HemanthDogiparthi12 (approved):

PR #1873, opened 2025-09-04
RULES-975: Add Decision Table Resources and data source to CX as Code

brlenn:
Hi folks, 

This PR will allow users to create a [Decision Table](https://help.mypurecloud.com/articles/about-rule-based-decisions/) in Genesys Cloud.

As there is a lot to this resource, for now, users will only be able to create empty table with columns that are defined by a business rule schema. 

We have follow up ticket so that we can add rows, publish tables and create new versions. I am available to discuss any aspects of this PR or Rule-Based Decisions in general. 

I have executed ACC tests against my own org with correct permissions and also checked that tests are skipped when get all endpoint returns an non 200, I have included the fix from this [PR](https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/1845) so that all Business rules tests are passing.


genesyscloud/business_rules_decision_table/genesyscloud_business_rules_decision_table_proxy.go
```diff
@@ -0,0 +1,196 @@
...
+	resp, err := p.businessRulesApi.DeleteBusinessrulesDecisiontable(tableId, false)
+	if err == nil {
+		// Remove from cache after successful deletion
+		rc.DeleteCacheItem(p.BusinessRulesDecisionTableCache, tableId)
+	}
+	return resp, err
+}
+
+func getAllBusinessRulesDecisionTablesFn(ctx context.Context, p *BusinessRulesDecisionTableProxy) (*platformclientv2.Decisiontablelisting, *platformclientv2.APIResponse, error) {
+	return p.businessRulesApi.GetBusinessrulesDecisiontables("", "", nil, "")
```
BrianMoyles, line 165: This endpoint takes in a `pageSize` and `after` variable. as of right now if there is a pageSize limit, i believe this would only return the first result (eg first 100 resources) and not check for others. 
I think pageSize should be set to 100 unless this is not the size limit, and after checked. this functionality can be seen in the task_management_workbin_proxy file

genesyscloud/business_rules_decision_table/genesyscloud_business_rules_decision_table_proxy.go
```diff
@@ -0,0 +1,196 @@
...
+func deleteBusinessRulesDecisionTableFn(ctx context.Context, p *BusinessRulesDecisionTableProxy, tableId string) (*platformclientv2.APIResponse, error) {
+	resp, err := p.businessRulesApi.DeleteBusinessrulesDecisiontable(tableId, false)
+	if err == nil {
+		// Remove from cache after successful deletion
+		rc.DeleteCacheItem(p.BusinessRulesDecisionTableCache, tableId)
+	}
+	return resp, err
+}
+
+func getAllBusinessRulesDecisionTablesFn(ctx context.Context, p *BusinessRulesDecisionTableProxy) (*platformclientv2.Decisiontablelisting, *platformclientv2.APIResponse, error) {
```
BrianMoyles, line 164: name could be passed in to this function to send the getAll request with the name parameter. an example of this can be seen in the routing_skill_group_proxy. the functionality of the getBusinessRulesDecisionTablesByNameFn may have to be adjusted slightly with this, and an empty `""` passed in where we dont want to filter by a specific name

BrianMoyles (changes requested):

genesyscloud/business_rules_decision_table/genesyscloud_business_rules_decision_table_proxy.go
brlenn: Ok will take a look at this

shrutisuryawanshigenesys (approved):

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_utils.go
```diff
@@ -0,0 +1,430 @@
...
+import (
+	"context"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v165/platformclientv2"
+)
+
+// Queue lookup interface for dependency injection during testing
+type QueueLookupProvider interface {
```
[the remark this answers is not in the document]
brlenn: This QueueLookupProvider  is actually not required its unused so I will remove it, I have moved out the schema lookup

Issue #1891, opened 2025-09-10
Unable to use RuleSet Action Type: APPEND_CUSTOM_ENTRY_TO_DNC_LIST


jgec1981:
_Using  mypurecloud/genesyscloud v1.68.5_

_When trying to add the RuleSet:_

resource "genesyscloud_outbound_ruleset" "add_to_dnc_list" {
name                  = "Add to DNC List"
rules {
name = "Add to Custom Entry to DNC List"
order = 0
category = "DIALER_WRAPUP"
conditions {
type              = "systemDispositionCondition"
inverted          = false
codes             = [ "ININ-OUTBOUND-FAILED-TO-REACH-AGENT" ]
}
actions {
type = "Action"
**action_type_name = "APPEND_CUSTOM_ENTRY_TO_DNC_LIST"**
properties = {
"Action" = "APPEND_CUSTOM_ENTRY_TO_DNC_LIST",
"neverExpire" = "false",
"expirationDuration" = "P3DT0H0M"
} } } }

_An Error occurs:_

**Error: expected** rules.0.actions.0.**action_type_name to be one of**
["DO_NOT_DIAL" "MODIFY_CONTACT_ATTRIBUTE" "SWITCH_TO_PREVIEW" "APPEND_NUMBER_TO_DNC_LIST" "SCHEDULE_CALLBACK" "CONTACT_UNCALLABLE" "NUMBER_UNCALLABLE" "SET_CALLER_ID" "SET_SKILLS" "DATA_ACTION"], 
**got APPEND_CUSTOM_ENTRY_TO_DNC_LIST**

with module.add_to_dnc_list_ruleset.genesyscloud_outbound_ruleset.add_to_dnc_list,
on main.tf line 14, in resource "genesyscloud_outbound_ruleset" "add_to_dnc_list":
│   14:       **action_type_name      = "APPEND_CUSTOM_ENTRY_TO_DNC_LIST"**

PR #1894, opened 2025-09-10
[No Jira] Added missing action_type_name

BrianMoyles:
Resolves https://github.com/MyPureCloud/terraform-provider-genesyscloud/issues/1891

Attribute was not in the `ValidateFunc`
Added it to a test to verify API handles it without failure
configuration in issue was able to be created after changes made


fergalmcl (approved):

jgec1981:
Is there an indicative date for these changes to be released?

PR #1933, opened 2025-09-29
Fix DEVTOOLING-1408 Don't prevent creating a script resource if it has a matching name

bbbco:
Don't prevent creating a script resource if it has a matching name (the UI and API permit this)


HemanthDogiparthi12 (commented):
Are there any regression test cases that effect this ? if yes can you change them too.

PR #1943, opened 2025-10-06
No Jira Knowledge Tests

BrianMoyles:
This PR adds rebuild functionality to the delete context of knowledgebases. 
Currently, if the dependencies are not complete, it will fail to delete any knowledgebase.
The Dependencies are rebuilt before delete is attempted to prevent this error

### DIFF
--- a/genesyscloud/knowledge_knowledgebase/resource_genesyscloud_knowledge_knowledgebase.go
+++ b/genesyscloud/knowledge_knowledgebase/resource_genesyscloud_knowledge_knowledgebase.go
@@ -132,11 +132,63 @@ func updateKnowledgeKnowledgebase(ctx context.Context, d *schema.ResourceData, m
 	return readKnowledgeKnowledgebase(ctx, d, meta)
 }
 
+func rebuildDatabase() diag.Diagnostics {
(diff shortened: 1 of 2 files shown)


genesyscloud/knowledge_knowledgebase/resource_genesyscloud_knowledge_knowledgebase.go
```diff
@@ -132,12 +132,64 @@ func updateKnowledgeKnowledgebase(ctx context.Context, d *schema.ResourceData, m
 	return readKnowledgeKnowledgebase(ctx, d, meta)
 }
 
+func rebuildDatabase() diag.Diagnostics {
+	log.Printf("Rebuilding knowledge base database")
+
+	architectApi := platformclientv2.NewArchitectApi()
+
+	resp, architectError := architectApi.PostArchitectDependencytrackingBuild()
```
fergalmcl, line 140: Since this takes a while it might be good to make a GET call first and check if the status is `OPERATIONALNEEDSREBUILD`, `NOTBUILT` or `BUILDINCOMPLETE` and only continue if so

PR #1982, opened 2025-11-03
[Devtooling-1167] Error when changing queue from bullseye to standard routing

fergalmcl:
When removing bullseye rings from a routing queue if the bullseye rings have members in any ring (users linked to group, skillgroup or team in `member_groups`) then deleting them all at once to switch to standard routing will cause the API to throw an error. 
To get around this, when switching from bullseye to standard routing an SDK call will be made to remove the`member_groups` from the rings before the main SDK update call to remove the rings entirely.

### DIFF
--- a/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
+++ b/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
@@ -1049,7 +1049,6 @@ func clearBullseyeRingMemberGroups(ctx context.Context, d *schema.ResourceData, 
 	if err != nil {
 		return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to get queue %s error: %s", d.Id(), err), resp)
 	}
-
 	if currentQueue.Bullseye == nil || currentQueue.Bullseye.Rings == nil {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -1044,6 +1044,49 @@ func flattenQueueWrapupCodes(ctx context.Context, queueID string, proxy *Routing
...
+	updateQueue.Bullseye = &platformclientv2.Bullseye{
+		Rings: &clearedRings,
+	}
+
+	_, resp, err = proxy.updateRoutingQueue(ctx, d.Id(), updateQueue)
+	if err != nil {
+		return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to clear member_groups from bullseye rings in queue %s error: %s", d.Id(), err), resp)
+	}
+
+	updateQueue.Bullseye = nil
```
HemanthDogiparthi12: in case of no bulls eye update - after removing the members, should you re-update the queue to have no bulleye

HemanthDogiparthi12: sounds good. I see the re update that will be done as part of the usual update-context and not part of this method. that clarifies my question.

PR #1983, opened 2025-11-05
Feature/ai studio summary

mcphee11:
I added into the project teh Ai Studio Summary Settings which gives the abilty to add custom Copilot Summaries to cahnge the formatting of the summaries generated

### DIFF
--- a/genesyscloud/ai_studio_summary_setting/data_source_genesyscloud_ai_studio_summary_setting_test.go
+++ b/genesyscloud/ai_studio_summary_setting/data_source_genesyscloud_ai_studio_summary_setting_test.go
@@ -20,7 +20,6 @@ func TestAccDataSourceAiStudioSummarySetting(t *testing.T) {
 		aiStudioSummarySettingDataLabel = "data-aiStudioSummarySetting"
 
 		aiStudioSummarySettingResourceLabel = "resource-aiStudioSummarySetting"
-		resourceName                        = "test-summary-setting"
 		name                                = "summary setting test"
 		language                            = "en-au"
 		summaryType                         = "Concise"
@@ -39,7 +38,7 @@ func TestAccDataSourceAiStudioSummarySetting(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				// Create Summary Setting
-				Config: GenerateFullAiStudioSummarySettingResource(aiStudioSummarySettingResourceLabel, name, language, summaryType, settingType, format, maskPii, participantLabelInternal, participantLabelExternal, predefinedInsights0, predefinedInsights1, prompt) + generateAiStudioSummarySettingDataSource(aiStudioSummarySettingDataLabel, resourceName, "genesyscloud_ai_studio_summary_setting."+aiStudioSummarySettingResourceLabel),
+				Config: GenerateFullAiStudioSummarySettingResource(aiStudioSummarySettingResourceLabel, name, language, summaryType, settingType, format, maskPii, participantLabelInternal, participantLabelExternal, predefinedInsights0, predefinedInsights1, prompt) + generateAiStudioSummarySettingDataSource(aiStudioSummarySettingDataLabel, name, "genesyscloud_ai_studio_summary_setting."+aiStudioSummarySettingResourceLabel),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttrPair(
 						"data.genesyscloud_ai_studio_summary_setting."+aiStudioSummarySettingDataLabel, "id",
--- a/genesyscloud/ai_studio_summary_setting/genesyscloud_ai_studio_summary_setting_proxy.go
+++ b/genesyscloud/ai_studio_summary_setting/genesyscloud_ai_studio_summary_setting_proxy.go
@@ -109,7 +109,7 @@ func getAllAiStudioSummarySettingFn(ctx context.Context, p *aiStudioSummarySetti
 	var allSummarySettings []platformclientv2.Summarysetting
 	const pageSize = 100
 
-	summarySettings, resp, err := p.aIStudioApi.GetConversationsSummariesSettings(name, "", "", "", 1, pageSize)
+	summarySettings, resp, err := p.aIStudioApi.GetConversationsSummariesSettings("", name, "", "", 1, pageSize)
 	if err != nil {
 		return nil, resp, err
 	}
@@ -122,7 +122,7 @@ func getAllAiStudioSummarySettingFn(ctx context.Context, p *aiStudioSummarySetti
 	}
 
 	for pageNum := 2; pageNum <= *summarySettings.PageCount; pageNum++ {
-		summarySettings, _, err := p.aIStudioApi.GetConversationsSummariesSettings(name, "", "", "", pageNum, pageSize)
+		summarySettings, _, err := p.aIStudioApi.GetConversationsSummariesSettings("", name, "", "", pageNum, pageSize)
 		if err != nil {
 			return nil, resp, err
 		}


genesyscloud/aistudio_summary_setting/genesyscloud_aistudio_summary_setting_proxy.go
```diff
@@ -0,0 +1,168 @@
...
+	return p.deleteAistudioSummarySettingAttr(ctx, p, id)
+}
+
+// createAistudioSummarySettingFn is an implementation function for creating a Genesys Cloud aistudio summary setting
+func createAistudioSummarySettingFn(ctx context.Context, p *aistudioSummarySettingProxy, aistudioSummarySetting *platformclientv2.Summarysetting) (*platformclientv2.Summarysetting, *platformclientv2.APIResponse, error) {
+	return p.aIStudioApi.PostConversationsSummariesSettings(*aistudioSummarySetting)
+}
+
+// getAllAistudioSummarySettingFn is the implementation for retrieving all aistudio summary setting in Genesys Cloud
+func getAllAistudioSummarySettingFn(ctx context.Context, p *aistudioSummarySettingProxy) (*[]platformclientv2.Summarysetting, *platformclientv2.APIResponse, error) {
```
BrianMoyles: name could be passed in here as a parameter. The getByNameFn could then call this function instead of duplicating code, and any function call to return all resources can pass in an empty string. This is the standard used in other resources which can be used as a reference

BrianMoyles (changes requested):
I also wonder would the naming convention be better as `ai_studio` rather than `aistudio` to align more with the naming conventions of the other resources

genesyscloud/aistudio_summary_setting/genesyscloud_aistudio_summary_setting_proxy.go
mcphee11: I have completed these changes in my branch

BrianMoyles (changes requested):

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_schema.go
```diff
@@ -0,0 +1,168 @@
...
+				Description: `Name of the summary setting.`,
+				Required:    true,
+				Type:        schema.TypeString,
+			},
+			`language`: {
+				Description: `Language of the generated summary, e.g. en-US, it-IT.`,
+				Required:    true,
+				Type:        schema.TypeString,
+			},
+			`summary_type`: {
```
fergalmcl: On API Explorer I can see that the fields:

- `summaryType`
- `format`
- `settingType`

have only certain values that can be set. You could use a `ValidateFunc` on these attributes here to ensure they are set appropriately:
```go
validation.StringInSlice()
```

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_schema.go
```diff
@@ -0,0 +1,168 @@
...
+				Elem:        summarySettingPIIResource,
+			},
+			`participant_labels`: {
+				Description: `How to refer to interaction participants in the generated summary.`,
+				Optional:    true,
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Elem:        summarySettingParticipantLabelsResource,
+			},
+			`predefined_insights`: {
```
fergalmcl: This attribute can be validated with the same function as above. However, as this is a list of strings you can do the validation inside the nested schema in `Elem`

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,112 @@
...
+		summaryType                   = "Concise"
+		settingType                   = "Basic"
+		format                        = "BulletPoints"
+		maskPii                       = "true"
+		participantLabelInternal      = "Advisor"
+		participantLabelExternal      = "Member"
+		predefinedInsightsLabel       = "label1"
+		predefinedInsightsDescription = "description1"
+		prompt                        = "Summaries should be no more then 300 characters and include 3 dot points of key information"
+		externalSystemUrl             = "https://externalsystemurl.com"
```
fergalmcl: same as with the data source test

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,112 @@
...
+		setting_type     = "%s"
+		format           = "%s"
+		mask_p_i_i {
+		all = %s
+		}
+		participant_labels {
+			internal = "%s"
+			external = "%s"
+		}
+		predefined_insights {
```
fergalmcl: This test configuration is invalid. Since `predefined_insights` is a list of string and not a list of schema types, this needs to be changed to:
```tf
predefined_insights = [ "%s", "%s" ]
```

genesyscloud/ai_studio_summary_setting/data_source_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,61 @@
...
+						"data.genesyscloud_ai_studio_summary_setting."+aiStudioSummarySettingDataLabel, "id",
+						"genesyscloud_ai_studio_summary_setting."+aiStudioSummarySettingResourceLabel, "id",
+					),
+				),
+			},
+		},
+	})
+}
+
+func generateAiStudioSummarySettingDataSource(resresourceName string, resourceName string, dependsOn string) string {
```
fergalmcl: there is a typo the resourceName argument

genesyscloud/ai_studio_summary_setting/data_source_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestAccDataSourceAiStudioSummarySetting(t *testing.T) {
+	//t.Parallel()
+	var (
+		aiStudioSummarySettingDataLabel = "data-aiStudioSummarySetting"
+
+		aiStudioSummarySettingResourceLabel = "resource-aiStudioSummarySetting"
+		resourceName                        = "test-summary-setting"
+		name                                = "summary setting test"
+		language                            = "en-au"
```
fergalmcl: The language in both data source and resource tests is causing an API error. It seems the languages need to be upper case for the country i.e 'en-AU'

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_utils.go
```diff
@@ -0,0 +1,154 @@
...
+		ParticipantLabels:  buildSummarySettingParticipantLabelss(d.Get("participant_labels").([]interface{})),
+		PredefinedInsights: lists.BuildSdkStringListFromInterfaceArray(d, "predefined_insights"),
+		CustomEntities:     buildSummarySettingCustomEntitys(d.Get("custom_entities").([]interface{})),
+		SettingType:        platformclientv2.String(d.Get("setting_type").(string)),
+		Prompt:             platformclientv2.String(d.Get("prompt").(string)),
+	}
+}
+
+// buildSummarySettingPIIs maps an []interface{} into a Genesys Cloud *[]platformclientv2.Summarysettingpii
+func buildSummarySettingPIIs(summarySettingPIIs []interface{}) *platformclientv2.Summarysettingpii {
```
fergalmcl: The 2 'build' util functions here (PII and ParticipantLabels) are all panicking with nil pointer exceptions in the tests for this resource from the return statement here. Since they are a list with a max items of 1 you don't need to loop over them.

You can return nil early if the block isn't found in the config like this:
```go
func buildSummarySettingPIIs(summarySettingPIIs []interface{}) *platformclientv2.Summarysettingpii {
	if len(summarySettingPIIs) == 0 {
		return nil
	}

	summarySettingPIIsMap, ok := summarySettingPIIs[0].(map[string]interface{})
	if !ok {
		return nil
	}

	return &platformclientv2.Summarysettingpii{
		All: resourcedata.GetNillableValueFromMap[bool](summarySettingPIIsMap, "all", false),
	}
}
```

This will need to be done for the function below too

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_utils.go
```diff
@@ -0,0 +1,154 @@
...
+The resource_genesyscloud_ai_studio_summary_setting_utils.go file contains various helper methods to marshal
+and unmarshal data into formats consumable by Terraform and/or Genesys Cloud.
+*/
+
+// getAiStudioSummarySettingFromResourceData maps data from schema ResourceData object to a platformclientv2.Summarysetting
+func getAiStudioSummarySettingFromResourceData(d *schema.ResourceData) platformclientv2.Summarysetting {
+	return platformclientv2.Summarysetting{
+		Name:               platformclientv2.String(d.Get("name").(string)),
+		Language:           platformclientv2.String(d.Get("language").(string)),
+		SummaryType:        platformclientv2.String(d.Get("summary_type").(string)),
```
fergalmcl: Since summary_type is optional if it is omitted d.Get will read the empty value as and empty string. This causes the API to throw an error 
`API Error: 400 - Value [] is not valid for field type [SummarySettingSumaryType]`

Instead you should use the util function below to only get the value if it is set in the config
```go
resourcedata.GetNonZeroPointer
```

edit: the same is true for `format` and `settings_type` so they need to be updated too

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,112 @@
...
+			{
+				// Update
+				Config: GenerateFullAiStudioSummarySettingResource(resourceName, name, language, summaryType, settingType, format, maskPii, participantLabelInternal, participantLabelExternal, predefinedInsightsLabel, predefinedInsightsDescription, prompt, externalSystemUrl),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "name", name),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "language", language),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "summary_type", summaryType),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "setting_type", settingType),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "format", format),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "mask_p_i_i", maskPii),
```
fergalmcl: This check need to be on the key "mask_p_i_i.0.all" since it is a list

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,112 @@
...
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "participant_labels.internal", participantLabelInternal),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "participant_labels.external", participantLabelExternal),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "predefined_insights.0.label", predefinedInsightsLabel),
```
fergalmcl: The `predefined_insights` is a list of strings so you need to access them sequentially e.g. predefined_insights.0, predefined_insights.1

The label and description key can be removed

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_schema.go
mcphee11: I have updated these items with the ValidateFunc

genesyscloud/ai_studio_summary_setting/data_source_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,61 @@
...
+		summaryType                         = "Concise"
+		settingType                         = "Basic"
+		format                              = "BulletPoints"
+		maskPii                             = "true"
+		participantLabelInternal            = "Advisor"
+		participantLabelExternal            = "Member"
+		predefinedInsightsLabel             = "label1"
+		predefinedInsightsDescription       = "description1"
+		prompt                              = "Summaries should be no more then 300 characters and include 3 dot points of key information"
+		externalSystemUrl                   = "https://externalsystemurl.com"
```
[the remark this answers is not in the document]
mcphee11: done also fixed naming typo in generateAiStudioSummarySettingDataSource

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_test.go
mcphee11: updated in test

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_schema.go
mcphee11: I now see what you meant i was looking at the wrong thing i have added this in a lower commit for the tests to align

genesyscloud/ai_studio_summary_setting/data_source_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,61 @@
...
+		resourceName                        = "test-summary-setting"
+		name                                = "summary setting test"
+		language                            = "en-au"
+		summaryType                         = "Concise"
+		settingType                         = "Basic"
+		format                              = "BulletPoints"
+		maskPii                             = "true"
+		participantLabelInternal            = "Advisor"
+		participantLabelExternal            = "Member"
+		predefinedInsightsLabel             = "label1"
```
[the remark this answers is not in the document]
mcphee11: i think i have this right but not 100% this might needs some more work it also aligns with the commetn above on the formatting fo this in the scheam file...

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_utils.go
mcphee11: I also added "prompt" in the same format as that is also a non required string value.

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_test.go
```diff
@@ -0,0 +1,112 @@
...
+				// Update
+				Config: GenerateFullAiStudioSummarySettingResource(resourceName, name, language, summaryType, settingType, format, maskPii, participantLabelInternal, participantLabelExternal, predefinedInsightsLabel, predefinedInsightsDescription, prompt, externalSystemUrl),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "name", name),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "language", language),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "summary_type", summaryType),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "setting_type", settingType),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "format", format),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "mask_p_i_i", maskPii),
+					resource.TestCheckResourceAttr("genesyscloud_ai_studio_summary_setting."+resourceName, "participant_labels.internal", participantLabelInternal),
```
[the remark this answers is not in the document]
mcphee11: fixed item in test

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_test.go
mcphee11: fixed item in test

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_schema.go
```diff
@@ -0,0 +1,178 @@
...
+				Description: `Set which insights to include in the generated summary by default.`,
+				Optional:    true,
+				Type:        schema.TypeSet,
```
fergalmcl: Changing this to a TypeSet will require you to modify the logic for setting this in the resource data.

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting_utils.go
```diff
@@ -0,0 +1,138 @@
...
+// getAiStudioSummarySettingFromResourceData maps data from schema ResourceData object to a platformclientv2.Summarysetting
+func getAiStudioSummarySettingFromResourceData(d *schema.ResourceData) platformclientv2.Summarysetting {
+	return platformclientv2.Summarysetting{
+		Name:               platformclientv2.String(d.Get("name").(string)),
+		Language:           platformclientv2.String(d.Get("language").(string)),
+		SummaryType:        resourcedata.GetNonZeroPointer[string](d, "summary_type"),
+		Format:             resourcedata.GetNonZeroPointer[string](d, "format"),
+		MaskPII:            buildSummarySettingPIIs(d.Get("mask_p_i_i").([]interface{})),
+		ParticipantLabels:  buildSummarySettingParticipantLabelss(d.Get("participant_labels").([]interface{})),
+		PredefinedInsights: lists.BuildSdkStringListFromInterfaceArray(d, "predefined_insights"),
```
fergalmcl: This needs to be updated to work with TypeSet. you could use:
```go
PredefinedInsights: lists.SetToStringList(d.Get("predefined_insights").(*schema.Set))
```

genesyscloud/ai_studio_summary_setting/resource_genesyscloud_ai_studio_summary_setting.go
```diff
@@ -0,0 +1,148 @@
...
+				return retry.RetryableError(util.BuildWithRetriesApiDiagnosticError(ResourceType, fmt.Sprintf("Failed to read ai studio summary setting %s: %s", d.Id(), getErr), resp))
+			}
+			return retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(ResourceType, fmt.Sprintf("Failed to read ai studio summary setting %s: %s", d.Id(), getErr), resp))
+		}
+
+		resourcedata.SetNillableValue(d, "name", summarySetting.Name)
+		resourcedata.SetNillableValue(d, "language", summarySetting.Language)
+		resourcedata.SetNillableValue(d, "summary_type", summarySetting.SummaryType)
+		resourcedata.SetNillableValue(d, "format", summarySetting.Format)
+		resourcedata.SetNillableValueWithInterfaceArrayWithFunc(d, "mask_p_i_i", summarySetting.MaskPII, func(item *platformclientv2.Summarysettingpii) []interface{} {
```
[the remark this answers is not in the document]
mcphee11: cleaned up and removed nil checks

genesyscloud/ai_studio_summary_setting/genesyscloud_ai_studio_summary_setting_proxy.go
```diff
@@ -0,0 +1,179 @@
...
+func createAiStudioSummarySettingFn(ctx context.Context, p *aiStudioSummarySettingProxy, aiStudioSummarySetting *platformclientv2.Summarysetting) (*platformclientv2.Summarysetting, *platformclientv2.APIResponse, error) {
+	return p.aIStudioApi.PostConversationsSummariesSettings(*aiStudioSummarySetting)
+}
+
+// getAllAiStudioSummarySettingFn is the implementation for retrieving all ai studio summary setting in Genesys Cloud
+func getAllAiStudioSummarySettingFn(ctx context.Context, p *aiStudioSummarySettingProxy, name string) (*[]platformclientv2.Summarysetting, *platformclientv2.APIResponse, error) {
+	var allSummarySettings []platformclientv2.Summarysetting
+	const pageSize = 100
+
+	summarySettings, resp, err := p.aIStudioApi.GetConversationsSummariesSettings(name, "", "", "", 1, pageSize)
```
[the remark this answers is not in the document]
mcphee11: I understand you point but i can't get the exporter to throw this error in my environment as it is? i have cleared everything and run a fresh exporter with configs in my ORG but i get them totally fine... I cant simualte this error so i dont want to change this to beign raw restful unless i have to. I have had to do this in the past in the GO SDK for other things. Lets have a chat about this if possible. ill ping you on GC

PR #2010, opened 2025-11-12
Intitial module - add WFM business units terraform configuration

StormeHawke:


HemanthDogiparthi12:
can you rebase this and resolve the conflicts.

StormeHawke:
> can you rebase this and resolve the conflicts.

Figured it out, had to rebase off of upstream/dev. Should be right now

genesyscloud/util/util_retries.go
```diff
@@ -196,3 +203,80 @@ func IsTimeoutError(errDiag error) bool {
...
+		now := time.Now()
+		delay := httpDate.Sub(now)
+		if delay > 0 {
+			return delay, true
+		}
+		return 0, false
+	}
+
+	// Try parsing as RFC 822 format (alternative date format)
+	if httpDate, err := time.Parse(time.RFC822, retryAfterHeader); err == nil {
```
DavidRigglemanININ: There seems to be a lot of duplication here. The code seems identical except for the format. Could you not just make a loop that iterates through the 3 formats.

genesyscloud/workforcemanagement_businessunits/genesyscloud_workforcemanagement_businessunits_proxy.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// createWorkforcemanagementBusinessunitsFn is an implementation function for creating a Genesys Cloud workforcemanagement businessunits
+func createWorkforcemanagementBusinessunitsFn(ctx context.Context, p *workforcemanagementBusinessunitsProxy, workforcemanagementBusinessunits *platformclientv2.Createbusinessunitrequest) (*platformclientv2.Businessunitresponse, *platformclientv2.APIResponse, error) {
+	return p.workforceManagementApi.PostWorkforcemanagementBusinessunits(*workforcemanagementBusinessunits, false)
+}
+
+// getAllWorkforcemanagementBusinessunitsFn is the implementation for retrieving all workforcemanagement businessunits in Genesys Cloud
+func getAllWorkforcemanagementBusinessunitsFn(ctx context.Context, p *workforcemanagementBusinessunitsProxy) (*[]platformclientv2.Businessunitlistitem, *platformclientv2.APIResponse, error) {
+	businessUnitResponses, resp, err := p.workforceManagementApi.GetWorkforcemanagementBusinessunits("", "")
```
DavidRigglemanININ: why the empty strings? Again, not too familiar with go, but is there the concept of optional parameters or default parameters? And is null not a first class thing kotlin (and why you are using empty string)

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
```diff
@@ -0,0 +1,128 @@
...
+	proxy := getWorkforcemanagementBusinessunitsProxy(sdkConfig)
+	cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, ResourceWorkforcemanagementBusinessunits(), constants.ConsistencyChecks(), resourceName)
+
+	log.Printf("Reading workforcemanagement businessunits %s", d.Id())
+
+	return util.WithRetriesForRead(ctx, d, func() *retry.RetryError {
+		businessUnitResponse, resp, getErr := proxy.getWorkforcemanagementBusinessunitsById(ctx, d.Id())
+		if getErr != nil {
+			if util.IsStatus404(resp) {
+				return util.RetryableErrorWithRetryAfter(ctx, util.BuildWithRetriesApiDiagnosticError(resourceName, fmt.Sprintf("Failed to read workforcemanagement businessunits %s: %s", d.Id(), getErr), resp), resp)
```
DavidRigglemanININ: Am I reading this wrong, or why are we saying that a 404 is a retryable error?

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_schema.go
```diff
@@ -0,0 +1,224 @@
...
+
+/*
+resource_genesycloud_workforcemanagement_businessunits_schema.go holds four functions within it:
+
+1.  The registration code that registers the Datasource, Resource and Exporter for the package.
+2.  The resource schema definitions for the workforcemanagement_businessunits resource.
+3.  The datasource schema definitions for the workforcemanagement_businessunits datasource.
+4.  The resource exporter configuration for the workforcemanagement_businessunits exporter.
+*/
+const resourceName = "genesyscloud_workforcemanagement_businessunits"
```
bbbco: Set `resourceName` to `ResourceName` so it can be referenced by other resource packages.

genesyscloud/workforcemanagement_businessunits/genesyscloud_workforcemanagement_businessunits_proxy.go
StormeHawke: The two parameters being passed are `feature string, divisionId string`. as mentioned previously, strings can't be null in Go. Further note: This is a call into the actual generated Go sdk so can't edit the method to take pointers

genesyscloud/workforcemanagement_businessunits/genesyscloud_workforcemanagement_businessunits_proxy.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return businessUnitResponses.Entities, resp, nil
+}
+
+// getWorkforcemanagementBusinessunitsIdByNameFn is an implementation of the function to get a Genesys Cloud workforcemanagement businessunits by name
+func getWorkforcemanagementBusinessunitsIdByNameFn(ctx context.Context, p *workforcemanagementBusinessunitsProxy, name string) (string, *platformclientv2.APIResponse, bool, error) {
```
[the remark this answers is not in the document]
StormeHawke: yes, commit incoming

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
```diff
@@ -0,0 +1,128 @@
...
+		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to create workforcemanagement businessunits: %s", err), resp)
+	}
+
+	d.SetId(*businessUnitResponse.Id)
+	log.Printf("Created workforcemanagement businessunits %s", *businessUnitResponse.Id)
+	return readWorkforcemanagementBusinessunits(ctx, d, meta)
+}
+
+// readWorkforcemanagementBusinessunits is used by the workforcemanagement_businessunits resource to read a workforcemanagement businessunits from genesys cloud
+func readWorkforcemanagementBusinessunits(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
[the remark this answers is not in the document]
StormeHawke: name update incoming in next commit

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
StormeHawke: updated in next commit

genesyscloud/util/util_retries.go
DavidRigglemanININ: I also didn't even realize that all these formats were valid? Is that more an HTTP spec thing and not a genesys platform thing? Because all of our public api doc only lists retry-after as seconds as the format, right? I mean RTA would probably completely ignore Retry-After if you called WFM service and I gave you a date timestamp back, right?

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_schema.go
```diff
@@ -0,0 +1,224 @@
...
+			},
+		},
+	}
+
+	schedulerMessageTypeSeverityResource := &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			`type`: {
+				Description: `The type of the message`,
+				Optional:    true,
+				Type:        schema.TypeString,
```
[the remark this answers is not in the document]
StormeHawke: It doesn't appear so. It looks like the SDK just has strings, and no convenient list of the valid values to validate. I can supply the valid values (63 of them, for message types) so that the terraform resource can validate... but that risks getting out of sync. Given there's no built in list of valid enum values, my inclination here is to let the API validate rather than risk having an out of sync resource if/when we add new values.

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_schema.go
```diff
@@ -0,0 +1,224 @@
...
+			},
+			`time_zone`: {
+				Description: `The time zone for this business unit, using the Olsen tz database format`,
+				Required:    true,
+				Type:        schema.TypeString,
+			},
+			`short_term_forecasting`: {
+				Description: `Short term forecasting settings`,
+				Optional:    true,
+				Type:        schema.TypeList,
```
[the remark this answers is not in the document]
StormeHawke: Optional objects are represented as TypeList with MaxItems: 1 in Terraform schema
This allows the field to be omitted (empty list) or present with one item (the object)

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_schema.go
```diff
@@ -0,0 +1,224 @@
...
+		},
+	}
+}
+
+// WorkforcemanagementBusinessunitsExporter returns the resourceExporter object used to hold the genesyscloud_workforcemanagement_businessunits exporter's config
+func WorkforcemanagementBusinessunitsExporter() *resourceExporter.ResourceExporter {
+	return &resourceExporter.ResourceExporter{
+		GetResourcesFunc: provider.GetAllWithPooledClient(getAllAuthWorkforcemanagementBusinessunits),
+		RefAttrs:         map[string]*resourceExporter.RefAttrSettings{
+			// TODO: Add any reference attributes here
```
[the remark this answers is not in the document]
StormeHawke: yes, apparently this needs to list references to external objects, in this case division.

genesyscloud/workforcemanagement_businessunits/genesyscloud_workforcemanagement_businessunits_proxy.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, resp, err
+	}
+
+	return businessUnitResponses.Entities, resp, nil
+}
+
+// getWorkforceManagementBusinessUnitIdByExactNameFn is an implementation of the function to get a Genesys Cloud workforce management business unit ID by exact name match
+func getWorkforceManagementBusinessUnitIdByExactNameFn(ctx context.Context, p *workforceManagementBusinessUnitsProxy, name string) (string, *platformclientv2.APIResponse, bool, error) {
+	// Empty strings for feature and divisionId parameters mean no filtering - retrieve all business units to search by name
+	businessUnitResponses, resp, err := p.workforceManagementApi.GetWorkforcemanagementBusinessunits("", "")
```
fergalmcl: Instead of duplicating the get all code you can call the function you already have in the proxy here i.e. 
```go
p.getAllWorkforceManagementBusinessUnits(ctx)
```

genesyscloud/util/util_retries.go
```diff
@@ -196,3 +203,80 @@ func IsTimeoutError(errDiag error) bool {
...
+// Retry-After can be either:
+//   - A number of seconds (integer)
+//   - An HTTP date (RFC 1123 format)
+func GetRetryAfterDelay(apiResponse *platformclientv2.APIResponse) (time.Duration, bool) {
+	if apiResponse == nil || apiResponse.Response == nil {
+		return 0, false
+	}
+
+	retryAfterHeader := apiResponse.Response.Header.Get("Retry-After")
+	if retryAfterHeader == "" {
```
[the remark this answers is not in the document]
StormeHawke: Thought that was covered by answers elsewhere. Strings can't be nil in go, they're primitives

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_schema.go
```diff
@@ -0,0 +1,226 @@
...
+		UpdateContext: provider.UpdateWithPooledClient(updateWorkforceManagementBusinessUnits),
+		DeleteContext: provider.DeleteWithPooledClient(deleteWorkforceManagementBusinessUnits),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+		Schema: map[string]*schema.Schema{
+			`name`: {
+				Description: ``,
+				Optional:    true,
```
fergalmcl: should these attributes not be required?

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_utils.go
```diff
@@ -0,0 +1,382 @@
...
+
+		buSchedulingSettingsResponsesSlice = append(buSchedulingSettingsResponsesSlice, sdkBuSchedulingSettingsResponse)
+	}
+
+	return &buSchedulingSettingsResponsesSlice[0]
+}
+
+
+// buildCreateBusinessUnitSettingsRequest maps an interface{} into a Genesys Cloud *platformclientv2.Createbusinessunitsettingsrequest
+func buildCreateBusinessUnitSettingsRequest(businessUnitSettingsResponses []interface{}) *platformclientv2.Createbusinessunitsettingsrequest {
```
fergalmcl: This function is causing the provider to panic is the settings is not defined in the config (one of the acceptance tests does this).

If the settings is supposed to be required this function should be fine, but otherwise there is no need to have a loop here since settings has a maxitems of 1. You can check the length of settings and return nil early instead. For example:
```go
func buildCreateBusinessUnitSettingsRequest(businessUnitSettingsResponses []interface{}) *platformclientv2.Createbusinessunitsettingsrequest {
	if len(businessUnitSettingsResponses) == 0 {
		return nil
	}

	businessUnitSettingsResponsesMap, ok := businessUnitSettingsResponses[0].(map[string]interface{})
	if !ok {
		return nil
	}

	var sdkBusinessUnitSettingsResponse platformclientv2.Createbusinessunitsettingsrequest
	
	resourcedata.BuildSDKStringValueIfNotNil(&sdkBusinessUnitSettingsResponse.StartDayOfWeek, businessUnitSettingsResponsesMap, "start_day_of_week")
	resourcedata.BuildSDKStringValueIfNotNil(&sdkBusinessUnitSettingsResponse.TimeZone, businessUnitSettingsResponsesMap, "time_zone")
	resourcedata.BuildSDKInterfaceArrayValueIfNotNil(&sdkBusinessUnitSettingsResponse.ShortTermForecasting, businessUnitSettingsResponsesMap, "short_term_forecasting", buildBuShortTermForecastingSettings)
	resourcedata.BuildSDKInterfaceArrayValueIfNotNil(&sdkBusinessUnitSettingsResponse.Scheduling, businessUnitSettingsResponsesMap, "scheduling", buildBuSchedulingSettings)

	return &sdkBusinessUnitSettingsResponse
}
```

the same logic can be applied to all other list types that have a maxitems of 1

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return util.BuildAPIDiagnosticError(resourceName, fmt.Sprintf("Failed to update workforcemanagement businessunits %s: %s", d.Id(), err), resp)
+	}
+
+	log.Printf("Updated workforcemanagement businessunits %s", *businessUnitResponse.Id)
+	return readWorkforcemanagementBusinessunits(ctx, d, meta)
+}
+
+// deleteWorkforcemanagementBusinessunits is used by the workforcemanagement_businessunits resource to delete a workforcemanagement businessunits from Genesys cloud
+func deleteWorkforcemanagementBusinessunits(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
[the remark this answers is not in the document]
StormeHawke: The customer does whenever they run a deploy/sync of their configured resources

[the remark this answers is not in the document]
DavidRigglemanININ: So do they update some config file and then basically run a single "sync" operation. In other words, when you add MUs in a later phase, will it be your responsibility to make sure the MU delete operations happen before the BU delete operations. But on the flip side, the BU create needs to happen before the MU create?

genesyscloud/util/util_retries.go
DavidRigglemanININ: Still waiting on answer to this. To me the top choices are aligning with platform standards or fully aligning with the spec. Although I still feel like all of this is unnecessary. This is only used on genesys APIs, right? If any API started returning Retry-After as something than an integer, it would seem like that should be viewed as a significant breaking change.

genesyscloud/util/util_retries.go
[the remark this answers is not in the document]
DavidRigglemanININ: It looks like all maps in go have the same behavior, which is what I was referring to. But I'm good to close this out

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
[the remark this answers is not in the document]
DavidRigglemanININ: You brought in the parallel of terraform, so I was assuming you were working off of a state file. But maybe that's not the case? In other words, at a super trivial level, I send the config I went to genesys cloud (formatting of course will be different)
```
BusinessUnit1
    ManagementUnit1
BusinessUnit2
    ManagementUnit2
```
Then I want to update my state file so I send you
```
BusinessUnit1
    ManagementUnit1
```
That should auto cleanup those resources (exactly how terraform or cloudformation works), but the key part that I'm bringing up is that to do this, the order of operations matters and was wondering if that had been considered as part of this PR. Or if I'm thinking about this all wrong and it's not at all like terraform.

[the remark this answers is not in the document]
StormeHawke: This is terraform, that's this whole project. 

We do have this thing where we don't include BU references in everything (I think the deepest we go with Bu references is MU level generally) so I'm not sure how references to upstream stuff would even work. 

I don't have an answer.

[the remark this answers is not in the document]
StormeHawke: I don't know, terraform supports what terraform supports.

I'm betting since the MU schema has references to business unit that will work, but activity codes, work plans (under MU), et al we intentionally didn't expose the bu/mu reference on the schema under the assumption the BU ID was available on the request path.

examples/resources/genesyscloud_workforcemanagement_businessunits/resource.tf
bbbco: Did you run `make testexamples` to ensure these examples pass? I don't think the `division_id` attribute will pass as is in your second example. Refer to the README.md file inside the examples directory. You can also take a look at the example for the `genesyscloud_team` resource for a simple example alongside the `locals.tf` file.

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
```diff
@@ -0,0 +1,131 @@
...
+	proxy := getWorkforceManagementBusinessUnitsProxy(sdkConfig)
+	cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, ResourceWorkforcemanagementBusinessunits(), constants.ConsistencyChecks(), ResourceName)
+
+	log.Printf("Reading workforce management business unit %s", d.Id())
+
+	return util.WithRetriesForRead(ctx, d, func() *retry.RetryError {
+		businessUnitResponse, resp, getErr := proxy.getWorkforceManagementBusinessUnitById(ctx, d.Id())
+		if getErr != nil {
+			// Retry on 409 (conflict) or 429 (rate limit) errors
+			if util.IsStatus409(resp) || (resp != nil && resp.StatusCode == http.StatusTooManyRequests) {
```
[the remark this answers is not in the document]
bbbco: The provider automagically handles 429 status responses with a retry. You shouldn't need the logic for `StatusTooManyRequests`

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_schema.go
```diff
@@ -0,0 +1,226 @@
...
+				Type:        schema.TypeInt,
+			},
+		},
+	}
+
+	schedulerMessageTypeSeverityResource := &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			// See API documentation for valid enum values: https://developer.mypurecloud.com/api/rest/v2/workforcemanagement/
+			`type`: {
+				Description: `The type of the message. Validation is handled by the API to avoid maintaining a potentially stale list of enum values. See API documentation for valid values: https://developer.mypurecloud.com/api/rest/v2/workforcemanagement/`,
```
bbbco: Update link to https://developer.genesys.cloud/useragentman/workforcemanagement/#post-api-v2-workforcemanagement-businessunits

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_schema.go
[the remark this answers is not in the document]
bbbco: Too bad the enum list couldn't be provided via the platform SDK.

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
```diff
@@ -0,0 +1,130 @@
...
+	proxy := getWorkforceManagementBusinessUnitsProxy(sdkConfig)
+	cc := consistency_checker.NewConsistencyCheck(ctx, d, meta, ResourceWorkforcemanagementBusinessunits(), constants.ConsistencyChecks(), ResourceName)
+
+	log.Printf("Reading workforce management business unit %s", d.Id())
+
+		return util.WithRetriesForRead(ctx, d, func() *retry.RetryError {
+			businessUnitResponse, resp, getErr := proxy.getWorkforceManagementBusinessUnitById(ctx, d.Id())
+			if getErr != nil {
+				// Retry on 409 (conflict) errors
+				if util.IsStatus409(resp) {
```
DavidRigglemanININ: Just going to drop a note in general that 409s in many cases are non-retryable errors. Specifically, if a version number is supplied and you get a 409, it will fail forever going forward no matter how many times you retry. Now if a version number isn't supplied, then a 409 may or may not work. So more pointing that out for general awareness. But in this case, it seems super weird to handle a 409 on a GET only operation (this looks to be a BU get) when that's not even something like the backend can throw from my knowledge. So what was the motivation specifically calling this out here?

DavidRigglemanININ (approved):

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits.go
```diff
@@ -0,0 +1,125 @@
...
+	proxy := newWorkforceManagementBusinessUnitsProxy(clientConfig)
+	resources := make(resourceExporter.ResourceIDMetaMap)
+
+	businessUnitResponses, resp, err := proxy.getAllWorkforceManagementBusinessUnits(ctx)
+	if err != nil {
+		return nil, util.BuildAPIDiagnosticError(ResourceName, fmt.Sprintf("Failed to get workforce management business units: %v", err), resp)
+	}
+
+	for _, businessUnitResponse := range *businessUnitResponses {
+		resources[*businessUnitResponse.Id] = &resourceExporter.ResourceMeta{BlockLabel: *businessUnitResponse.Name}
```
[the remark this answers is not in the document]
bbbco: Good deal! That's exactly what I was hoping to hear!

genesyscloud/workforcemanagement_businessunits/data_source_genesyscloud_workforcemanagement_businessunits.go
```diff
@@ -0,0 +1,42 @@
...
+
+	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
+		businessUnitResponseId, resp, retryable, err := proxy.getWorkforceManagementBusinessUnitIdByExactName(ctx, name)
+
+		if err != nil && !retryable {
+			return retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(ResourceName, fmt.Sprintf("Error searching workforce management business unit %s | error: %s", name, err), resp))
+		}
+
+		if retryable {
+			return util.RetryableErrorWithRetryAfter(ctx, util.BuildWithRetriesApiDiagnosticError(ResourceName, fmt.Sprintf("No workforce management business unit found with name %s", name), resp), resp)
```
HemanthDogiparthi12: modify this to ResourceType

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_utils.go
```diff
@@ -0,0 +1,382 @@
...
+
+	return &wfmServiceGoalImpactsSlice[0] // TODO is this right?
+}
+
+// buildWfmServiceGoalImpactSettings maps an []interface{} into a Genesys Cloud *[]platformclientv2.Wfmservicegoalimpactsettings
+func buildWfmServiceGoalImpactSettings(wfmServiceGoalImpactSettings []interface{}) *platformclientv2.Wfmservicegoalimpactsettings {
+	wfmServiceGoalImpactSettingsSlice := make([]platformclientv2.Wfmservicegoalimpactsettings, 0)
+	for _, wfmServiceGoalImpactSettings := range wfmServiceGoalImpactSettings {
+		var sdkWfmServiceGoalImpactSettings platformclientv2.Wfmservicegoalimpactsettings
+		wfmServiceGoalImpactSettingsMap, ok := wfmServiceGoalImpactSettings.(map[string]interface{})
```
HemanthDogiparthi12: do a length check and then do. 
wfmServiceGoalImpactSettingsMap, ok := wfmServiceGoalImpactSettings[0].(map[string]interface{})
	if !ok {
		return nil
	}
	
	you dont need to do "return &wfmServiceGoalImpactSettingsSlice[0] // TODO is this right "

StormeHawke: I think this comment might have been on the wrong line, but I think I addressed the issue above

genesyscloud/workforcemanagement_businessunits/resource_genesyscloud_workforcemanagement_businessunits_utils.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+
+		resourcedata.BuildSDKStringValueIfNotNil(&sdkBusinessUnitSettingsResponse.StartDayOfWeek, businessUnitSettingsResponsesMap, "start_day_of_week")
+		resourcedata.BuildSDKStringValueIfNotNil(&sdkBusinessUnitSettingsResponse.TimeZone, businessUnitSettingsResponsesMap, "time_zone")
+		resourcedata.BuildSDKInterfaceArrayValueIfNotNil(&sdkBusinessUnitSettingsResponse.ShortTermForecasting, businessUnitSettingsResponsesMap, "short_term_forecasting", buildBuShortTermForecastingSettings)
+		resourcedata.BuildSDKInterfaceArrayValueIfNotNil(&sdkBusinessUnitSettingsResponse.Scheduling, businessUnitSettingsResponsesMap, "scheduling", buildBuSchedulingSettings)
+
+		businessUnitSettingsResponsesSlice = append(businessUnitSettingsResponsesSlice, sdkBusinessUnitSettingsResponse)
+	}
+
```
HemanthDogiparthi12: settings is optional.  if there is no setting in resourcedata. &businessUnitSettingsResponsesSlice[0]  would give a panic

PR #2014, opened 2025-11-14
[Devtooling-1268] Consistency check timeout fix for updating quality forms survey

fergalmcl:
On a survey update when published is true, updates the Id to the new unpublished survey returned from publishing the current so that the read context reads the newest version


BrianMoyles (approved):

BrianMoyles:
looks good. there is just a conflict that needs to be resolved

PR #2050, opened 2025-12-03
NO-JIRA: Added the ability to annotate a resource and then extract it…

carnellj-genesys:
… into a report


HemanthDogiparthi12:
the changes for these are tracked and done in another PR. this PR can be closed.

PR #2098, opened 2026-01-14
PDNA-26733: add learning modules and users rules resources

reginald-cui-genesys:
JIRA ticket: https://inindca.atlassian.net/browse/PDNA-26733


genesyscloud/learning_modules/resource_genesyscloud_learning_modules.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to update learning module %s error: %s", *learningModule.Name, err), resp)
+	}
+
+	log.Printf("Updated learning module %s", *module.Id)
+
+	time.Sleep(2 * time.Second)
+
+	published := d.Get("is_published").(bool)
+	if published {
```
gus-silva-genesys: Same thing here as in https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/2098/files#r2736305434

gus-silva-genesys (approved):
Left a few questions, they've been answered. Approved 👍

genesyscloud/learning_modules/resource_genesyscloud_learning_modules_test.go
```diff
@@ -0,0 +1,242 @@
...
+		if rs.Type != ResourceType {
+			continue
+		}
+
+		learningModule, resp, err := learningApi.GetLearningModule(rs.Primary.ID, []string{})
+		if learningModule != nil {
+			continue
+		}
+
+		if learningModule != nil {
```
HemanthDogiparthi12: unreachable code.

HemanthDogiparthi12 (changes requested):

genesyscloud/users_rules/genesyscloud_users_rules_proxy.go
```diff
@@ -0,0 +1,226 @@
...
+			pageSize,
+			expand,
+			enabled,
+			searchTerm,
+			sortOrder,
+		)
+		if err != nil {
+			return nil, resp, fmt.Errorf("Failed to get users rules: %v", err)
+		}
+		if usersRules.Entities == nil || len(*usersRules.Entities) == 0 {
```
sureshperiyappan, line 135: If "Learning" type returns zero results, this `return` exits the entire function
without ever querying "ActivityPlan" rules. Should be `continue` instead of
`return &allUsersRules, resp, nil`.

genesyscloud/learning_modules/genesyscloud_learning_modules_proxy.go
```diff
@@ -0,0 +1,257 @@
...
+	return module, resp, nil
+}
+
+// getAllLearningModulesFn is the implementation for retrieving all learning modules in Genesys Cloud
+func getAllLearningModulesFn(ctx context.Context, p *learningModulesProxy, searchTerm string) (*[]platformclientv2.Learningmodule, *platformclientv2.APIResponse, error) {
+	// Set resource context for SDK debug logging
+	ctx = provider.EnsureResourceContext(ctx, ResourceType)
+
+	var allModules []platformclientv2.Learningmodule
+	const isArchived = true
```
sureshperiyappan: The isArchived parameter is set to `true`, which includes archived modules in results.
This means the exporter and data source will return archived modules. Is this intentional?
For normal CRUD and data source lookups, you'd typically want `false` to exclude archived
modules. If it's intentional, please add a comment explaining why.

genesyscloud/users_rules/resource_genesyscloud_users_rules_schema.go
```diff
@@ -0,0 +1,174 @@
...
+		ReadContext:   provider.ReadWithPooledClient(readUsersRules),
+		UpdateContext: provider.UpdateWithPooledClient(updateUsersRules),
+		DeleteContext: provider.DeleteWithPooledClient(deleteUsersRules),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
```
sureshperiyappan, line 124: Both `name` and `type` are marked Optional, but createUsersRules() unconditionally sends
them via platformclientv2.String(d.Get("name").(string)) and
platformclientv2.String(d.Get("type").(string)). If a user omits these, empty strings
are sent to the API. These should be Required: true since the create request needs them.

clarify this ?

genesyscloud/users_rules/resource_genesyscloud_users_rules_schema.go
```diff
@@ -0,0 +1,174 @@
...
+				Description: "The name of the rule",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"description": {
+				Description: "The description of the rule",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+			"type": {
```
sureshperiyappan, line 134: Both `name` and `type` are marked Optional, but createUsersRules() unconditionally sends
them via platformclientv2.String(d.Get("name").(string)) and
platformclientv2.String(d.Get("type").(string)). If a user omits these, empty strings
are sent to the API. These should be Required: true since the create request needs them.

clarify this ?

genesyscloud/users_rules/resource_genesyscloud_users_rules.go
```diff
@@ -0,0 +1,146 @@
...
+			d.Set("criteria", flattenUsersRulesCriteria(usersRule.Criteria))
+		}
+
+		log.Printf("Read users rule %s %s", d.Id(), *usersRule.Name)
+		return cc.CheckState(d)
+	})
+}
+
+// updateUsersRules is used by the users_rules resource to update a users rule in Genesys Cloud
+func updateUsersRules(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
sureshperiyappan, line 97: The update function builds Usersrulesupdaterulerequest with Name, Description, and
Criteria, but never sets VarType. If a user changes `type`, the change is silently
ignored. Either include `type` in the update payload, or add `ForceNew: true` to the
`type` field in the schema so Terraform knows to recreate the resource.

genesyscloud/users_rules/resource_genesyscloud_users_rules.go
```diff
@@ -0,0 +1,146 @@
...
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := getUsersRulesProxy(sdkConfig)
+
+	var usersRuleRequest platformclientv2.Usersrulesupdaterulerequest
+	name := d.Get("name").(string)
+	if name != "" {
+		usersRuleRequest.Name = &name
+	}
+	description := d.Get("description").(string)
+	if description != "" {
```
sureshperiyappan: If a user removes the description (sets it to ""), the update won't send it to the API,
so the old description persists server-side. The `if description != ""` guard prevents
clearing the field. Consider always setting it, or using a nil-aware approach to
distinguish "not configured" from "explicitly set to empty".

genesyscloud/users_rules/genesyscloud_users_rules_proxy.go
reginald-cui-genesys: Must have missed this when adding the for loop for the different types, should be fixed now, thanks!

genesyscloud/learning_modules/genesyscloud_learning_modules_proxy.go
reginald-cui-genesys: This is actually a deprecated field in the PAPI that no longer does anything, I'll update it to "false" for now though.

genesyscloud/users_rules/resource_genesyscloud_users_rules_schema.go
reginald-cui-genesys: Updated the necessary fields in the schema to be required, thanks.

genesyscloud/users_rules/resource_genesyscloud_users_rules.go
reginald-cui-genesys: Added ForceNew to the schema

genesyscloud/users_rules/resource_genesyscloud_users_rules.go
reginald-cui-genesys: Changed to use hasChange to determine whether to send description in the request.

sureshperiyappan:
The generated docs at users_rules.md show name, type, and criteria as Optional, but the schema code has them as Required. The docs need to be regenerated:

go generate
This is likely because the docs were generated from an earlier version of the schema before these fields were made Required.

genesyscloud/learning_modules/resource_genesyscloud_learning_modules_schema.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	return &schema.Resource{
+		Description: `Genesys Cloud learning modules`,
+
+		CreateContext: provider.CreateWithPooledClient(createLearningModule),
+		ReadContext:   provider.ReadWithPooledClient(readLearningModule),
+		UpdateContext: provider.UpdateWithPooledClient(updateLearningModule),
+		DeleteContext: provider.DeleteWithPooledClient(deleteLearningModule),
+		Importer: &schema.ResourceImporter{
```
sureshperiyappan, line 304: Unnecessary Importer block on Data Sources

Both DataSourceLearningModules() and DataSourceUsersRules() include an Importer block:

Importer: &schema.ResourceImporter{
    StateContext: schema.ImportStatePassthroughContext,
},
Data sources don't support import — only resources do. No other data source in the project has this. It should be removed from both.

genesyscloud/learning_modules/resource_genesyscloud_learning_modules_schema.go
```diff
@@ -0,0 +1,419 @@
...
+
+/*
+ResourceName is defined in this file along with four functions:
+
+1.  The registration code that registers the Datasource, Resource and Exporter for the package.
+2.  The resource schema definitions for the learning_modules resource.
+3.  The datasource schema definitions for the learning_modules datasource.
+4.  The resource exporter configuration for the learning_modules exporter.
+*/
+const ResourceName = "genesyscloud_learning_modules"
```
sureshperiyappan: ResourceName + ResourceType = ResourceName pattern

The new resources define:

const ResourceName = "genesyscloud_learning_modules"
const ResourceType = ResourceName
The vast majority of existing resources use just:

const ResourceType = "genesyscloud_auth_division"
This isn't wrong, but it's inconsistent with the rest of the codebase. The ResourceName constant is unused outside the schema file — SetRegistrar could just use ResourceType directly. Minor, non-blocking.

reginald-cui-genesys:
> The generated docs at users_rules.md show name, type, and criteria as Optional, but the schema code has them as Required. The docs need to be regenerated:
> 
> go generate This is likely because the docs were generated from an earlier version of the schema before these fields were made Required.

Missed this, thanks! Regenerated the docs

genesyscloud/learning_modules/resource_genesyscloud_learning_modules_schema.go
reginald-cui-genesys: Removed the unnecessary import blocks

genesyscloud/learning_modules/resource_genesyscloud_learning_modules_schema.go
reginald-cui-genesys: Removed the unnecessary ResourceName vars

sureshperiyappan (approved):

sanjaysg21 (approved):

PR #2101, opened 2026-01-16
Devtooling 1513

sanjaysg21:


tools/resource_metadata_exporter/commands.go
```diff
@@ -0,0 +1,409 @@
...
+		return exportCSV(cliMetadata, output)
+	default:
+		return fmt.Errorf("unsupported format: %s", exportFormat)
+	}
+}
+
+func runValidate(cmd *cobra.Command, args []string) error {
+	fmt.Printf("Validating metadata in: %s\n", validatePath)
+
+	// Mock validation results
```
fergalmcl: The validate function is still using test data for the valid and invalid resources. It could also be good to log a warning if any of the other annotations are missing if they are not required.

PR #2103, opened 2026-01-21
[FIX] automatic_time_zone_mapping returns false even when set to true

valentinhegedusgenesys:


bbbco (changes requested):
Some concerns about this implementation:

1. This is a workaround, but does not root cause fix - This is treating a symptom of an API inconsistency rather than fixing the underlying issue. See 4.

2. Asymmetric logic - Only handles the false → true case. What if API returns true but config has false?

3. Deviates from standard patterns - Other attributes in the same file use simple d.Set() calls. This special-casing could confuse future maintainers

4. API behavior unclear - Is this a bug in the Genesys Cloud API? Should it be reported upstream?

5. Testing - No test changes included. Should have acceptance tests validating this behavior

6. Documentation - No comments explaining WHY this special handling is needed

HemanthDogiparthi12 (approved):

HemanthDogiparthi12 (changes requested):
this requires further analysis.

can you log the request object's timezonemapping attribute. (during create context)
and when you retrieve the resource object back from the api, (during get context) what is the autoamtictimezonemapping that is logged.

it could be a mutating api.

and based on the log statements , we have to see if we need to approach the concerned api team

@valentinhegedusgenesys

sureshperiyappan:
closing the PR. if require we can take it later with JIRA ticket

PR #2123, opened 2026-02-03
MRMO-44 document permissions and scopes

timsmithgenesys:
This change adds permission and scope information for resources. It looks at the API endpoints specified in the `apis.md` file for each terraform resource and looks up the necessary permissions and scopes for those endpoints. It then appends that information in the resource doc page. Additionally, the permissions and scopes are dumped into a `public/data/resource_permissions-{version}.json` file for programmatic consumption.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,5 @@ test/temp/*
 
 public/data/dependency_tree*
 public/data/resource_permissions*
+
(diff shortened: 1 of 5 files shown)


bbbco:
this is simply amazing!

carnellj-genesys (approved):
Like a boss tim.

PR #2136, opened 2026-02-10
Created Resource Group Greeting

daragh-kearney:


genesyscloud/group_greeting/genesyscloud_group_greeting_proxy.go
```diff
@@ -0,0 +1,173 @@
...
+		getAllGreetingsAttr: getAllGreetingsFn,
+		createGreetingAttr:  createGroupGreetingFn,
+		getGreetingByIdAttr: getGroupGreetingByIdFn,
+		updateGreetingAttr:  updateGroupGreetingFn,
+		deleteGreetingAttr:  deleteGroupGreetingFn,
+	}
+}
+
+func getGreeetingProxy(clientConfig *platformclientv2.Configuration) *greetingProxy {
+	if internalProxy == nil {
```
HemanthDogiparthi12: do this via sync.Once

genesyscloud/group_greeting/resource_genesyscloud_group_greeting_schema.go
```diff
@@ -0,0 +1,105 @@
...
+				Description: "Greeting audio TTS.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+		},
+	}
+}
+
+func GreetingExporter() *resourceExporter.ResourceExporter {
+	return &resourceExporter.ResourceExporter{
```
bbbco, line 112: There is no handling of exporting the audio file. We normally do this for such resources (see user prompt for example).

I did notice that the organization greeting has the same problem. If you fix it here with a shared greeting media exporter function you could easily add it to that resource type as well

daragh-kearney: Think everything you mentioned should be addressed now, let me know if anything needs to be changed

sureshperiyappan (approved):

PR #2149, opened 2026-02-17
Fix: Don't send IdleTokenTimeoutSeconds when timeout is disabled

valentinhegedusgenesys:
When enable_idle_token_timeout is false, set IdleTokenTimeoutSeconds to nil to avoid API validation errors. Added unit tests for disabled timeout scenarios.


docs/data-sources/integration_webhookmake si.md
HemanthDogiparthi12: is this. a misplaces commit ?

sureshperiyappan (approved):

PR #2183, opened 2026-03-03
DEVTOOLING-1585 - Provider crash when applying multiple genesyscloud_tf_export resources at once

valentinhegedusgenesys:
Added a global mutex (attributesDecodedMutex) to protect the attributesDecoded map. Modified postProcessHclBytes() to create a copy of the map under mutex protection and iterate over the copy instead of the original.

### DIFF
--- a/genesyscloud/tfexporter/json_exporter.go
+++ b/genesyscloud/tfexporter/json_exporter.go
@@ -305,6 +305,20 @@ func writeConfig(jsonMap map[string]interface{}, path string) diag.Diagnostics {
 
 func postProcessJsonBytes(resource []byte) []byte {
 	resourceStr := string(resource)
+
+	// Create a copy of attributesDecoded map with mutex protection to avoid concurrent map iteration and map write
+	attributesDecodedMutex.Lock()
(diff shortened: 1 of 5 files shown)


HemanthDogiparthi12 (commented):
Hi @valentinhegedusgenesys  we have two exporters once hcl and other json. You might need to add a similar change for json exporter too.

bbbco (approved):

genesyscloud/tfexporter/json_exporter.go
```diff
@@ -305,6 +305,11 @@ func writeConfig(jsonMap map[string]interface{}, path string) diag.Diagnostics {
 
 func postProcessJsonBytes(resource []byte) []byte {
 	resourceStr := string(resource)
+
+	for placeholderId, val := range attributesDecoded {
+		resourceStr = strings.Replace(resourceStr, fmt.Sprintf("\"%s\"", placeholderId), val, -1)
```
bbbco, line 310: It looks like this block was copy/pasted from the HCL log, but I don't believe this processing is required for the JSON output (only the HCL). Please remove the copy pasta.

PR #2185, opened 2026-03-04
OBR-213 Add advanced processing with condition groups and execution settings

danilkulakov:


genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_schema.go
```diff
@@ -336,9 +369,8 @@ func ResourceOutboundCampaignrule() *schema.Resource {
 				Elem:        outboundCampaignRuleEntities,
 			},
 			`campaign_rule_conditions`: {
```
sureshperiyappan, line 398: There's no validation enforcing the relationship between campaign_rule_processing, campaign_rule_conditions, and condition_groups. Currently a user could set campaign_rule_processing = "v2" with campaign_rule_conditions instead of condition_groups, provide both simultaneously, or omit both entirely all of which would pass terraform plan and only fail at API time.

At minimum, consider adding ConflictsWith between campaign_rule_conditions and condition_groups to prevent users from setting both. For the conditional requirement (which one is needed depends on campaign_rule_processing), either a CustomizeDiff for plan-time validation or clear documentation noting the dependency would work. What does the API error look like when the wrong combination is sent? If it's clear enough, relying on the API may be acceptable for now.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule.go
```diff
@@ -101,16 +101,15 @@ func readOutboundCampaignRule(ctx context.Context, d *schema.ResourceData, meta
...
-		if campaignRule.CampaignRuleActions != nil {
-			d.Set("campaign_rule_actions", flattenCampaignRuleAction(campaignRule.CampaignRuleActions, flattenCampaignRuleActionEntities))
-		} else {
-			d.Set("campaign_rule_actions", nil)
-		}
+		resourcedata.SetNillableValueWithInterfaceArrayWithFunc(d, "campaign_rule_actions", campaignRule.CampaignRuleActions, func(actions *[]platformclientv2.Campaignruleaction) []interface{} {
+			return flattenCampaignRuleAction(actions, flattenCampaignRuleActionEntities)
+		})
 		resourcedata.SetNillableValue(d, "match_any_conditions", campaignRule.MatchAnyConditions)
+		resourcedata.SetNillableValueWithInterfaceArrayWithFunc(d, "execution_settings", campaignRule.ExecutionSettings, flattenExecutionSettings)
```
sureshperiyappan, line 112: SetNillableValueWithInterfaceArrayWithFunc has signature func[T any](*T, func(*T) []interface{}). This works if campaignRule.ExecutionSettings is *Campaignruleexecutionsettings, but please verify the SDK type matches this assumption. If the SDK field is a different shape (e.g. a slice pointer), this will panic at runtime.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_schema.go
```diff
@@ -354,6 +386,25 @@ func ResourceOutboundCampaignrule() *schema.Resource {
 				Default:     false,
 				Type:        schema.TypeBool,
 			},
+			`campaign_rule_processing`: {
+				Description:  `Campaign rule processing algorithm. Use "v2" to enable condition groups.`,
+				Optional:     true,
+				Type:         schema.TypeString,
+				ValidateFunc: validation.StringInSlice([]string{"", "v2"}, true),
```
sureshperiyappan: Including "" in the allowed values means a user can explicitly write campaign_rule_processing = "" in their config without getting a validation error. Since the field is Optional, omitting it already produces a zero-value the empty string in the slice is redundant and could mask accidental misconfiguration. Suggest changing to validation.StringInSlice([]string{"v2"}, true).

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_utils.go
```diff
@@ -14,16 +14,38 @@ func getCampaignruleFromResourceData(d *schema.ResourceData) platformclientv2.Ca
...
+	if v, ok := d.GetOk("campaign_rule_processing"); ok && v.(string) != "" {
+		campaignRule.CampaignRuleProcessing = platformclientv2.String(v.(string))
+	}
+	if v, ok := d.GetOk("condition_groups"); ok && len(v.([]interface{})) > 0 {
+		campaignRule.ConditionGroups = buildCampaignRuleConditionGroups(v.([]interface{}))
+	}
+	if v, ok := d.GetOk("campaign_rule_conditions"); ok && len(v.([]interface{})) > 0 {
+		campaignRule.CampaignRuleConditions = buildCampaignRuleConditions(v.([]interface{}))
+	}
+	if v, ok := d.GetOk("execution_settings"); ok && len(v.([]interface{})) > 0 {
```
sureshperiyappan, line 33: Terraform can produce nil as the first element of an optional list block. If v.([]interface{})[0] is nil, the type assertion to map[string]interface{} will panic. Please add a nil guard on list[0] before the type assertion.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule.go
danilkulakov: The types are correct: `campaignRule.ExecutionSettings` is `*platformclientv2.Campaignruleexecutionsettings` and `flattenExecutionSettings` has signature `func(*platformclientv2.Campaignruleexecutionsettings) []interface{}`.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_schema.go
danilkulakov: Added ConflictsWith on both fields and CustomizeDiff

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_utils.go
danilkulakov: Added nil guard

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_schema.go
danilkulakov: Removed the empty string from allowed values.

sureshperiyappan (approved):

PR #2190, opened 2026-03-10
DEVTOOLING-1611 - replace_with_datasource now works without "::"

daragh-kearney:
- Exporter fix: replace_with_datasource now accepts type-only entries like "genesyscloud_auth_division" and treats them as match-all for that type (equivalent to "genesyscloud_auth_division::") so :: is no longer needed now.
- Tests: added unit test cases covering the new type-only matching behavior.
- Docs: updated export guide/schema docs to document type == type::, and regenerated docs which also updated routing_email_route permissions.


bbbco:
Love this. @scheblein will love it too I'm sure!

sureshperiyappan (approved):

PR #2191, opened 2026-03-10
OBR-866: Add diagnostics_settings block to outbound_campaign resource

obolobova:


genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_test.go
```diff
@@ -1211,6 +1218,87 @@ func TestAccResourceOutboundCampaignPower(t *testing.T) {
 	})
 }
 
+// TestAccResourceOutboundCampaignDiagnosticsSettingsInvalidDialingMode tests that diagnostics_settings fails validation for unsupported dialing modes
```
sureshperiyappan: The CustomizeDiff validation in validateDiagnosticsSettingsDialingMode explicitly allows both power and predictive, but the only positive test is TestAccResourceOutboundCampaignPower which uses power mode. There's no test that verifies diagnostics_settings actually works end-to-end with predictive.

The negative test here only covers agentless. Consider adding at least one test step (or a new test function) that creates a predictive campaign with diagnostics_settings and asserts the value is read back correctly. Without it, we're trusting that the API handles diagnostics_settings identically for both modes but never verifying it

sureshperiyappan (approved):

PR #2220, opened 2026-03-20
Feature/add content location responseasset

jojoger:
Exposed the content_location attribute as a computed field to allow users to retrieve the URL of a response asset. Also added TestCheckResourceAttrSet to the acceptance tests. Reason for implementation was to have an environment independent Terraform File to create Canned Responses with Asset URLS in  tags


sureshperiyappan (approved):

jojoger:
As this is my first pull request for terraform-provider-genesyscloud, is it normal, that acceptance needs that long? When can we expect it to be accepted and merged to main?

PR #2222, opened 2026-03-23
DEVTOOLING-1607 - genesyscloud_quality_forms_evaluation

daragh-kearney:
topic_ids are now references instead of raw GUIDs, improving cross-org/region portability.


genesyscloud/speechandtextanalytics_topic/resource_genesyscloud_speechandtextanalytics_topic.go
```diff
@@ -0,0 +1,120 @@
+package speechandtextanalytics_topic
```
shrutisuryawanshigenesys, line 1: tests missing for speechandtextanalytics_topic resource

genesyscloud/speechandtextanalytics_topic/data_source_genesyscloud_speechandtextanalytics_topic.go
```diff
@@ -0,0 +1,43 @@
...
+
+func dataSourceSpeechAndTextAnalyticsTopicRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	name := d.Get("name").(string)
+	dialect := d.Get("dialect").(string)
+
+	sdkConfig := meta.(*provider.ProviderMeta).ClientConfig
+	proxy := getSttTopicProxy(sdkConfig)
+
+	const pageSize = 100
+	for pageNum := 1; pageNum <= 50; pageNum++ {
```
shrutisuryawanshigenesys: Is there a reason for capping this at 50 pages? This could miss data for larger orgs, might be better to loop until API returns no more results.

genesyscloud/speechandtextanalytics_topic/resource_genesyscloud_speechandtextanalytics_topic_exporter.go
```diff
@@ -0,0 +1,48 @@
...
+	resourceExporter "github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/resource_exporter"
+	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util"
+)
+
+func getAllTopics(ctx context.Context, clientConfig *platformclientv2.Configuration) (resourceExporter.ResourceIDMetaMap, diag.Diagnostics) {
+	proxy := getSttTopicProxy(clientConfig)
+	resources := make(resourceExporter.ResourceIDMetaMap)
+
+	const pageSize = 100
+	for pageNum := 1; pageNum <= 50; pageNum++ {
```
shrutisuryawanshigenesys: Same here

genesyscloud/speechandtextanalytics_topic/resource_genesyscloud_speechandtextanalytics_topic_utils.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	if topic.Published != nil {
+		_ = d.Set("published", *topic.Published)
+	}
+
+	return nil
+}
+
+func waitForPublishJob(ctx context.Context, proxy *sttTopicProxy, jobId string, timeout time.Duration) error {
```
sureshperiyappan: This hand-rolled polling loop works, but the rest of the codebase uses util.WithRetries for similar async polling (e.g. webdeployments_deployment uses it for deployment activation, telephony_providers_edges_trunkbasesettings for delete verification, etc.). Consider refactoring to use util.WithRetries for consistency. Not blocking, just a suggestion to align with established patterns.

genesyscloud/speechandtextanalytics_topic/resource_genesyscloud_speechandtextanalytics_topic.go
```diff
@@ -0,0 +1,120 @@
...
+					return diag.Errorf("%s", err)
+				}
+			}
+		}
+	}
+
+	return readTopic(ctx, d, meta)
+}
+
+func deleteTopic(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
sureshperiyappan, line 108: The sibling resource speechandtextanalytics_dictionaryfeedback fires the DELETE and then polls with util.WithRetries for up to 180s until it gets a 404 confirming the resource is actually gone. This resource just fires the DELETE and returns immediately. If the Genesys Cloud API is eventually consistent, a subsequent terraform plan or dependent resource operation could still find the topic alive. Please add delete verification consistent with the sibling resource:


return util.WithRetries(ctx, 180*time.Second, func() *retry.RetryError {
    _, resp, err := proxy.getTopic(ctx, d.Id())
    if err != nil {
        if util.IsStatus404(resp) {
            log.Printf("Deleted speech and text analytics topic %s", d.Id())
            return nil
        }
        return retry.NonRetryableError(util.BuildWithRetriesApiDiagnosticError(ResourceType, fmt.Sprintf("Error verifying deletion of topic %s: %s", d.Id(), err), resp))
    }
    return retry.RetryableError(util.BuildWithRetriesApiDiagnosticError(ResourceType, fmt.Sprintf("speech and text analytics topic %s still exists", d.Id()), resp))
})

genesyscloud/resource_exporter/resource_exporter_custom.go
```diff
@@ -12,6 +12,47 @@ import (
...
+
+	label := strings.ReplaceAll(fmt.Sprintf("%s_%s", name, dialect), " ", "_")
+	ds := map[string]interface{}{
+		"name":    name,
+		"dialect": dialect,
+	}
+
+	// If a map is provided, set a conventional key for callers that want to use it.
+	// (Some resolvers mutate a specific attribute key; array handlers will build the reference string directly.)
+	if configMap != nil {
```
sureshperiyappan: This block is dead code. The only call site that handles topic_ids (which is an array attribute) is sanitizeConfigArray, and it passes nil as the configMap:

dsType, dsLabel, dsConfig, resolve := refAttrCustomResolver.ResolveToDataSourceFunc(nil, strVal, sdkConfig)
The other call site (resolveValueToDataSource) passes a real configMap, but that handles scalar string attributes — not arrays. And even there, nothing downstream reads _resolved_topic_datasource since it's not a schema attribute. Compare with OutboundCampaignAgentScriptResolver which sets configMap["script_id"] — an actual schema field that gets written to the export.

Please check this block of code once again. to avoid confusion for future maintainers.

genesyscloud/speechandtextanalytics_topic/resource_genesyscloud_speechandtextanalytics_topic.go
```diff
@@ -0,0 +1,120 @@
...
+
+	req := buildTopicRequest(d)
+	log.Printf("Creating Speech & Text Analytics Topic %s", d.Get("name").(string))
+	topic, resp, err := proxy.createTopic(ctx, req)
+	if err != nil {
+		input, _ := util.InterfaceToJson(*req)
+		return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to create speech and text analytics topic: %s\n(input: %+v)", err, input), resp)
+	}
+
+	if topic != nil && topic.Id != nil {
```
sureshperiyappan: If the API returns err == nil but topic is nil or topic.Id is nil, d.SetId() is never called and the resource ID stays empty. The code then proceeds to publishTopics(ctx, []string{d.Id()}) and readTopic() with an empty ID, which will produce confusing downstream errors. Same issue exists in updateTopic at line 82. Add an early return with a diagnostic error if the ID isn't set:

if topic == nil || topic.Id == nil {
    return diag.Errorf("API returned success but no topic ID for %s", d.Get("name").(string))
}
d.SetId(*topic.Id)

shrutisuryawanshigenesys (approved):

PR #2233, opened 2026-03-24
Fixes exporter output bug when outputting both data and resources using the split by resource option

bbbco:
### Problem

When exporting with `split_files_by_resource = true` and `export_format = "json"`, if a resource type was exported as both a resource and a data source (via `replace_with_datasource`), the data source JSON file would overwrite the resource JSON file. This happened because both used the same filename: `{resource_type}.tf.json`.

For example, if `genesyscloud_routing_queue` had some instances exported as resources and others replaced with data sources, both would write to `genesyscloud_routing_queue.tf.json`, with the second write silently overwriting the first.

### Root Cause

In `json_exporter.go`, the data source file path was generated using the same pattern as the resource file path:

```go
resourceJSONFilePath := filepath.Join(j.dirPath, fmt.Sprintf("%s.%s", resType, resourceJSONFileExt))
```

### Fix

Prefixed data source filenames with `data_` to avoid collisions:

```go
resourceJSONFilePath := filepath.Join(j.dirPath, fmt.Sprintf("data_%s.%s", resType, resourceJSONFileExt))
```

This produces filenames like `data_genesyscloud_routing_queue.tf.json` for data sources, keeping them separate from the resource file `genesyscloud_routing_queue.tf.json`.

### Test improvements

- Updated `TestAccResourceTfExportSplitFilesAsJSON` to export resources alongside data sources (using `replace_with_datasource`) and validate that both resource and data source files are created with correct content
- Added `ValidateJSONFileKeyValue` to `util/test_utils.go` — a generic, reusable test helper that validates a value at an arbitrary dot-delimited path within a JSON file, replacing the previous `validateExportedResourceAttributeValue` which was hardcoded to the `resource.{type}.{label}` path structure
- Moved `LoadJsonFileToMap` from the test file to `util/util_json.go` so it can be shared across packages
- Registered `genesyscloud_auth_division` and `genesyscloud_routing_queue` data sources in the test init to support the new test scenarios
- Fixed a copy-paste bug in the test where `test-queue-2` was using `test-queue-1-` as its exported label prefix

---

Note: This was found during MEC1 migration, but NOT needed for a hotfix

### DIFF
--- a/genesyscloud/util/util_json.go
+++ b/genesyscloud/util/util_json.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"log"
+	"os"
 	"reflect"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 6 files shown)


genesyscloud/util/util_json.go
```diff
@@ -121,3 +123,22 @@ func MapToJson(m *map[string]interface{}) (string, error) {
...
+
+	jsonFile, err := os.Open(filename)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open export file at path %s: %v", filename, err)
+	}
+	defer func(jsonFile *os.File) {
+		_ = jsonFile.Close()
+	}(jsonFile)
+
+	byteValue, _ := io.ReadAll(jsonFile)
```
sureshperiyappan: The error returned by io.ReadAll is assigned to _ and silently discarded. Now that this is a shared exported utility, a failed read would produce a confusing unmarshal error downstream instead of surfacing the actual I/O issue. Should be handled.

sureshperiyappan (approved):

PR #2249, opened 2026-03-28
NO-JIRA Fix: User-defined exclude_attributes silently ignored during dependency resolution exports

bbbco:
### Problem

When using `enable_dependency_resolution = true` (which sets `addDependsOn`), the `exclude_attributes` configuration was silently ignored for resource types discovered through dependency resolution.

For example, with this config:

```hcl
resource "genesyscloud_tf_export" "export" {
  include_filter_resources_by_id = var.include_filter_resources_by_id
  exclude_attributes = [
    "genesyscloud_flow.file_content_hash",
    "genesyscloud_routing_queue.outbound_messaging_sms_address_id",
    "genesyscloud_routing_queue.outbound_email_address",
    "genesyscloud_routing_queue.outbound_messaging_open_messaging_recipient_id",
    "genesyscloud_routing_queue.outbound_messaging_whatsapp_recipient_id"
  ]
  enable_dependency_resolution = true
}
```

All five exclusions were logged as warnings and never applied:

```
Ignoring exclude attribute file_content_hash on genesyscloud_flow resources. Since exporter is not retrieved
Ignoring exclude attribute outbound_messaging_sms_address_id on genesyscloud_routing_queue resources. Since exporter is not retrieved
...
```

This caused the excluded attributes to appear in the exported config, breaking downstream consumers.

### Root Cause

The old `populateConfigExcluded` method was called inside `retrieveExporters`, which runs multiple times during the export lifecycle. On the first call, the exporters map only contained resource types matching the user's `include_filter_resources_by_id` filter. Resource types like `genesyscloud_flow` and `genesyscloud_routing_queue` that were later discovered through dependency resolution were not yet in the map, so their exclusions were silently skipped.

When `retrieveExporters` was called again during `rebuildExports` (as part of dependency resolution), it fetched fresh exporter instances from the global registry. Even though the resource types were now present, the complex cycle of deep-copying, rebuilding, and merging exporters meant the `ExcludedAttributes` set during one pass could be lost or never applied to the exporter instances that were ultimately used during `sanitizeConfigMap`.

### Fix

Replaced the fragile approach of trying to attach exclusions to exporter instances at the right time with a single final pass that applies user-defined `exclude_attributes` directly to the finalized config maps, right before output file generation.

**Removed:**
- `populateConfigExcluded` method and its call in `retrieveExporters`
- `ExcludedAttributes` merging logic in `mergeExporters`

**Added:**
- `removeUserDefinedExcludedAttributesFromConfigMaps` — called once in `Export()` after all resource data is finalized (between step 6 and step 7). Parses `exclude_attributes` from the Terraform config and nils out matching attributes directly on the final `resourceTypesMaps`. Supports exact type match, regex patterns, and nested attribute paths.

Hardcoded `ExcludedAttributes` defined in resource schemas (e.g. `webdeployments_deployment` excluding `configuration.version`) are unaffected — they continue to be handled by `IsAttributeExcluded` during `sanitizeConfigMap` as before.

No impact to MRMO or BCP exporters — neither uses `populateConfigExcluded` or `exclude_attributes`.

### Testing

- Updated `TestUnitTfExportTestExcludeAttributes` to test the new `removeExcludedAttrsFromMap` function
- All existing unit tests pass

### Release

- Suggested release with next main release (1.78.0)
- @HemanthDogiparthi12

### DIFF
--- a/genesyscloud/tfexporter/genesyscloud_resource_exporter.go
+++ b/genesyscloud/tfexporter/genesyscloud_resource_exporter.go
@@ -352,7 +352,7 @@ func (g *GenesysCloudResourceExporter) Export() (diagErr diag.Diagnostics) {
 	// (enable_dependency_resolution=true), resource types may be discovered
 	// after the initial populateConfigExcluded call, meaning their exclusions
 	// were never applied to the exporter instances used during sanitization.
-	g.removeUserDefinedExcludedAttributesFromConfigMaps()
+	diagErr = append(diagErr, g.removeUserDefinedExcludedAttributesFromConfigMaps()...)
 
 	// Step #7 Write the terraform state file along with either the HCL or JSON
 	diagErr = append(diagErr, g.generateOutputFiles()...)
@@ -2551,18 +2551,19 @@ func (g *GenesysCloudResourceExporter) sanitizeConfigArray(
 // This ensures exclusions work regardless of when a resource type was discovered
 // during dependency resolution. Hardcoded ExcludedAttributes defined in resource
 // schemas are handled separately by IsAttributeExcluded during sanitizeConfigMap.
-func (g *GenesysCloudResourceExporter) removeUserDefinedExcludedAttributesFromConfigMaps() {
+func (g *GenesysCloudResourceExporter) removeUserDefinedExcludedAttributesFromConfigMaps() (diags diag.Diagnostics) {
 	excludedAttrs, ok := g.d.GetOk("exclude_attributes")
 	if !ok {
-		return
+		return diag.Errorf("Error retrieving \"exclude_attributes\" value: %v", ok)
 	}
 
 	// Parse exclude_attributes into a map of resourceType -> []attributes
-	exclusionsByType := make(map[string][]string) // exact type -> attrs
+	exclusionsByType := make(map[string][]string)  // exact type -> attrs
 	exclusionPatterns := make(map[string][]string) // regex pattern -> attrs
 	for _, excluded := range lists.InterfaceListToStrings(excludedAttrs.([]interface{})) {
 		resourceIdx := strings.Index(excluded, ".")
 		if resourceIdx == -1 || len(excluded) == resourceIdx {
+			tflog.Warn(g.ctx, fmt.Sprintf("Skipping invalid exclude_attributes entry (missing attribute path): %s", excluded))
 			continue
 		}
 		resourceTypePattern := excluded[:resourceIdx]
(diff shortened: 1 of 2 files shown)


genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -2552,50 +2546,88 @@ func (g *GenesysCloudResourceExporter) sanitizeConfigArray(
...
+
+	// Parse exclude_attributes into a map of resourceType -> []attributes
+	exclusionsByType := make(map[string][]string) // exact type -> attrs
+	exclusionPatterns := make(map[string][]string) // regex pattern -> attrs
+	for _, excluded := range lists.InterfaceListToStrings(excludedAttrs.([]interface{})) {
 		resourceIdx := strings.Index(excluded, ".")
-		if resourceIdx == -1 {
-			return diag.Errorf("Invalid excluded_attribute %s", excluded)
+		if resourceIdx == -1 || len(excluded) == resourceIdx {
+			continue
```
sureshperiyappan, line 2567: These silently skipped entries could frustrate users debugging why an exclusion isn't working. 
Consider adding a tflog.Warn here, e.g.:

tflog.Warn(g.ctx, fmt.Sprintf("Skipping invalid exclude_attributes entry (missing attribute path): %s", excluded))

Not a blocker — just a one-liner that saves someone a debugging headache down the road.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -2552,50 +2546,88 @@ func (g *GenesysCloudResourceExporter) sanitizeConfigArray(
...
-					exporter1.AddExcludedAttribute(excludedAttr)
-					tflog.Info(g.ctx, fmt.Sprintf("Excluding attribute %s on %s resources.", excludedAttr, resourceTypePattern))
-					matchFound = true
-					continue
-				}
+// getExcludedAttrsForType returns the list of attribute paths to exclude for a given resource type.
+func (g *GenesysCloudResourceExporter) getExcludedAttrsForType(resourceType string, exactMap map[string][]string, patternMap map[string][]string) []string {
+	attrs := exactMap[resourceType]
+	for pattern, patternAttrs := range patternMap {
+		if match, _ := regexp.MatchString(pattern, resourceType); match {
```
sureshperiyappan: Suggestion: regexp.MatchString returns an error as its second value and it's being 
discarded with `_`. If a user provides an invalid regex pattern in exclude_attributes, 
it silently fails to match anything with no feedback. Worth either:

- Pre-compiling with regexp.Compile() upfront and logging a warning for invalid patterns, or
- At minimum checking the error here and logging it

The performance angle (MatchString recompiles on every call) is minor given typical 
export sizes, but pre-compiling would address both issues in one shot.

sureshperiyappan (approved):

PR #2252, opened 2026-03-31
Rules 1399 Update CX as code to use bulk operations

Jx-hou:


docs/resources/business_rules_decision_table.md
```diff
@@ -36,9 +36,6 @@ The following permissions are required to use this resource:
 * `businessrules:decisionTable:publish`
 * `businessrules:decisionTable:search`
 * `businessrules:decisionTable:view`
-* `businessrules:decisionTableRow:add`
```
brlenn: these permissions should be still required?

brlenn, line 39: These are required permission for bulk row ops

brlenn:
Can you add a description?

genesyscloud/business_rules_decision_table/decision_table_bulk_http.go
```diff
@@ -0,0 +1,123 @@
...
+	Outputs *map[string]platformclientv2.Decisiontablerowparametervalue `json:"outputs,omitempty"`
+}
+
+type bulkUpdateDecisionTableRowsBody struct {
+	Rows []bulkUpdateDecisionTableRowBody `json:"rows"`
+}
+
+func decisionTableRowsBulkPath(tableId string, version int, suffix string) string {
+	v := strconv.Itoa(version)
+	return "/api/v2/businessrules/decisiontables/" + tableId + "/versions/" + v + "/rows/bulk/" + suffix
```
brlenn: We should be using the platform SDK and not hardcoding paths, for example here is where we use the sdk for single row create - https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/business_rules_decision_table/genesyscloud_business_rules_decision_table_proxy.go#L310

genesyscloud/business_rules_decision_table/decision_table_bulk_http.go
```diff
@@ -0,0 +1,123 @@
...
+
+	headerParams := make(map[string]string)
+	for key := range p.clientConfig.DefaultHeader {
+		headerParams[key] = p.clientConfig.DefaultHeader[key]
+	}
+	headerParams["Authorization"] = "Bearer " + p.clientConfig.AccessToken
+	headerParams["Content-Type"] = "application/json"
+	headerParams["Accept"] = "application/json"
+
+	response, err := apiClient.CallAPI(fullPath, method, body, headerParams, nil, nil, "", nil, "")
```
sureshperiyappan: response, err := apiClient.CallAPI(fullPath, method, body, headerParams, nil, nil, "", nil, "")

Hey — so CallAPI is the raw HTTP method on the SDK client. The SDK's built-in retry (configured with RetryMax: 20 in provider.go) only kicks in for the generated high-level methods, not for direct CallAPI calls. That means these bulk operations have zero retry on 429 (rate limit) responses.

Given that we're firing multiple bulk requests back-to-back in a loop (especially for tables with 16+ rows), we're pretty likely to hit rate limits in production. Could we wrap the chunk calls in util.RetryWhen similar to how publishDecisionTableVersion handles 409s? Or at minimum add retry logic inside callDecisionTableBulkAPI for 429 status codes.

genesyscloud/business_rules_decision_table/decision_table_bulk_http.go
```diff
@@ -0,0 +1,123 @@
...
+func callDecisionTableBulkAPI(ctx context.Context, p *BusinessRulesDecisionTableProxy, method string, path string, body interface{}) (*platformclientv2.APIResponse, error) {
+	ctx = provider.EnsureResourceContext(ctx, ResourceType)
+	apiClient := &p.clientConfig.APIClient
+	fullPath := p.clientConfig.BasePath + path
+
+	headerParams := make(map[string]string)
+	for key := range p.clientConfig.DefaultHeader {
+		headerParams[key] = p.clientConfig.DefaultHeader[key]
+	}
+	headerParams["Authorization"] = "Bearer " + p.clientConfig.AccessToken
```
sureshperiyappan: headerParams["Authorization"] = "Bearer " + p.clientConfig.AccessToken

This reads the access token once and uses it for all subsequent chunk requests. If the token expires mid-way through processing a large table (say 50+ rows = multiple chunks with network latency), later chunks will get 401s. The SDK's generated methods handle token refresh transparently, but since we're bypassing that here, we're on our own.

Worth verifying: does CallAPI use the header we pass in, or does it pull from Configuration.AccessToken internally? If it uses our header, we should read the token fresh inside each chunk iteration rather than once upfront.

genesyscloud/business_rules_decision_table/decision_table_bulk_http.go
```diff
@@ -0,0 +1,123 @@
...
+	headerParams["Accept"] = "application/json"
+
+	response, err := apiClient.CallAPI(fullPath, method, body, headerParams, nil, nil, "", nil, "")
+	if err != nil {
+		return response, err
+	}
+	if response == nil {
+		return nil, errors.New("nil API response from bulk decision table rows call")
+	}
+	if response.Error != nil {
```
sureshperiyappan: if response.Error != nil {
    return response, errors.New(response.ErrorMessage)
}
if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
Bulk APIs often return 200 with partial failures in the response body (e.g., {"failedRows": [...], "successRows": [...]}). Right now we only check the HTTP status code and never parse the response body. If 3 out of 15 rows fail validation server-side but the API returns 200, we'd silently report success.

Can you check the API docs for the bulk add/remove/update response schema? If partial failures are possible, we need to parse the body and surface which rows failed. If the API is all-or-nothing (entire batch fails on any error), then this is fine — but let's add a comment saying so.

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table.go
```diff
@@ -211,23 +211,31 @@ func addRowsToVersion(ctx context.Context, proxy *BusinessRulesDecisionTableProx
...
 			return fmt.Errorf("failed to convert row %d (table %s, version %d): %s", i+1, tableId, version, err)
 		}
+		// Bulk add does not support rowIndex; rows are appended in request order.
+		sdkRow.RowIndex = nil
+		sdkRows = append(sdkRows, sdkRow)
+	}
 
-		// Add the row to the version
-		_, err = proxy.createDecisionTableRow(ctx, tableId, version, &sdkRow)
+	limAdd, _, _ := getBulkChunkLimits()
```
sureshperiyappan, line 226: limAdd, _, _ := getBulkChunkLimits()
for i := 0; i < len(sdkRows); i += limAdd {
    end := i + limAdd
    if end > len(sdkRows) {
        end = len(sdkRows)
    }
    chunk := sdkRows[i:end]
This is the same chunking logic that appears in applyRowChanges (utils file, ~line 1020). Both do: convert rows → nil out RowIndex → chunk by limit → call bulkAdd. If the bulk add request format changes (it's a preview API after all), we'd need to update both places.

Could we extract something like bulkAddConvertedRows(ctx, proxy, tableId, version, sdkRows) and call it from both addRowsToVersion and applyRowChanges?

Also — the repo already has chunks.ChunkBy[T] in genesyscloud/util/chunks/. The team resource uses it. Kinda funny that the unit test imports chunks.ChunkBy to validate equivalence but the production code reimplements it with manual index math. Not a bug, but feels inconsistent.

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_utils.go
```diff
@@ -979,75 +962,93 @@ func applyRowChanges(ctx context.Context, proxy *BusinessRulesDecisionTableProxy
...
-	// Track successfully added rows for potential rollback
-	var addedRows []string
+	limAdd, limUpd, limRem := getBulkChunkLimits()
 
-	// Delete rows first
-	for _, rowId := range changes.deletes {
-		log.Printf("Deleting row %s", rowId)
-		_, err := proxy.deleteDecisionTableRow(ctx, tableId, version, rowId)
+	// Delete rows first (bulk)
+	for i := 0; i < len(changes.deletes); i += limRem {
```
sureshperiyappan: for i := 0; i < len(changes.deletes); i += limRem {
    ...
    _, err := proxy.bulkRemoveDecisionTableRows(ctx, tableId, version, chunk)
    if err != nil {
        return fmt.Errorf("failed to bulk delete rows: %s", err)
    }
When there are 100 rows to delete (chunks of 49+49+2) and chunk 2 fails, this error message doesn't tell you which chunk or which row IDs were in the failing batch. Compare with addRowsToVersion which does include the range ("failed to bulk add rows [%d:%d]"). Can we add similar context here and in the update section?

Something like:

return fmt.Errorf("failed to bulk delete rows (batch %d-%d of %d): %s", i, end-1, len(changes.deletes), err

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_utils.go
```diff
@@ -979,75 +962,93 @@ func applyRowChanges(ctx context.Context, proxy *BusinessRulesDecisionTableProxy
...
+			end = len(changes.deletes)
+		}
+		chunk := changes.deletes[i:end]
+		log.Printf("Bulk deleting %d rows", len(chunk))
+		_, err := proxy.bulkRemoveDecisionTableRows(ctx, tableId, version, chunk)
 		if err != nil {
-			return fmt.Errorf("failed to delete row %s: %s", rowId, err)
+			return fmt.Errorf("failed to bulk delete rows: %s", err)
+		}
+		if len(chunk) > 0 {
```
sureshperiyappan: resource_genesyscloud_business_rules_decision_table_utils.go, lines ~975, ~1005, ~1035

if len(chunk) > 0 {
    log.Printf("Successfully bulk deleted %d rows", len(chunk))
}
Nit: these if len(chunk) > 0 guards are dead code. The loop only enters when there are items, and each iteration always produces at least 1 element. You can just do the log.Printf directly. Not a big deal, but it reads like the author wasn't sure the loop was correct.

genesyscloud/business_rules_decision_table/decision_table_bulk_chunk_unit_test.go
```diff
@@ -0,0 +1,196 @@
...
+		bulkRemoveLens = append(bulkRemoveLens, len(rowIds))
+		return &platformclientv2.APIResponse{StatusCode: http.StatusNoContent}, nil
+	}
+	proxy.bulkUpdateDecisionTableRowsAttr = func(ctx context.Context, p *BusinessRulesDecisionTableProxy, tableId string, version int, rows []bulkUpdateDecisionTableRowBody) (*platformclientv2.APIResponse, error) {
+		t.Fatal("bulk update should not be called in this test")
+		return nil, nil
+	}
+
+	// Small limits: 5 adds → expect chunk sizes 2,2,1 ; 7 deletes → 3,3,1
+	bulkChunkLimitsOverride = &struct {
```
sureshperiyappan: bulkChunkLimitsOverride = &struct {
    Add, Update, Remove int
}{Add: 2, Update: 15, Remove: 3}
defer func() { bulkChunkLimitsOverride = nil }()
This is a package-level mutable global. If this test panics before the defer runs, or if someone later adds t.Parallel() to it, other tests in the package that call getBulkChunkLimits() could read the overridden values. The first test in this file already uses t.Parallel().

Would be good to at least add a comment like // Must NOT use t.Parallel() — mutates package-level bulkChunkLimitsOverride. Or better, pass limits as a parameter to applyRowChanges so tests don't need shared mutable state.

genesyscloud/business_rules_decision_table/decision_table_bulk_chunk_unit_test.go
```diff
@@ -0,0 +1,196 @@
...
+		deletes: deletes,
+		adds:    adds,
+		updates: nil,
+	})
+	if err != nil {
+		t.Fatalf("applyRowChanges: %v", err)
+	}
+
+	wantAddChunks := [][]int{{2}, {2}, {1}}
+	if len(bulkAddLens) != len(wantAddChunks) {
```
sureshperiyappan: if len(bulkAddLens) != len(wantAddChunks) {
    t.Fatalf("bulk add calls: got %v (len=%d), want %d calls", ...)
}
The rest of the test suite uses testify/assert (see resource_genesyscloud_business_rules_decision_table_unit_test.go — it's assert.Equal, assert.NoError everywhere). This test uses raw t.Fatalf. Minor style inconsistency but it stands out when reading the test files together.

docs/resources/business_rules_decision_table.md
```diff
@@ -36,9 +36,6 @@ The following permissions are required to use this resource:
 * `businessrules:decisionTable:publish`
```
sureshperiyappan, line 39: -* `businessrules:decisionTableRow:add`
-* `businessrules:decisionTableRow:delete`
-* `businessrules:decisionTableRow:edit`
Just confirming — have we verified that the bulk endpoints don't require these permissions? If existing users have OAuth clients scoped to exactly these permissions (following our current docs), upgrading the provider will break them if the bulk APIs need different permissions. Might be worth a note in the changelog about the permission change.

Jx-hou:
Opened a new PR after upgrading to the new sdk version V191 https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/2374

PR #2261, opened 2026-04-07
DEVTOOLING-1605 - Fix routing email domain export/data source lookup for subdomains

daragh-kearney:
- Updated the exporter for genesyscloud_routing_email_domain to generate the data source using the full domain ID (Terraform instance id) instead of the stripped domain_id (fixes subdomains)
- Enhanced data source lookup to be case-insensitive and support subdomain prefix matching
- Fixed API response handling so the list call returns a non-nil response, preventing the “unable to build a message” diagnostic
- Added unit tests covering exporter resolution + case-insensitive and prefix matching behavior


genesyscloud/resource_exporter/resource_exporter.go
```diff
@@ -281,6 +281,13 @@ func (r *ResourceExporter) DataResolver(instanceState *terraform.InstanceState,
 }
 
 func (r *ResourceExporter) fetchFromInstanceState(instanceState *terraform.InstanceState, pattern string) string {
+	// Some data sources accept the Terraform instance ID as their lookup input.
+	// The instance ID is not always present in the attributes map, so allow exporters
+	// to request it explicitly via pattern "id" / "^id$".
+	if instanceState != nil && (pattern == "id" || pattern == "^id$") && instanceState.ID != "" {
```
sureshperiyappan: if instanceState != nil && (pattern == "id" || pattern == "^id$") && instanceState.ID != "" {
    return instanceState.ID
}
This function is called by every exporter in the provider. Hard-coding a string match for "id" introduces a hidden contract — if any resource ever has a legitimate attribute named "id" in its state map, this short-circuit silently wins and returns the instance ID instead of the attribute value.

The "^id$" variant has no callers today and just widens the surface area for accidental collisions.

genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_unit_test.go
```diff
@@ -0,0 +1,77 @@
+package routing_email_domain
```
sureshperiyappan, line 1: Every unit test file in the codebase follows the pattern:
resource_genesyscloud_{name}_unit_test.go

genesyscloud/organization_presence_definition/genesyscloud_organization_presence_definition_proxy.go
sureshperiyappan, line 1: are there any code changes in this files?

daragh-kearney: No code changes here, this was just a line-ending/whitespace-only diff that showed up in the compare view

sureshperiyappan (approved):

PR #2266, opened 2026-04-09
NO-JIRA | Fix issue with downloading export file throws 429

bbbco:
### Description

The `downloadExportFileWithAccessToken` function in `genesyscloud/util/files/util_files.go` makes raw HTTP requests to download export files but had no handling for 429 (Too Many Requests) responses. This could cause downloads to fail unnecessarily during periods of rate limiting.

This PR adds retry logic for 429 responses, following the same pattern already established in `makeFlowRequest` within the architect flow proxy.

### Changes

**`genesyscloud/util/files/util_files.go`**

- Wrapped the HTTP request logic in a retry loop (up to 10 attempts)
- On 429 responses, respects the `Retry-After` header via `util.GetRetryAfterDelay()`
- Falls back to exponential backoff (capped at 30s) when no `Retry-After` header is present
- Logs each retry attempt with the file name, delay, and attempt count
- Returns a descriptive error if all retries are exhausted
- Ensures the response body is properly closed before each retry to prevent resource leaks

### Existing patterns followed

This implementation mirrors the 429 handling in:
- `makeFlowRequest` in `genesyscloud/architect_flow/resource_genesyscloud_architect_flow_proxy.go`
- `createExportJobFn` in the same file

All three use the shared `util.IsStatus429()` and `util.GetRetryAfterDelay()` helpers from `genesyscloud/util/util_retries.go`.

### DIFF
--- a/genesyscloud/util/files/util_files.go
+++ b/genesyscloud/util/files/util_files.go
@@ -252,8 +252,11 @@ func downloadExportFileWithAccessToken(directory, fileName, uri, accessToken str
 		}
 		apiResp, apiErr := platformclientv2.NewAPIResponse(resp, nil)
 
-		if util.IsStatus429(apiResp) && attempt < maxRetries {
+		if util.IsStatus429(apiResp) {
 			resp.Body.Close()
+			if attempt >= maxRetries {
+				return apiResp, fmt.Errorf("exhausted %d retries downloading export file %s due to rate limiting", maxRetries, fileName)
+			}
 			delay, doRetry := util.GetRetryAfterDelay(apiResp)
 			if !doRetry {
 				delay = min(time.Duration(1<<attempt)*time.Second, 30*time.Second)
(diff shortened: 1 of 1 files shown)


genesyscloud/util/files/util_files.go
```diff
@@ -233,44 +233,61 @@ func downloadExportFile(directory, fileName, uri string) (*platformclientv2.APIR
...
+		if err != nil {
+			return nil, err
+		}
+		apiResp, apiErr := platformclientv2.NewAPIResponse(resp, nil)
 
-	if apiErr != nil {
-		return apiResp, apiErr
-	}
-	defer resp.Body.Close()
+		if util.IsStatus429(apiResp) && attempt < maxRetries {
```
sureshperiyappan: When attempt == maxRetries (10), the condition attempt < maxRetries is false, so the 429 check is skipped entirely. Control falls through to if apiErr != nil — but since we're calling NewAPIResponse(resp, nil) with a raw http.Client (not an SDK-generated call), apiErr will be nil. The function then proceeds to os.MkdirAll → os.Create → io.Copy, writing the 429 error response body (likely JSON) to disk as if it were the export file. The "exhausted retries" error at the bottom is dead code for the 429 path.

bbbco: Addressed this issue. Thank you for this catch!

sureshperiyappan (approved):

sanjaysg21 (approved):

PR #2273, opened 2026-04-13
DEVTOOLING-1418 - Export & Apply of genesyscloud_task_management_worktype no longer results in API 400 error with parameter "disable_default_status_creation": true

daragh-kearney:
- Stop export/apply drift: disable_default_status_creation no longer causes a plan/apply update after export
- Create-only flag: set disable_default_status_creation to ForceNew and suppress diffs for it
- Refresh stability: keep disable_default_status_creation in state from config (API doesn’t return it)
- Tests updated: unit tests cover the flag handling


genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype.go
```diff
@@ -123,6 +123,14 @@ func readTaskManagementWorktype(ctx context.Context, d *schema.ResourceData, met
 			resourcedata.SetNillableValue(d, "flow_rules_enabled", worktype.RuleSettings.FlowRulesEnabled)
 		}
 
+		// disable_default_status_creation is a write-only (create-time) flag in the API/SDK.
+		// Keep state aligned with config to avoid perpetual diffs/invalid plans on refresh.
+		if v, ok := d.GetOkExists("disable_default_status_creation"); ok {
```
sureshperiyappan: d.GetOkExists() is deprecated in the Terraform Plugin SDK (ref: hashicorp/terraform-plugin-sdk#817). This codebase already has a resourcedata.GetNillableBool() helper in 
resourcedata.go
 that wraps this same call with a deprecation acknowledgment. Please use that helper instead of calling the deprecated method directly.

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_schema.go
```diff
@@ -134,7 +134,12 @@ func ResourceTaskManagementWorktype() *schema.Resource {
 				Description: `Optionally set this flag to disable Default Status creation`,
 				Optional:    true,
 				Type:        schema.TypeBool,
-				Computed:    true,
+				ForceNew:    true,
```
sureshperiyappan: ForceNew: true combined with a DiffSuppressFunc that always returns true is contradictory. ForceNew means "recreate the resource if this value changes," but the diff suppression ensures no diff is ever detected, so ForceNew can never trigger. It's dead code. Since the goal of this JIRA is to prevent the field from ever appearing in a PATCH (which the DiffSuppressFunc correctly achieves), I'd recommend removing ForceNew to avoid confusing future readers. If ForceNew semantics are desired later, the DiffSuppressFunc would need to be made conditional rather than a blanket return true. 

Check this and take a decision

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_utils.go
```diff
@@ -39,12 +39,7 @@ type worktypeConfig struct {
 
 // getWorktypeCreateFromResourceData maps data from schema ResourceData object to a platformclientv2.Worktypecreate
 func getWorktypecreateFromResourceData(d *schema.ResourceData) platformclientv2.Worktypecreate {
-	DisableDefaultStatusCreation := platformclientv2.Bool(func() bool {
-		if v, ok := d.GetOk("disable_default_status_creation"); ok {
-			return v.(bool)
-		}
-		return false
-	}())
+	DisableDefaultStatusCreation := platformclientv2.Bool(d.Get("disable_default_status_creation").(bool))
```
sureshperiyappan: The bare type assertion d.Get("disable_default_status_creation").(bool) works today because TypeBool zero-value is false, but it's fragile and inconsistent with how other optional fields in this same function use GetNillableValue / GetNonZeroPointer. Consider using resourcedata.GetNillableBool() for consistency and defensiveness.

sureshperiyappan: DisableDefaultStatusCreation is PascalCase which in Go convention means an exported identifier. Local variables should be camelCase (disableDefaultStatusCreation). Pre-existing issue but since this line is being modified, worth cleaning up.

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_unit_test.go
```diff
@@ -34,11 +34,12 @@ func TestUnitResourceWorktypeCreate(t *testing.T) {
...
-		defaultQueueId:    uuid.NewString(),
-		defaultSkillIds:   []string{uuid.NewString(), uuid.NewString()},
-		defaultScriptId:   uuid.NewString(),
-		assignmentEnabled: false,
+		defaultLanguageId:            uuid.NewString(),
+		defaultQueueId:               uuid.NewString(),
+		defaultSkillIds:              []string{uuid.NewString(), uuid.NewString()},
+		defaultScriptId:              uuid.NewString(),
+		assignmentEnabled:            false,
+		disableDefaultStatusCreation: true,
```
sureshperiyappan: Test coverage gaps: (1) The create test only covers disableDefaultStatusCreation: true. Add a case for false to verify the zero-value path. (2) TestUnitResourceWorktypeRead doesn't set or assert disable_default_status_creation at all — since the read function now has explicit logic for this field, the read test should verify it's preserved correctly. (3) The actual JIRA scenario is export → apply → 400 on PATCH. Consider adding a test that verifies the update path does not include disable_default_status_creation in the update payload when the field is present in state.

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_unit_test.go
```diff
@@ -157,6 +160,7 @@ func TestUnitResourceWorktypeCreate(t *testing.T) {
 	//Found this gem TestResourceDataRaw here: https://github.com/hashicorp/terraform/issues/890
 	d := schema.TestResourceDataRaw(t, resourceSchema, resourceDataMap)
 	d.SetId(tId)
+	_ = d.Set("disable_default_status_creation", wt.disableDefaultStatusCreation)
```
sureshperiyappan: This explicit d.Set() call after TestResourceDataRaw looks like a workaround. Please add a brief comment explaining why it's needed (e.g., TestResourceDataRaw doesn't properly handle ForceNew/DiffSuppressFunc fields), otherwise the next person will wonder why it's not handled through buildWorktypeResourceMap like every other field.

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype.go
```diff
@@ -123,6 +123,14 @@ func readTaskManagementWorktype(ctx context.Context, d *schema.ResourceData, met
...
+			_ = d.Set("disable_default_status_creation", v.(bool))
+		} else {
+			_ = d.Set("disable_default_status_creation", nil)
```
sureshperiyappan: For a TypeBool schema field, setting nil via d.Set() effectively stores the zero value (false). If the attribute was never in state, it's already false. This else branch is a no-op and is slightly misleading since it suggests the field can be "unset" in state. Consider removing the else branch or adding a comment clarifying the intent.

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype.go
daragh-kearney: replaced direct d.GetOkExists() usage with resourcedata.GetNillableBool() in the read logic

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_schema.go
daragh-kearney: I have now removed ForceNew and kept diff suppression

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_utils.go
daragh-kearney: create payload now uses resourcedata.GetNillableBool() so unset stays nil and explicitly-set false stays &false

daragh-kearney: Done, I've renamed to disableDefaultStatusCreation

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_unit_test.go
daragh-kearney: Now create test covers unset/true/false, read test sets+asserts the field, update test asserts marshaled update JSON doesn’t contain the field

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype_unit_test.go
daragh-kearney: Addressed by clarifying in the test (kept the explicit d.Set when simulating “configured false”, since TestResourceDataRaw can omit optional bool=false)

genesyscloud/task_management_worktype/resource_genesyscloud_task_management_worktype.go
daragh-kearney: I've now removed the else branch

sureshperiyappan (approved):

Issue #2274, opened 2026-04-14
Issue during the import of a routing_queue : mismatch on attribute media_settings_message.sub_type_settings.media_type


fabiengautier-axa:
Hello,

I face an issue during the `genesyscloud_routing_queue` creation.
I have the following error :
```
genesyscloud_routing_queue.XXX: Still modifying... [id=YYY-YYY-YYY, 05m00s elapsed]
╷
│ Error: mismatch on attribute media_settings_message.0.sub_type_settings.2.media_type:
│ expected value: <nil>
**│ actual value: webmessaging**
│ 
│   with genesyscloud_routing_queue.XXX,
│   on genesyscloud_routing_queue.tf line XXX, in resource "genesyscloud_routing_queue" "XXX":
│ 20585: resource "genesyscloud_routing_queue" "XXX" {
```

The terraform code is : 
```
resource "genesyscloud_routing_queue" "XXX" {
[...]
  default_script_ids = {
    MESSAGE = "${genesyscloud_script.XXX}"
  }
  media_settings_message {
    service_level_duration_ms = 20000
    service_level_percentage  = 0.8
    sub_type_settings {
      enable_auto_answer = true
      **media_type         = "webmessaging"**
    }
    sub_type_settings {
      media_type         = "whatsapp"
      enable_auto_answer = true
    }
    [...]
    alerting_timeout_sec      = 30
    enable_auto_answer        = true
    enable_inactivity_timeout = false
  }
  scoring_method          = "TimestampAndPriority"
  enable_transcription    = false
  division_id             = "${genesyscloud_auth_division.XXX.id}"
  auto_answer_only        = false
  last_agent_routing_mode = "QueueMembersOnly"
  media_settings_call {
    alerting_timeout_sec      = 8
    enable_auto_answer        = false
    service_level_duration_ms = 20000
    service_level_percentage  = 0.8
  }
  [...]
}
```

I'm using the last provider version : 1.77.4

What value should I use for web messaging?

Thanks in advance,

bbbco:
Hi,

It seems that the issue might be that the API is returning three different `sub_type_settings` entries (notice the `2` in `media_settings_message.0.sub_type_settings.2.media_type` indicates a zero-based index of at least three items) for this instance when your Terraform configuration only has two `sub_type_settings` entries configured. This is what the consistency checker is getting caught on.

[...]

PR #2292, opened 2026-04-23
Aligns Routing Queue resource's attributes with SDK's attributes and descriptions

bbbco:
## Impetus

While looking into https://github.com/MyPureCloud/terraform-provider-genesyscloud/issues/2274 I realized that there were some inconsistencies, particularly with the extra `sub_type_settings` in other media settings (and not just the messaging settings). While doing manual comparisons between the API Explorer clued me into this issue, I decided to ask Amazon Q about any inconsistencies it found between the existing attributes and to fix them.

## Summary

Aligns the `genesyscloud_routing_queue` Terraform schema with the Genesys Cloud SDK v179.1.0 `Createqueuerequest` model. Fixes incorrect field placement, adds missing fields, and corrects descriptions to match the SDK.

### Schema corrections (`resource_genesyscloud_routing_queue_schema.go`)

**Removed `sub_type_settings` from incorrect media settings:**
- Removed from `queueMediaSettingsResource` (used by call, chat, email). The SDK's `Mediasettings` and `Emailmediasettings` types do not have `SubTypeSettings`.
- Removed inherited reference from `queueCallbackMediaSettingsResource`. The SDK's `Callbackmediasettings` does not have `SubTypeSettings`.
- `sub_type_settings` remains correctly on `queueMediaSettingsMessageResource` only, matching the SDK's `Messagemediasettings`.

**Added missing fields to `queueMediaSettingsResource`:**
- `auto_answer_alert_tone_seconds` — present on SDK's `Mediasettings` and `Emailmediasettings` but was missing from the Terraform schema for call, chat, and email.
- `manual_answer_alert_tone_seconds` — same as above.

**Added missing field to `subTypeSettingsResource`:**
- `enable_inactivity_timeout` — present on SDK's `Messagesubtypesettings` but was missing from the Terraform schema.

**Fixed 11 descriptions to match SDK model comments:**
- `agent_owned_routing` — was "Agent Owned Routing."
- `enable_agent_owned_callbacks` — was "Enable Agent Owned Callbacks"
- `enable_auto_dial_and_end` — was "Auto Dial and End"
- `auto_dial_delay_seconds` — was "Auto Dial Delay Seconds."
- `auto_end_delay_seconds` — was "Auto End Delay Seconds."
- `direct_routing` — was "Used by the System to set Direct Routing settings for a system Direct Routing queue."
- `backup_queue_id` — was vague, now matches SDK
- `agent_wait_seconds` — was missing valid range
- `wait_for_agent` — was slightly inaccurate
- `acw_timeout_ms` — was missing min/max info from SDK
- `media_type` (subTypeSettingsResource) — was "The name of the social media company"
- `canned_response_libraries` — was "Agent Owned Routing." (copy-paste error)
- `max_owned_callback_hours` — was "Auto End Delay Seconds Must be >= 7"
- `max_owned_callback_delay_hours` — was "Max Owned Call Back Delay Hours >= 7"

### Build/flatten function updates (`resource_genesyscloud_routing_queue_utils.go`)

- `buildSdkMediaSetting` — now sets `AutoAnswerAlertToneSeconds` and `ManualAnswerAlertToneSeconds`
- `buildSdkMediaEmailSetting` — now sets `AutoAnswerAlertToneSeconds` and `ManualAnswerAlertToneSeconds`
- `buildSubTypeSettings` — now sets `EnableInactivityTimeout` on `Messagesubtypesettings`
- `flattenMediaSetting` — now reads `AutoAnswerAlertToneSeconds` and `ManualAnswerAlertToneSeconds`
- `flattenMediaEmailSetting` — now reads `AutoAnswerAlertToneSeconds` and `ManualAnswerAlertToneSeconds`
- `flattenSubTypeSettings` — now reads `EnableInactivityTimeout`

### Dev tooling

- Added `.sdk-reference` symlink to Go module cache for Amazon Q SDK reference access
- Added `.sdk-reference` to `.gitignore`
- Added `.amazonq/rules/sdk-reference.md` rule to ensure SDK models are always consulted for field definitions

## Migration

No state migration is required. All changes are transparent:
- New Optional fields without Defaults don't break existing state
- Removed `sub_type_settings` on call/chat/email/callback was never read from the API or written to state by the flatten functions
- Description changes don't affect state

## Testing

All existing unit tests pass. No new tests required — the existing `TestUnitResourceRoutingQueueUpdate` and `TestUnitBuildSdkMediaSettingCallback` tests validate the build/flatten changes.

## AI Disclosure

This PR was developed with Amazon Q Developer assistance. The following prompts were used and can be reused to arrive at the same conclusions:

1. **Schema comparison**: "Compare the schema defined on this swagger link: https://developer.genesys.cloud/devapps/api-explorer#post-api-v2-routing-queues with the schema in this file. What is different?" — followed by redirecting to the SDK source at `$GOHOME/go/pkg/mod/github.com/mypurecloud/platform-client-sdk-go/v179@v179.1.0/platformclientv2/` for accurate comparison against `Createqueuerequest`, `Mediasettings`, `Emailmediasettings`, `Callbackmediasettings`, `Messagemediasettings`, `Messagesubtypesettings`, and `Cannedresponselibraries` Go structs.
2. **sub_type_settings analysis**: "Let's talk about sub_type_settings and dig into exactly what needs to be changed for this across all the various media_settings_* attributes" — comparing which SDK media settings types have `SubTypeSettings` (only `Messagemediasettings`) vs which Terraform schemas incorrectly included it.
3. **Migration assessment**: "What would the migration look like? Is a migration required?" — concluded no migration needed because removed fields were never written to state by flatten functions, and new fields are Optional without Defaults.
4. **Description audit**: "Can you check all of the descriptions in all of the fields in this schema?" — systematic comparison of every Terraform field description against the corresponding SDK struct comment, reading directly from the Go module cache.

### DIFF
--- a/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
+++ b/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
@@ -140,7 +140,7 @@ func buildSdkMediaEmailSetting(settings []interface{}) *platformclientv2.Emailme
 		return nil
 	}
 
-	return &platformclientv2.Emailmediasettings{
+	emailSetting := &platformclientv2.Emailmediasettings{
 		AlertingTimeoutSeconds: platformclientv2.Int(settingsMap["alerting_timeout_sec"].(int)),
 		EnableAutoAnswer:       platformclientv2.Bool(settingsMap["enable_auto_answer"].(bool)),
 		ServiceLevel: &platformclientv2.Servicelevel{
@@ -148,6 +148,11 @@ func buildSdkMediaEmailSetting(settings []interface{}) *platformclientv2.Emailme
 			DurationMs: platformclientv2.Int(settingsMap["service_level_duration_ms"].(int)),
 		},
 	}
+
+	emailSetting.AutoAnswerAlertToneSeconds = resourcedata.GetNillableValueFromMap[float64](settingsMap, "auto_answer_alert_tone_seconds", false)
+	emailSetting.ManualAnswerAlertToneSeconds = resourcedata.GetNillableValueFromMap[float64](settingsMap, "manual_answer_alert_tone_seconds", false)
+
+	return emailSetting
 }
 
 func buildSdkMediaSetting(settings []interface{}) *platformclientv2.Mediasettings {
@@ -159,7 +164,7 @@ func buildSdkMediaSetting(settings []interface{}) *platformclientv2.Mediasetting
 		return nil
 	}
 
-	return &platformclientv2.Mediasettings{
+	mediaSetting := &platformclientv2.Mediasettings{
 		AlertingTimeoutSeconds: platformclientv2.Int(settingsMap["alerting_timeout_sec"].(int)),
 		EnableAutoAnswer:       platformclientv2.Bool(settingsMap["enable_auto_answer"].(bool)),
 		ServiceLevel: &platformclientv2.Servicelevel{
@@ -167,6 +172,11 @@ func buildSdkMediaSetting(settings []interface{}) *platformclientv2.Mediasetting
 			DurationMs: platformclientv2.Int(settingsMap["service_level_duration_ms"].(int)),
 		},
 	}
+
+	mediaSetting.AutoAnswerAlertToneSeconds = resourcedata.GetNillableValueFromMap[float64](settingsMap, "auto_answer_alert_tone_seconds", false)
+	mediaSetting.ManualAnswerAlertToneSeconds = resourcedata.GetNillableValueFromMap[float64](settingsMap, "manual_answer_alert_tone_seconds", false)
+
+	return mediaSetting
 }
 
 func buildSdkMediaSettingsMessage(settings []any) *platformclientv2.Messagemediasettings {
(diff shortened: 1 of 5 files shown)


genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -298,10 +308,13 @@ func buildSubTypeSettings(subTypeList []interface{}) *map[string]platformclientv
 		subTypeMap := subTypeItem.(map[string]interface{})
 		mediaType := subTypeMap["media_type"].(string)
 		enableAutoAnswer := subTypeMap["enable_auto_answer"].(bool)
-		baseMediaSettings := platformclientv2.Messagesubtypesettings{
+		subTypeSetting := platformclientv2.Messagesubtypesettings{
 			EnableAutoAnswer: &enableAutoAnswer,
 		}
-		returnObj[mediaType] = baseMediaSettings
+		if enableInactivityTimeout, ok := subTypeMap["enable_inactivity_timeout"].(bool); ok && enableInactivityTimeout {
```
sureshperiyappan: enable_inactivity_timeout = false will never propagate to the API

if enableInactivityTimeout, ok := subTypeMap["enable_inactivity_timeout"].(bool); ok && enableInactivityTimeout {
    subTypeSetting.EnableInactivityTimeout = &enableInactivityTimeout
}
The && enableInactivityTimeout guard means this only sets the field when the value is true. If a user explicitly configures enable_inactivity_timeout = false (e.g. to disable it after previously enabling it), the condition short-circuits, EnableInactivityTimeout stays nil, the SDK omits it from the request, and the API value is never updated. This results in a perpetual plan diff that apply can never resolve.

Suggestion check it?:

if v, ok := subTypeMap["enable_inactivity_timeout"].(bool); ok {
    subTypeSetting.EnableInactivityTimeout = &v
}
This is consistent with how other boolean fields are handled elsewhere in this file (e.g. EnableAutoAnswer is set unconditionally via platformclientv2.Bool(...)).

bbbco: Great find. Thanks for this! Fixed as suggested.

sureshperiyappan (approved):

PR #2300, opened 2026-04-28
DEVTOOLING-1671 - genesyscloud_speechandtextanalytics_topic now exports as resources instead of data sources

daragh-kearney:
- Removed the ExportAsDataFunc override on genesyscloud_speechandtextanalytics_topic, so topics now export as normal resource blocks by default (customers can still opt-in to data sources via replace_with_datasource).
- Fix tf_export crash (fatal error: concurrent map writes): GetCustomRetryTimeout() now serializes access to the shared providerConfig (schema.ResourceData) to avoid concurrent access during parallel export/read operations.
- Added a unit test that calls GetCustomRetryTimeout() concurrently to ensure the provider doesn’t reintroduce the concurrency panic.

### DIFF
--- a/genesyscloud/provider/provider_utils.go
+++ b/genesyscloud/provider/provider_utils.go
@@ -129,10 +129,13 @@ func GetOrgDefaultCountryCode() string {
 	return meta.DefaultCountryCode
 }
 
-func parseCustomRetryTimeout(config *schema.ResourceData) time.Duration {
-	// Provider config takes precedence over env var/default.
-	// This function is only intended to be called during provider configuration,
-	// where ResourceData access is not concurrent.
+// GetCustomRetryTimeout returns the configured custom retry timeout.
+// It first checks the provider configuration, then falls back to the environment variable,
+// and finally uses the default value (5 minutes).
+// A timeout of 0 means no retries (immediate fail-fast behavior).
+func GetCustomRetryTimeout() time.Duration {
+	// First try to get from provider configuration
+	config := GetProviderConfig()
 	if config != nil {
 		if v, ok := config.GetOk(AttrCustomRetryTimeout); ok {
 			if timeout, err := time.ParseDuration(v.(string)); err == nil {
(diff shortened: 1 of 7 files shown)


genesyscloud/provider/provider_utils.go
```diff
@@ -134,15 +134,21 @@ func GetOrgDefaultCountryCode() string {
...
-	// First try to get from provider configuration
-	config := GetProviderConfig()
+	// First try to get from provider configuration.
+	// IMPORTANT: schema.ResourceData is not safe for concurrent access. Since exporter and
+	// read operations can run in parallel goroutines, we must serialize access to the
+	// stored provider config to avoid "fatal error: concurrent map writes" inside the SDK.
+	mutex.Lock()
+	config := providerConfig
 	if config != nil {
 		if v, ok := config.GetOk(AttrCustomRetryTimeout); ok {
```
bbbco, line 140: We should be caching the AttrCustomRetryTimeout value on the Meta object once so it can be easily referenced whenever it is needed instead of introducing another lock mechanism, otherwise the lock will potentially slow down all the Read operations since this is called by WithRetriesForRead. 
Also consider, the AttrCustomRetryTimeout value is immutable after provider configuration (it's set once in configure() and never changes). Locking on every read for a constant value is wasteful.

We already have the pattern for reading once and setting on Meta established for `GetOrgDefaultCountryCode()` so we should just do the same approach.

genesyscloud/provider/provider_utils.go
```diff
@@ -134,15 +134,21 @@ func GetOrgDefaultCountryCode() string {
 // and finally uses the default value (5 minutes).
 // A timeout of 0 means no retries (immediate fail-fast behavior).
 func GetCustomRetryTimeout() time.Duration {
-	// First try to get from provider configuration
-	config := GetProviderConfig()
+	// First try to get from provider configuration.
+	// IMPORTANT: schema.ResourceData is not safe for concurrent access. Since exporter and
+	// read operations can run in parallel goroutines, we must serialize access to the
+	// stored provider config to avoid "fatal error: concurrent map writes" inside the SDK.
```
bbbco: Should this be broken out into a separate PR so it doesn't pollute the intent of the original PR?

bbbco (changes requested):

bbbco (approved):

sureshperiyappan (approved):

shrutisuryawanshigenesys (approved):

Issue #2303, opened 2026-04-28
Campaign Rule objects don't support "Time Of The Day" conditions.


mikebailey2git:
We are managing all of our campaign object with terraform. Today we added a new Campaign Rule based on campaign progress and the time of day, Is it possible to add the time of day type conditions to these objects? Current supported list of conditions types seem to be:
"campaignProgress" "campaignAgents" "campaignRecordsAttempted" "campaignContactsMessaged" "campaignBusinessSuccess" "campaignBusinessNeutral" "campaignBusinessFailure" "campaignValidAttempts" "campaignRightPartyContacts"

mikebailey2git:
Just in case it isn't obvious, we would want the parameters that are needed to time of day condition types to be added as well.

bbbco:
Thanks for the report! I submitted a fix for it.

PR #2307, opened 2026-04-29
Fixes DEVTOOLING-1678 and GH-2303

bbbco:
Fixes DEVTOOLING-1678 and GH-2303: Align the getAllowedConditions with what the API specs indicate are valid

### DIFF
--- a/docs/resources/outbound_campaignrule.md
+++ b/docs/resources/outbound_campaignrule.md
@@ -139,7 +139,7 @@ Optional:
 
 Required:
 
-- `condition_type` (String) The type of condition to evaluate (campaignProgress | campaignAgents | campaignRecordsAttempted | campaignContactsMessaged | campaignBusinessSuccess | campaignBusinessNeutral | campaignBusinessFailure | campaignValidAttempts | campaignRightPartyContacts)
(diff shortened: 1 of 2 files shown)


sureshperiyappan:
@bbbco 
Docs not updated: 
outbound_campaignrule.md
 still lists only the original 9 condition types in the condition_type description. The 7 new values need to be added there too. If this file is auto-generated via tfplugindocs, the description string in the schema source needs updating so the docs regenerate correctly.

sureshperiyappan:
@bbbco 
No test coverage for new types: The existing tests only exercise campaignProgress and campaignAgents. The time/date-based conditions (timeOfDay, dayOfWeek, etc.) are structurally different — they likely don't require a campaign entity reference and may need different parameter combinations. At minimum, one acceptance test using a time-based condition type (e.g., timeOfDay) would help confirm the provider handles them end-to-end without unexpected validation or API errors.

sureshperiyappan (approved):
please address those changes mentioned in the comment. otherwise, it looks good.

bbbco:
> @bbbco No test coverage for new types: The existing tests only exercise campaignProgress and campaignAgents. The time/date-based conditions (timeOfDay, dayOfWeek, etc.) are structurally different — they likely don't require a campaign entity reference and may need different parameter combinations. At minimum, one acceptance test using a time-based condition type (e.g., timeOfDay) would help confirm the provider handles them end-to-end without unexpected validation or API errors.

As I was looking into addressing this item, I found that it is expanding the scope of this PR as you were correct that time-based conditions rely on the `dateTimeParameters` field that needs to be added as well. Unfortunately, these  time based condition fields are only showing up in a later version of the SDK.

I also found that the time/date based conditions are primarily enabled in the UI via the "advanced" usage which enables the Condition Groups from the v2 rules (the provider is only currently enabled to use the default v1 rules). A work item is open for this: DEVTOOLING-1679

I think we need to put this PR on hold until the following items have been addressed:
1. SDK is updated to v184+
2. Support for v2 rules is added

Issue #1904, opened 2025-09-15
Add a note to the genesyscloud_integration_action docs page about "static" data actions and that we don't export them


scheblein:
I just wasted a few hours troubleshooting why all my Data Actions weren't exporting. Finally looked in the provider source code and saw that we don't export static actions. That is fine, but I think we should give our users of the provider a heads up that we don't export static data actions and the reason why. I believe this is the best location to put this note: https://registry.terraform.io/providers/MyPureCloud/genesyscloud/latest/docs/resources/integration_action

PR #2294, opened 2026-04-23
Fix #2290 Add note to Integration Actions docs about exporter behavior

bbbco:
Fixes https://github.com/MyPureCloud/terraform-provider-genesyscloud/issues/2290

PR #2309, opened 2026-04-30
DEVTOOLING-1438 - genesyscloud_integration_action - Updated provider docs with note about static data actions

daragh-kearney:
This is a fix for https://github.com/MyPureCloud/terraform-provider-genesyscloud/issues/1904

### DIFF
--- a/docs/resources/integration_action.md
+++ b/docs/resources/integration_action.md
@@ -42,6 +42,20 @@ The following OAuth scopes are required to use this resource:
 * [PUT /api/v2/integrations/actions/{actionId}/draft/function](https://developer.genesys.cloud/api/rest/v2/integrations/#put-api-v2-integrations-actions--actionId--draft-function)
 * [POST /api/v2/integrations/actions/{actionId}/draft/publish](https://developer.genesys.cloud/api/rest/v2/integrations/#post-api-v2-integrations-actions--actionId--draft-publish)
 
+## Export Behavior
+
+### Static Data Actions Are Not Exported
+
+When exporting integration actions via the `genesyscloud_tf_export` resource, **static (built-in) data actions are intentionally excluded from the exported configuration**. Static data actions are the pre-installed system actions that ship with each Genesys Cloud integration; their IDs are prefixed with `static` (for example, `static_e7b86b86-...`).
+
+These actions are owned and managed by Genesys Cloud and cannot be created, updated, or deleted through the public Integration Actions API. Exporting them would therefore produce Terraform configuration that fails on apply, so the exporter skips them on purpose.
+
+#### What this means for you
+
+- Only custom integration actions that you (or your team) have created will appear in the exported `.tf`/`.tf.json` output.
+- If a static data action is missing from your export, this is expected behavior — not a bug or a permissions issue.
+- If you need to reference a static data action from another resource (for example, an Architect flow), reference it by its existing static ID directly rather than expecting it to be present in the exported configuration.
+
 ## Example Usage
 
 ```terraform
--- a/docs/resources/integration_action.md
+++ b/docs/resources/integration_action.md
@@ -44,17 +44,18 @@ The following OAuth scopes are required to use this resource:
 
 ## Export Behavior
 
-### Static Data Actions Are Not Exported
+### Static Data Actions Are Exported as Data Sources
 
-When exporting integration actions via the `genesyscloud_tf_export` resource, **static (built-in) data actions are intentionally excluded from the exported configuration**. Static data actions are the pre-installed system actions that ship with each Genesys Cloud integration; their IDs are prefixed with `static` (for example, `static_e7b86b86-...`).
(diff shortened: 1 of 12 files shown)


docs/resources/integration_action.md
```diff
@@ -42,6 +42,20 @@ The following OAuth scopes are required to use this resource:
...
+
+When exporting integration actions via the `genesyscloud_tf_export` resource, **static (built-in) data actions are intentionally excluded from the exported configuration**. Static data actions are the pre-installed system actions that ship with each Genesys Cloud integration; their IDs are prefixed with `static` (for example, `static_e7b86b86-...`).
+
+These actions are owned and managed by Genesys Cloud and cannot be created, updated, or deleted through the public Integration Actions API. Exporting them would therefore produce Terraform configuration that fails on apply, so the exporter skips them on purpose.
+
+#### What this means for you
+
+- Only custom integration actions that you (or your team) have created will appear in the exported `.tf`/`.tf.json` output.
+- If a static data action is missing from your export, this is expected behavior — not a bug or a permissions issue.
+- If you need to reference a static data action from another resource (for example, an Architect flow), reference it by its existing static ID directly rather than expecting it to be present in the exported configuration.
```
bbbco: Should we export static items as data resources so they can be referenced?

daragh-kearney: If they haven't been exported before then there may be a reason why they aren't, but I'm not certain so will discuss with the team to fully understand the implementation

bbbco: Yeah, also, I had already put in a fix for this last week: https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/2294

bbbco: I'm good with either doc update. Just glad to get it in to lessen any confusion. However, I do think exporting static as data resources is the right way to go.

daragh-kearney: Oh nice, I didn't realize you put in a fix so either is good! Think Suresh is going to take a look at this when he gets the chance and we should get some clarity on the implementation.

If we're going ahead with exporting static as data resources I can update the PR to accommodate that!

sureshperiyappan: @bbbco  & @daragh-kearney  Agreed — to make dependencies work seamlessly, we should export static actions as data sources so they can be properly referenced by other resources.

sureshperiyappan (commented):
Provided my comment

genesyscloud/integration_action/resource_genesyscloud_integration_action.go
```diff
@@ -51,10 +51,10 @@ func getAllIntegrationActions(ctx context.Context, clientConfig *platformclientv
...
 
 	for _, action := range *actions {
-		// Don't include "static" actions
-		if strings.HasPrefix(*action.Id, "static") {
-			continue
-		}
+		// Static (built-in) data actions are owned by Genesys Cloud and cannot be created,
+		// updated, or deleted via the public API. They are included here so the exporter
+		// can emit them as data sources (see ExportAsDataFunc on the exporter) rather than
+		// as managed resources, allowing other resources to reference them.
```
bbbco: Should we also distinguish the block label with the name of the parent integration?

bbbco (approved):

genesyscloud/integration_action/resource_genesyscloud_integration_action.go
daragh-kearney: Good shout, working on an update for this now

genesyscloud/integration_action/resource_genesyscloud_integration_action.go
```diff
@@ -50,20 +50,91 @@ func getAllIntegrationActions(ctx context.Context, clientConfig *platformclientv
...
+	}
+
+	if action.IntegrationId == nil {
+		return category + "_" + name
+	}
+	integrationName, ok := integrationNamesById[*action.IntegrationId]
+	if !ok || integrationName == "" {
+		return category + "_" + name
+	}
+	return category + "_" + integrationName + "_" + name
```
bbbco: I'd recommend putting integrationName first, with a distinct delimiter to make distinguishing the integrationName from category and name. Maybe three underscores?

```
return integrationName + "___" + category + "_" + name
```

In fact, to DRY things up a bit you could set:
```
blockLabel := category + "_" + name
```
and 
```
return blockLabel
```
in lines 125, 129, and 133 and then in line 135:
```
blockLabel = integrationName + "___" + blockLabel
return blockLabel
```

daragh-kearney: Apologies for delay on this I've been out of office, updated so buildIntegrationActionBlockLabel now sets blockLabel := category + "_" + name once and the early-return branches all just return blockLabel. For static actions it becomes blockLabel = integrationName + "___" + blockLabel

sureshperiyappan (approved):

PR #2318, opened 2026-05-08
fix for DEVTOOLING-1680

sureshperiyappan:


bbbco:
You should probably add a note in the docs indicating that these resources were previously introduced under a different resource type name in the prior version, but fixed in this version going forward to ensure compatibility with the existing API naming conventions.

bbbco (approved):

shrutisuryawanshigenesys (approved):

sanjaysg21 (approved):

PR #2320, opened 2026-05-12
Bug/devtooling 1657  provider.sdk_debug doesn't log PUT api response body for HTTP 500 errors

shrutisuryawanshigenesys:
This PR fixes the missing request and response logs when API calls fail with 429 errors after all retries are exhausted. It matches errors to the provider log file using the correlationID from the request hook.


genesyscloud/provider/provider_sdk_debug.go
```diff
@@ -8,18 +8,212 @@ import (
...
+
+var sdkDebugErrorMirrorMu sync.Mutex
+
+// TF-Correlation-Id -> last outbound body (so error mirror can log request after send).
+var sdkDebugMirrorRequestBodies sync.Map
+
+// Max bytes written per mirrored request/response body in sdk_debug.log.
+const maxMirrorBodyBytes = 1 << 20
+
+func storeSDKDebugMirrorRequestBodyForHook(dr *sdkDebugRequest) {
```
sureshperiyappan, line 35: This file introduces several new pure functions that are highly testable in isolation but have no corresponding unit tests. Consider adding a provider_sdk_debug_test.go in a follow-up with coverage for:

sdkDebugStatusNeedsErrorMirror — table-driven test for status code boundaries (429, 5xx, 501 exclusion, etc.)
cloneHTTPHeaderRedactAuth — verify Authorization is redacted, other headers preserved
storeSDKDebugMirrorRequestBodyForHook / popSDKDebugMirrorRequestBody — store/pop lifecycle, empty correlation ID, disabled flag
readRequestBodyForHook — prealloc path vs fallback path
These are all side-effect-free and don't need mocking, so they'd be quick to write and would protect against regressions.

sureshperiyappan (approved):

PR #2321, opened 2026-05-12
DEVTOOLING-1681 - genesyscloud_outbound_campaign agentOwnedColumn + other missing fields

daragh-kearney:
- Added missing agent_owned_column field to genesyscloud_outbound_campaign 
- Adds missing max_calls_per_agent_decimal field 
- Adds missing diagnostics_settings block with report_low_max_calls_per_agent_alert
- Regenerated docs and updated example resource.tf to demonstrate diagnostics_settings
- Extended TestAccResourceOutboundCampaignBasic to assert the new diagnostics_settings value


genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_test.go
```diff
@@ -465,6 +465,7 @@ func TestAccResourceOutboundCampaignBasic(t *testing.T) {
 						contactSortNumeric,
```
sureshperiyappan, line 465: In TestAccResourceOutboundCampaignBasic, the final update step adds generateDiagnosticsSettingsBlock(util.TrueValue) to a campaign using dialingModeUpdate = "agentless". The CustomizeDiff function validateDiagnosticsSettingsDialingMode explicitly rejects this combination:

diagnostics_settings is only applicable to Power and Predictive dialing modes, but dialing_mode is set to 'agentless'

This test step will fail at plan time. Either remove the diagnostics_settings block from this step (it's already well-covered in TestAccResourceOutboundCampaignPower and TestAccResourceOutboundCampaignPredictiveDiagnosticsSettings), or change the dialing mode for that step to "power".

examples/resources/genesyscloud_outbound_campaign/resource.tf
```diff
@@ -55,4 +55,7 @@ resource "genesyscloud_outbound_campaign" "campaign2" {
   phone_columns {
     column_name = "Cell"
   }
+  diagnostics_settings {
```
sureshperiyappan: campaign2 has dialing_mode = "agentless" but includes:

diagnostics_settings {
  report_low_max_calls_per_agent_alert = true
}
This is invalid per the validation logic and would fail if a user tried to apply it. The example should either remove the diagnostics_settings block from campaign2, or the power_campaign example already demonstrates it correctly.

genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_utils.go
```diff
@@ -195,6 +196,21 @@ func buildLineBalancingSettings(settings []interface{}) *platformclientv2.Dynami
 	return &sdkLineBalancingSettings
 }
 
+func buildDiagnosticsSettings(settings []interface{}) *platformclientv2.Diagnosticssettings {
```
sureshperiyappan, line 190: The new buildDiagnosticsSettings checks len(settings) < 1 but doesn't check settings == nil first. While this works in Go (nil slice has length 0), the other builder functions in the same file (buildSettings, buildLineBalancingSettings) explicitly check settings == nil || len(settings) < 1. For consistency and defensive coding, consider adding the nil check:


func buildDiagnosticsSettings(settings []interface{}) *platformclientv2.Diagnosticssettings {
	if settings == nil || len(settings) < 1 {
		return nil
	}
	...
}

genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign.go
```diff
@@ -128,11 +128,13 @@ func readOutboundCampaign(ctx context.Context, d *schema.ResourceData, meta inte
 		// Use MaxCallsPerAgentDecimal if available, fall back to MaxCallsPerAgent for backward compatibility
 		if campaign.MaxCallsPerAgentDecimal != nil {
 			resourcedata.SetNillableValue(d, "max_calls_per_agent", campaign.MaxCallsPerAgentDecimal)
+			resourcedata.SetNillableValue(d, "max_calls_per_agent_decimal", campaign.MaxCallsPerAgentDecimal)
 		} else if campaign.MaxCallsPerAgent != nil {
 			// Convert int to float64 for backward compatibility with existing campaigns
 			maxCallsFloat := float64(*campaign.MaxCallsPerAgent)
 			_ = d.Set("max_calls_per_agent", maxCallsFloat)
 		}
+		resourcedata.SetNillableValue(d, "agent_owned_column", campaign.AgentOwnedColumn)
```
sureshperiyappan, line 137: There are verbose log.Printf statements for call_analysis_response_set_id in both the build function and the read function. These look like debugging aids from development. They'll produce noise in production logs for every campaign CRUD operation. Consider removing them or gating behind a debug flag.

genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_schema.go
```diff
@@ -153,6 +153,17 @@ func ResourceOutboundCampaign() *schema.Resource {
 				Type:         schema.TypeFloat,
 				ValidateFunc: validation.FloatAtLeast(1),
 			},
+			`max_calls_per_agent_decimal`: {
```
sureshperiyappan, line 156: The max_calls_per_agent_decimal field is Optional: true, Computed: true but there's no code path that reads it from the config during create/update — the build function only reads max_calls_per_agent. If a user sets max_calls_per_agent_decimal in their config, it will be silently ignored. Consider either:

Making it Computed: true only (read-only, populated from API response), or
Actually reading it in getOutboundCampaignFromResourceData and using it when set
If the intent is for users to use max_calls_per_agent and this field is just for exposing the API's decimal response, it should be read-only.

sureshperiyappan (approved):

PR #2323, opened 2026-05-12
feat(auth_role): Suppress wildcard permission diff and add deprecation warning

sanjaysg21:


genesyscloud/auth_role/resource_genesyscloud_auth_role.go
```diff
@@ -99,7 +105,16 @@ func readAuthRole(ctx context.Context, d *schema.ResourceData, meta interface{})
 
 	log.Printf("Reading role %s", d.Id())
 
-	return util.WithRetriesForRead(ctx, d, func() *retry.RetryError {
+	// Capture configured permission policies before reading from API to detect wildcard usage
```
bbbco, line 102: You should consider pulling this logic into the Schema validator or CustomizeDiff so that it can be handled at the Plan stage. Otherwise, there will be constant drift.

genesyscloud/auth_role/resource_genesyscloud_auth_role_utils.go
shrutisuryawanshigenesys: since we are using the new method flattenRolePermissionPoliciesWithWildcardSuppress, remove this if unused

genesyscloud/auth_role/resource_genesyscloud_auth_role.go
sanjaysg21: The suppression happens during Read, before Terraform compares state vs config. By the time the diff is computed, the state already has the wildcard preserved, so there's no drift.

genesyscloud/auth_role/resource_genesyscloud_auth_role_utils.go
sanjaysg21: done

sureshperiyappan (approved):

PR #2327, opened 2026-05-14
fix(routing_queue): remove self-referential queue_id from conditional group routing rules during export

sanjaysg21:
… group routing rules during export

### DIFF
--- a/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
+++ b/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
@@ -881,7 +881,7 @@ func RoutingQueueExporter() *resourceExporter.ResourceExporter {
 			"outbound_email_address": {"route_id"},
 			"members":                {"user_id"},
 		},
-		RemoveIfSelfReferential: []string{"direct_routing.backup_queue_id", "conditional_group_routing_rules.queue_id"},
+		RemoveIfSelfReferential: []string{"direct_routing.backup_queue_id", "conditional_group_routing_rules.queue_id", "conditional_group_activation.pilot_rule.conditions.simple_metric.queue_id", "conditional_group_activation.rules.conditions.simple_metric.queue_id"},
 		AllowZeroValues:         []string{"bullseye_rings.expansion_timeout_seconds"},
 		CustomAttributeResolver: map[string]*resourceExporter.RefAttrCustomResolver{
 			"bullseye_rings.member_groups.member_group_id":              {ResolveRefTypeFunc: resourceExporter.MemberGroupsResolver},
--- a/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
+++ b/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
@@ -601,7 +601,7 @@ func ResourceRoutingQueue() *schema.Resource {
 				Elem:        conditionalGroupActivationResource,
 			},
 			"conditional_group_routing_rules": {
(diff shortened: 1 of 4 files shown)


genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
```diff
@@ -880,7 +881,7 @@ func RoutingQueueExporter() *resourceExporter.ResourceExporter {
 			"outbound_email_address": {"route_id"},
 			"members":                {"user_id"},
 		},
-		RemoveIfSelfReferential: []string{"direct_routing.backup_queue_id"},
+		RemoveIfSelfReferential: []string{"direct_routing.backup_queue_id", "conditional_group_routing_rules.queue_id"},
```
sureshperiyappan: The CGA simple_metric.queue_id has the same "null means current queue" API behavior and the same self-referential export risk. The paths are already registered in RefAttrs (lines 877-878) as {RefType: ResourceType}, and FlattenCgaSimpleMetric unconditionally stores whatever the API returns without filtering self-references. The field also has Computed: true, meaning the API can populate it even when the user didn't set it.

Should we also add the CGA paths here?


RemoveIfSelfReferential: []string{
    "direct_routing.backup_queue_id",
    "conditional_group_routing_rules.queue_id",
    "conditional_group_activation.rules.conditions.simple_metric.queue_id",
    "conditional_group_activation.pilot_rule.conditions.simple_metric.queue_id",
},
Without this, if the API ever returns the queue's own ID in simple_metric.queue (rather than null), it would be exported as a literal ID — which would break re-imports into a different org, same issue this PR fixes for CGR.

bbbco: I agree with this recommendation!

genesyscloud/routing_queue_conditional_group_routing/resource_genesyscloud_routing_queue_conditional_group_routing_schema.go
```diff
@@ -70,7 +70,7 @@ func ResourceRoutingQueueConditionalGroupRouting() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"evaluated_queue_id": {
-							Description: "The queue being evaluated for this rule. For rule 1, this is always the current queue, so should not be specified.",
+							Description: "The queue being evaluated for this rule. For rule 1, this is always the current queue, so should not be specified. The API interprets a null queue as the current queue.",
```
bbbco, line 73: What is "rule 1"?

bbbco (changes requested):

genesyscloud/routing_queue_conditional_group_routing/resource_genesyscloud_routing_queue_conditional_group_routing_schema.go
sanjaysg21: "Rule 1" = the first rule in the list. The API enforces that the first rule always evaluates the current queue  you can't point it at a different queue (only rules 2-5 can). This is the API's own terminology. Happy to reword to "the first rule" if that's clearer.

genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
sanjaysg21: will do it

sureshperiyappan (approved):

genesyscloud/routing_queue_conditional_group_routing/resource_genesyscloud_routing_queue_conditional_group_routing_schema.go
bbbco: That sounds like a good idea. Ty!

bbbco (approved):

PR #2334, opened 2026-05-15
Feature [devtooling-1697] - Auto generate apis.md files in examples directory

bbbco:
## Description

Enhances the `apidocs` tool to automatically generate and maintain `apis.md` files for both resources and data sources by scanning proxy files and mapping SDK method calls to API endpoints via the Swagger specification's `operationId` field.

### Problem

Previously, `apis.md` files were manually maintained, leading to:
- Missing API endpoints that the resource actually uses
- Stale documentation when proxy files were updated
- No API documentation at all for most data sources
- Addendum content (migration notes, export behavior) mixed in with endpoint listings
- And the primary impetus: **the Scopes and Permissions generation was based on these endpoints, so these became outdated too**

### Solution

The `apidocs` tool now:
1. Builds an `operationId → (HTTP method, path)` reverse lookup map from the fetched Swagger
2. For each resource and data source, scans the proxy file to detect API calls via:
   - SDK method calls (e.g., `p.scriptsApi.GetScripts(...)`) matched against the operationId map
   - Raw HTTP calls with explicit `/api/v2/` paths
   - `BasePath + "/api/v2/..."` concatenation patterns
3. Automatically appends any missing endpoints to the `apis.md` file and sorts all endpoints for consistency
4. Breaks out content previously appended to some `apis.md` files into a `notes.md` files for addendum content and appends them after the permissions section in the generated docs
5. Processes both `docs/resources/` and `docs/data-sources/` in a single pass

### Key Changes

- **`apidocs/main.go`** — Rewrote to integrate proxy auditing into the main doc generation flow. Added `processDocsFolder`, `updateApisMdFromProxy`, `findSourceFile`, `detectEndpointsFromProxy`, `buildOperationIdMap`, and supporting functions. Removed `insertPermissionsAndScopes` (no longer needed).
- **`apidocs/main_test.go`** — Updated tests for all new functions.
- **`templates/data-sources.md.tmpl`** — New template with `**No APIs**` placeholder so data source docs get API endpoint injection.
- **`examples/README.md`** — Updated to document auto-generated `apis.md`, `notes.md`, and `.source` file conventions.
- **`examples/resources/*/notes.md`** — Extracted addendum content from 5 resources' `apis.md` files into dedicated `notes.md` files.
- **`examples/data-sources/*/.source`** — Created for edge cases where the tool can't auto-locate the proxy file due to naming mismatches.
- **`examples/resources/*/apis.md`** — Updated with missing endpoints detected by the proxy audit.
- **`examples/data-sources/*/apis.md`** — Created for data sources that previously had no API documentation.
- **`public/data/resource_permissions-latest.json`** — Removed from repo (should be generated as part of release).

### `.source` File Convention

For resources/data sources where the package name doesn't match the resource name (e.g., `genesyscloud_script` → `scripts/` package), a `.source` file in the examples directory contains a relative path to the correct file to scan:

\`\`\`
../../../genesyscloud/integration/genesyscloud_integration_proxy.go
\`\`\`

### Known Limitations

- SDK method names that don't have a matching `operationId` in the Swagger spec cannot be auto-mapped (requires manual `apis.md` seeding)
- Raw HTTP calls where the HTTP method is defined far from the path string (>200 chars) may not be detected
- The tool only scans the single file identified by `findSourceFile` — API calls made from utility files in the same package won't be detected

### AI Disclosure

This implementation was developed with assistance from Amazon Q Developer. The same approach can be reproduced by:
1. Reviewing the pinned context rules in `.amazonq/rules/` (project structure, guidelines, new-resource instructions)
2. Asking to "write a tool that audits proxy files against apis.md using the Swagger spec's operationId field"
3. Iterating on edge cases (naming mismatches, legacy files, regex patterns)


apidocs/main.go
```diff
@@ -347,48 +626,6 @@ func extractPermissionsAndScopes(endpoints []APIEndpoint, spec *SwaggerSpec) str
 	return result.String()
 }
 
-// insertPermissionsAndScopes inserts the permissions and scopes section before the first
-// markdown heading (##) in the content, or at the end if no headings are present
-func insertPermissionsAndScopes(content, permissionsAndScopes string) string {
```
bbbco, line 352: We no longer need to surgically insert the permissions into the APIs.md file since we're building the docs on the fly (see line 180)

public/data/resource_permissions-latest.json
bbbco, line 1: It's debatable whether we should include this file as just a generated artifact (such as the dependency tree) or keep it as part of the committed code. I feel like we should treat it the same as the dependency tree as they both serve similar functions. They should either both be generated or both live as part of the code, but not one or the other. I'm deleting the perms file from the code so it can be generated like we do with the dependency tree. If we want to store both in the code directly, we can open a new PR to bring both files back in.

apidocs/main.go
```diff
@@ -17,13 +17,23 @@ import (
 	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util/lists"
 )
 
+const (
+	swaggerURL              = "https://api.mypurecloud.com/api/v2/docs/swagger"
+	proxyFileGlob           = "genesyscloud/*/genesyscloud_*_proxy.go"
```
sanjaysg21: This is declared here but its never used/ The findSourceFile function builds its own glob patterns inline. If this is not used anywhere we can remove this

apidocs/main.go
```diff
@@ -167,22 +214,101 @@ func main() {
...
+	for i, ep := range allEndpoints {
+		anchor := buildAnchor(ep.Method, ep.Path)
+		if i > 0 {
+			buf.WriteString("\n")
 		}
+		buf.WriteString(fmt.Sprintf("* [%s %s](https://developer.genesys.cloud/devapps/api-explorer#%s)", strings.ToUpper(ep.Method), ep.Path, anchor))
+	}
+	buf.WriteString("\n")
+
+	ioutil.WriteFile(apisMdFile, []byte(buf.String()), 0644)
```
sanjaysg21: updateApisMdFromProxy ignores the error from ioutil.WriteFile .A write failure is silently swallowed. Should at least log a warning.

apidocs/main.go
sureshperiyappan: for _, file := range files {
    // ...
    docFile, err := os.OpenFile(fullFilePath, os.O_RDWR, 0666)
    // ...
    defer docFile.Close()  // ← BUG: won't close until processDocsFolder returns
defer is scoped to the enclosing function, not the loop iteration. If you have 500 files, you'll hold 500 open file descriptors until processDocsFolder returns. This was a pre-existing issue in the old code, but since this PR rewrote the loop, it should be fixed now.

Suggestion: Extract the file-writing logic into a helper function, or close explicitly:


func writeDocFile(fullFilePath string, docFileBytes, newBytes []byte) error {
    docFile, err := os.OpenFile(fullFilePath, os.O_RDWR, 0666)
    if err != nil { return err }
    defer docFile.Close()
    docFile.Truncate(0)
    _, err = docFile.WriteAt(newBytes, 0)
    return err
}

apidocs/main.go
sureshperiyappan: ioutil.WriteFile(apisMdFile, []byte(buf.String()), 0644)
This is the write that persists the auto-generated apis.md. If it fails, the function returns silently with no indication that the file wasn't updated.

apidocs/main.go
```diff
@@ -221,6 +347,207 @@ func writePermissionsJSON(permissions []ResourcePermissions, outputDir, filename
...
+	}
+	for _, c := range candidates {
+		if _, err := os.Stat(c); err == nil {
+			return c
+		}
+	}
+
+	// Strategy 2: Find a proxy file inside a matching sub-package directory
+	// Handles cases like script->scripts, routing_sms_address->routing_sms_addresses
+	matches, _ := filepath.Glob(filepath.Join("genesyscloud", pkgName+"*", "*_proxy.go"))
```
sureshperiyappan: matches, _ := filepath.Glob(filepath.Join("genesyscloud", pkgName+"*", "*_proxy.go"))
For pkgName = "routing", this would match routing_queue, routing_sms_addresses, routing_email_route, etc. — returning the first alphabetically. This could incorrectly point to the wrong proxy. Consider requiring the matched directory to be an exact prefix with an 's' or 'es' suffix, or adding a length-based heuristic.

apidocs/main.go
```diff
@@ -89,65 +96,105 @@ func main() {
...
+		} else {
+			outputFile := fmt.Sprintf("%s-%s.json", outputFilename, version)
+			fmt.Printf("Permissions data written to: %s/%s\n", outputDir, outputFile)
+		}
+	}
+}
+
+// processDocsFolder iterates over doc files in a folder, audits proxy files,
+// updates apis.md, and enhances docs with permissions and notes.
+func processDocsFolder(docsFolder, examplesFolder, apiDocsTag string, ignoredExamples []string, swaggerSpec *SwaggerSpec, opMap map[string]APIEndpoint, allPerms []ResourcePermissions, missingExamples *[]string) []ResourcePermissions {
```
sureshperiyappan: func processDocsFolder(docsFolder, examplesFolder, apiDocsTag string, ignoredExamples []string, 
    swaggerSpec *SwaggerSpec, opMap map[string]APIEndpoint, allPerms []ResourcePermissions, 
    missingExamples *[]string) []ResourcePermissions {
This is hard to read and easy to misuse. Consider a config struct:


type docsProcessorConfig struct {
    DocsFolder      string
    ExamplesFolder  string
    APIDocsTag      string
    IgnoredExamples []string
    SwaggerSpec     *SwaggerSpec
    OpMap           map[string]APIEndpoint
}

apidocs/main.go
```diff
@@ -221,6 +347,207 @@ func writePermissionsJSON(permissions []ResourcePermissions, outputDir, filename
 	return nil
 }
 
+// findSourceFile locates the proxy or source file for a given resource package name.
```
sureshperiyappan: This is the most complex function in the PR with 5 strategies and multiple fallback paths. It deserves its own test covering:

.source file present and valid
.source file present but target doesn't exist (fallthrough)
Exact package match
Pluralized package name (script → scripts)
Compact name match (external_contacts → externalcontacts)
Legacy file
No match found

apidocs/main.go
```diff
@@ -89,65 +96,105 @@ func main() {
...
+		if err := writePermissionsJSON(allResourcePermissions, outputDir, outputFilename, version); err != nil {
+			log.Printf("Error writing permissions JSON: %v", err)
+		} else {
+			outputFile := fmt.Sprintf("%s-%s.json", outputFilename, version)
+			fmt.Printf("Permissions data written to: %s/%s\n", outputDir, outputFile)
+		}
+	}
+}
+
+// processDocsFolder iterates over doc files in a folder, audits proxy files,
```
sureshperiyappan: The main orchestration function has no direct test. While TestUpdateApisMdFromProxy provides some integration coverage, there's no test that verifies:

The **No APIs** placeholder replacement
Notes content appended correctly
Permissions injected correctly
Ignored resources get their doc files removed

apidocs/main.go
```diff
@@ -89,65 +96,105 @@ func main() {
...
+// updates apis.md, and enhances docs with permissions and notes.
```
sureshperiyappan: Throughout processDocsFolder:

log.Printf("Warning: Failed to read folder...") (line 137)
fmt.Printf("Missing APIs file: %s\n", ...) (line 166)
fmt.Printf("Couldn't open file: %s\n", ...) (line 200)
Pick one pattern for warning/error output. Recommend log.Printf for warnings/errors so they go to stderr, reserving fmt.Printf for informational progress messages.

apidocs/main.go
sureshperiyappan: // Method to insert the contents of each resource's apis.md file into the markdown documentation
func main() {
This was accurate before the rewrite but now understates what main() does. It should describe the full scope: auditing proxy files, generating apis.md, injecting permissions, writing the permissions JSON, etc.

sureshperiyappan: The project's go.mod specifies go 1.25.0. The io/ioutil package was deprecated in Go 1.16 (5 years ago). All usages should be migrated:

ioutil.ReadDir → os.ReadDir
ioutil.ReadFile → os.ReadFile
ioutil.ReadAll → io.ReadAll
ioutil.WriteFile → os.WriteFile
This isn't just style — go vet and linters flag this, and it sends the wrong signal for a repo that other teams maintain.

apidocs/main.go
```diff
@@ -221,6 +347,207 @@ func writePermissionsJSON(permissions []ResourcePermissions, outputDir, filename
...
+				}
+			}
+		}
+	}
+
+	return endpoints
+}
+
+// detectHTTPMethodNearPath looks for HTTP method indicators near a raw path usage
+func detectHTTPMethodNearPath(content, pathMatch string) string {
```
sureshperiyappan: detectHTTPMethodNearPath uses strings.Index(content, pathMatch) which always resolves to the first occurrence of a given path string. For proxy files that reuse the same raw path across multiple functions (e.g., process_automation_trigger uses BasePath + "/api/v2/processAutomation/triggers/" + triggerId for GET, PUT, and DELETE), every call resolves to the same 400-char window and returns the same HTTP method — so only one of the three gets detected.

In practice this is mitigated because: (a) most resources use SDK method calls which are handled by Pattern 1 without this issue, and (b) the merge logic preserves any pre-existing manually-seeded endpoints. But for newly generated apis.md files on raw-HTTP-only resources, some endpoints could be missed.

A future improvement would be to pass the match's byte offset (from FindAllStringSubmatchIndex) into this function rather than re-searching with strings.Index. This would also eliminate the minor ordering bias in the method priority checks (DELETE > POST > PUT > PATCH > GET) since each call would be scoped to the correct occurrence's context.

Not a blocker — just flagging for awareness since the PR description already acknowledges this limitation.

apidocs/main.go
bbbco: Yep, this was leftover from a previous implementation. Ty

apidocs/main.go
bbbco: Thank you. I am extracting to helper function as suggested.

bbbco: Thank you. I am fixing.

bbbco: Ty

apidocs/main.go
bbbco: Thank you. I have added tests as recommended.

apidocs/main.go
bbbco: Ty. Done.

apidocs/main.go
bbbco: Added error handling to show warning that the file failed to write

bbbco: Ty

apidocs/main.go
bbbco: Thank you. I am updating test coverage as recommended

apidocs/main.go
bbbco: Yeah, I decided to refactor the findSourceFile approach with a more explicit strategy. We will define the source of the proxy files in a comment at the top of each apis.md example and use that file to generate endpoints.

apidocs/main.go
bbbco: We recently pulled in my PR to consolidate all non-SDK calls into the Custom API Client package. I have removed this function and replaced the other patterns with a pattern checker for using the Custom API Client.

.github/workflows/build-docs.yml
bbbco, line 1: Added this as a safety net. If someone pushes with stale docs, the diff check will fail and inform them to run `make docs` (something that is easy to forget to do)

examples/ignored_resources.go
```diff
@@ -4,9 +4,11 @@ import (
 	lists "github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util/lists"
 )
 
+// GetIgnoredResources returns a list of resources that should be ignored when generating the api docs. These are resource types that do not expect to have examples, and should not be included in public documentation
 func GetIgnoredResources() []string {
 	return []string{
 		"genesyscloud_bcp_tf_exporter",
+		"genesyscloud_externalusers_identity",
```
bbbco, line 11: It seems this was added as a oneoff for a particular customer's use case but never made truly public.

genesyscloud/provider/sdk_client_pool.go
```diff
@@ -941,8 +941,6 @@ func WrapResourceWithType(resourceType string, resource *schema.Resource) {
 		return
 	}
 
-	log.Printf("[DEBUG] WrapResourceWithType: Wrapping resource type=%q", resourceType)
```
bbbco, line 944: This log line was getting really noisy when running `make docs` and feels unnecessary (probably was good for initial development, but is now vestigial)

bbbco:
@sanjaysg21 @sureshperiyappan I made a major refactoring of the apidocs tooling for my logic of processing the apis.md files incorporating all of your suggestions. I decided that it was best for the apis.md files to indicate where their sources for calling the APIs were coming from (most of the proxy files) so they could define for themselves where to discover the endpoints, so it does this via a comment at the top of each `apis.md` file. This ensures the script doesn't have to guess where to look. I did not touch the core of the permissions logic, only how it is called. I did add tests for all of the behavior. 

I also added a github workflow to help ensure `make docs` gets run more consistently

Please re-run your review tools against this PR again. Ty!

sureshperiyappan:
@bbbco  why do we need this file apidocs/apidocs?

bbbco:
> @bbbco why do we need this file apidocs/apidocs?

Thanks for catching this. This was inadvertently committed when I ran `git add apidocs`

sureshperiyappan (approved):

PR #2337, opened 2026-05-20
NO-JIRA: Add export option to exclude_deprecated fields

bbbco:
We have noticed that there are several fields that are deprecated, but are being included in the exports by default. This can cause issues with BCP syncs.

This PR adds an option to exclude deprecated fields on export (for instance, the recently deprecated `genesyscloud_outbound_contact_list.phone_columns[*].callable_time_column` field)


genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -1892,6 +1894,11 @@ func (g *GenesysCloudResourceExporter) getResourcesForType(resType string, schem
 						delete(instanceState.Attributes, resAttribute)
 						continue
 					}
+					// Remove deprecated attributes if exclude_deprecated is set
```
sureshperiyappan: thanks for the contribution  the intent here is solid and this is definitely something we need. However, I found a critical issue with the current approach that means it won't actually solve the problem described in the PR (the nested callable_time_column field).

The issue:

The filtering loop iterates over res.Schema which only contains top-level attributes:

for resAttribute, resSchema := range res.Schema {
    if resSchema.Deprecated != "" && g.excludeDeprecated {
        delete(instanceState.Attributes, resAttribute)
        continue
    }
}
For the genesyscloud_outbound_contact_list resource, res.Schema contains "phone_columns"  but that block itself is NOT deprecated (Deprecated: ""). The actual deprecated field callable_time_column lives inside phone_columns.Elem.(*schema.Resource).Schema, which this loop never enters.

On top of that, for TypeSet/TypeList attributes, the flat state map doesn't have a literal key "phone_columns"  it stores keys like "phone_columns.#", "phone_columns.HASH.callable_time_column", etc. So even if a complex-type attribute were deprecated at the top level, delete(instanceState.Attributes, "phone_columns") would be a no-op.

I traced the full export pipeline (getResourcesForType → instanceStateToMap → sanitizeConfigMap → removeExcludedAttrsFromMap → generateOutputFiles) and confirmed that no downstream step checks Deprecated on nested schemas either.

Suggested fix:

We already have removeExcludedAttrsFromMap which recursively walks the nested JSON config map and removes attributes by path  it's battle-tested and handles all nesting levels. We should leverage that instead of operating on the flat state map.

Walk the schema tree once to collect deprecated paths, then feed them into the existing removal function:

// collectDeprecatedPaths walks the schema tree and returns all deprecated attribute paths.
// Time: O(S) where S = total schema nodes. Space: O(D) where D = deprecated attributes found.
func collectDeprecatedPaths(schemaMap map[string]*schema.Schema, prefix string) []string {
    var paths []string
    for name, s := range schemaMap {
        fullPath := name
        if prefix != "" {
            fullPath = prefix + "." + name
        }
        if s.Deprecated != "" {
            paths = append(paths, fullPath)
            continue // skip children of a deprecated block
        }
        if elem, ok := s.Elem.(*schema.Resource); ok {
            paths = append(paths, collectDeprecatedPaths(elem.Schema, fullPath)...)
        }
    }
    return paths
}
Then apply it in buildResourceConfigMap after sanitizeConfigMap, using the existing infrastructure:

if g.excludeDeprecated {
    deprecatedPaths := collectDeprecatedPaths(res.Schema, "")
    if len(deprecatedPaths) > 0 {
        removeExcludedAttrsFromMap(configMap, deprecatedPaths, "")
    }
}
This gives us:
	Correctness  works for nested attributes at any depth, not just top-level simple types.
	
	Efficiency  O(S) one-time schema walk + reuses the existing O(N×D) map removal. The result of collectDeprecatedPaths is deterministic per resource type, so it can be cached in a map[string][]string if needed for large exports.
	
	Minimal code  no new recursive logic needed for the actual removal; we reuse what's already proven.

The current delete(instanceState.Attributes, resAttribute) block in getResourcesForType should be removed entirely since it only partially works for top-level simple types and creates a false sense of coverage.

A word of caution:
	The export pipeline is a sensitive code path it touches every resource in an org during full exports and runs concurrently. Changes here can silently produce incorrect exports that only surface during BCP syncs or terraform apply in a target org. I'd strongly recommend adding unit tests that verify nested deprecated attributes are actually excluded (following the pattern in TestUnitGetResourceStateRemovesComputedAttributes), and ideally a targeted acceptance test with a real resource like genesyscloud_outbound_contact_list that has known nested deprecated fields.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -147,6 +147,7 @@ type GenesysCloudResourceExporter struct {
 	// 1-byte alignment
 	// .. Booleans
 	addDependsOn         bool
+	excludeDeprecated    bool
```
sureshperiyappan: This needs unit test coverage before merging. The codebase already has a clear pattern for this — see TestUnitGetResourceStateRemovesComputedAttributes which uses table-driven tests with schema definitions and expected state assertions. I'd expect similar tests here covering: a top-level deprecated attribute is removed when excludeDeprecated = true, it's kept when false, a nested deprecated attribute inside a TypeSet/TypeList block is removed, and non-deprecated attributes at all nesting levels remain unaffected. Given how sensitive the export path is, we really need these to catch regressions early.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -1892,6 +1894,11 @@ func (g *GenesysCloudResourceExporter) getResourcesForType(resType string, schem
 						delete(instanceState.Attributes, resAttribute)
 						continue
 					}
+					// Remove deprecated attributes if exclude_deprecated is set
+					if resSchema.Deprecated != "" && g.excludeDeprecated {
+						delete(instanceState.Attributes, resAttribute)
+						continue
```
sureshperiyappan: Would be good to add a tflog.Debug here so operators can troubleshoot when fields go missing from their exports. Something like logging the resource type and attribute path being excluded. The computed attribute removal doesn't log either, but since this is a new opt-in feature, giving users visibility into what's being stripped will save support headaches down the line.

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -196,6 +196,13 @@ func ResourceTfExport() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"exclude_deprecated": {
```
sureshperiyappan: Schema definition looks good — follows the same pattern as export_computed and other boolean flags (TypeBool, Optional, Default: false, ForceNew: true). Description with the future default change note is consistent with how we've done it elsewhere. 👍

docs/guides/export.md
```diff
@@ -102,6 +102,19 @@ In its standard setup, this Terraform configuration exports only the dependencie
 
 On the other hand, Terraform also provides the `exclude_attributes` option for instances where certain fields need to be omitted from an export. This, along with the ability to automatically export additional dependencies, contributes to Terraform’s flexible framework for managing resource exports. It allows for granular control over the inclusion or exclusion of elements in the export, ensuring that your exported configuration aligns precisely with your requirements.
 
+## Excluding Deprecated Attributes:
```
sureshperiyappan, line 105: Docs section is clear and well-structured. HCL example is helpful and the warning note about the future default change is appropriate. Looks good. 👍

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
```diff
@@ -196,6 +196,13 @@ func ResourceTfExport() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"export_deprecated": {
```
bbbco, line 199: Renamed this to align with the `export_computed` attribute that already exists

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
bbbco: Thanks for the suggestion. I have done this, but to clarify, I also pulled in the exclude computed logic, and I placed _before_ the sanitizeConfig in order so sanitizeConfig properly handles nullified attributes.

genesyscloud/tfexporter/resource_genesyscloud_tf_export.go
bbbco: Thanks. Just to be clear, I renamed this attribute to `export_deprecated` and inverted the logic to align with the `export_computed` attribute that already exists.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
bbbco: Thank you. I have done this.

genesyscloud/tfexporter/genesyscloud_resource_exporter.go
bbbco: Added unit tests for these items.

sureshperiyappan (approved):

PR #2339, opened 2026-05-21
Fix worktype_status reference resolution in exporter for workitem, worktype_status, and onattributechange_rule resources

sanjaysg21:


genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_schema.go
```diff
@@ -125,7 +125,13 @@ func TaskManagementWorktypeStatusExporter() *resourceExporter.ResourceExporter {
 	return &resourceExporter.ResourceExporter{
 		GetResourcesFunc: provider.GetAllWithPooledClient(getAllAuthTaskManagementWorktypeStatuss),
 		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
-			"worktype_id": {RefType: "genesyscloud_task_management_worktype"},
+			"worktype_id":                   {RefType: "genesyscloud_task_management_worktype"},
+			"destination_status_ids":        {RefType: ResourceType},
+			"default_destination_status_id": {RefType: ResourceType},
+		},
+		CustomAttributeResolver: map[string]*resourceExporter.RefAttrCustomResolver{
+			"destination_status_ids":        {ResolverFunc: WorktypeStatusArrayRefResolver("destination_status_ids")},
```
bbbco, line 138: I'm wondering if we should have a generic array ref resolver that could be wrapped to iterate over an array and resolve for each value so you didn't have to rewrite the same type of resolver and could just reuse the one you defined for default_destination_status_id. Something like

```
"destination_status_ids":        {ArrayResolverFunc: WorktypeStatusRefResolver("destination_status_ids")},
```

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_util.go
```diff
@@ -114,3 +115,106 @@ func ValidateStatusIds(statusResource1 string, key1 string, statusResource2 stri
 		return nil
 	}
 }
+
+// WorktypeStatusRefResolver resolves a bare status ID to a proper Terraform reference
```
sureshperiyappan, line 132: No unit tests for WorktypeStatusRefResolver or WorktypeStatusArrayRefResolver. These are pure functions with clear inputs and outputs — perfect candidates for table-driven tests. Can we add tests covering at least: happy path resolution, already-resolved values (starts with ${), nil/missing exporter, and status ID not found in the map? Without tests we have no safety net if someone changes the composite key format or the resolver logic later.

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_schema.go
```diff
@@ -125,7 +125,13 @@ func TaskManagementWorktypeStatusExporter() *resourceExporter.ResourceExporter {
 	return &resourceExporter.ResourceExporter{
 		GetResourcesFunc: provider.GetAllWithPooledClient(getAllAuthTaskManagementWorktypeStatuss),
 		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
-			"worktype_id": {RefType: "genesyscloud_task_management_worktype"},
+			"worktype_id":                   {RefType: "genesyscloud_task_management_worktype"},
```
sureshperiyappan, line 133: destination_status_ids and default_destination_status_id are declared in both RefAttrs and CustomAttributeResolver. The standard resolver runs first in sanitizeConfigArray, fails the direct map lookup (bare ID vs composite key), and adds the ID to buildSecondDeps as an "unresolved dependency." Then the custom resolver runs and actually resolves it. The export result is correct, but you're left with phantom entries in buildSecondDeps that trigger unnecessary work in the dependency chain resolution pass. Worth adding a comment here explaining why both are needed (dependency graph ordering via RefAttrs + actual resolution via CustomAttributeResolver), so the next person doesn't try to "clean up" one of them.

genesyscloud/task_management_worktype_flow_onattributechange_rule/resource_genesyscloud_task_management_onattributechange_rule_schema.go
```diff
@@ -92,9 +92,9 @@ func TaskManagementOnAttributeChangeRuleExporter() *resourceExporter.ResourceExp
 	return &resourceExporter.ResourceExporter{
 		GetResourcesFunc: provider.GetAllWithPooledClient(getAllAuthTaskManagementOnAttributeChangeRule),
 		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
-			"worktype_id": {RefType: "genesyscloud_task_management_worktype"},
-			"new_value":   {RefType: "genesyscloud_task_management_worktype_status"},
-			"old_value":   {RefType: "genesyscloud_task_management_worktype_status"},
+			"worktype_id":         {RefType: "genesyscloud_task_management_worktype"},
+			"condition.new_value": {RefType: "genesyscloud_task_management_worktype_status"},
```
sureshperiyappan: The condition.new_value and condition.old_value ref attrs assume the value is always a worktype status ID, but looking at flattenSdkCondition, these fields only contain status IDs when condition.attribute == "statusId". For other attribute types (e.g., priority, queue), the standard resolver will fail the lookup and silently add the value to buildSecondDeps as a phantom unresolved genesyscloud_task_management_worktype_status dependency. It won't break the export, but it's not clean. Have you considered using a CustomAttributeResolver with ResolveRefTypeFunc that checks the attribute field before attempting resolution? Or at minimum, does the on-attribute-change rule only support statusId as the attribute today?

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_util.go
```diff
@@ -114,3 +115,106 @@ func ValidateStatusIds(statusResource1 string, key1 string, statusResource2 stri
...
+		}
+
+		idMetaMap := exporter.SanitizedResourceMap
+		if idMetaMap == nil {
+			return nil
+		}
+
+		// Search for a composite key that ends with "/<statusId>"
+		suffix := "/" + statusId
+		for compositeId, meta := range idMetaMap {
```
sureshperiyappan: The suffix-search logic is copy-pasted between the two resolvers. Small thing, but extracting a shared helper like findMetaBySuffix(idMetaMap, statusId) would reduce duplication and make it easier to add logging or change the matching logic in one place later.

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_util.go
```diff
@@ -114,3 +115,106 @@ func ValidateStatusIds(statusResource1 string, key1 string, statusResource2 stri
...
+
+		resolved := make([]interface{}, 0, len(arr))
+		for _, item := range arr {
+			statusId, ok := item.(string)
+			if !ok || statusId == "" {
+				resolved = append(resolved, item)
+				continue
+			}
+
+			// If already resolved to a reference, keep it
```
sureshperiyappan, line 201: The suffix-search logic is copy-pasted between the two resolvers. Small thing, but extracting a shared helper like findMetaBySuffix(idMetaMap, statusId) would reduce duplication and make it easier to add logging or change the matching logic in one place later.

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_schema.go
sanjaysg21: Good idea for a future improvement. For now the array resolver is self-contained and only used in one place, so I'd prefer to keep the scope of this PR focused on the bug fix. If more resources need this pattern we can extract a generic wrapper then.

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_schema.go
sanjaysg21: Good catch. Added a comment explaining why both are needed. RefAttrs is required here because destination_status_ids is a string array processed by sanitizeConfigArray, which only checks RefAttrs for reference resolution — there's no CustomAttributeResolver hook inside that function. Without the RefAttrs entry, the IDs would just get escaped and never added to buildSecondDeps, meaning the referenced status resources wouldn't be tracked for dependency export. The phantom entries in buildSecondDeps are a known tradeoff — the custom resolver corrects the actual values afterward, but we need RefAttrs to ensure the dependency graph is complete.

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_util.go
sanjaysg21: agreed added the test cases

genesyscloud/task_management_worktype_flow_onattributechange_rule/resource_genesyscloud_task_management_onattributechange_rule_schema.go
sanjaysg21: added custom attribute resolver to handle this

sureshperiyappan (approved):

PR #2341, opened 2026-05-25
feat: make routing skills divisionable

eclancy404:
- add a new optional field for division ids on routing skills
- add update functionality for updating the division id
- add testing
- update the platform go sdk version


sureshperiyappan:
@eclancy404 can you update you branch with latest dev branch and resolve all the conflict and update the PR.

genesyscloud/routing_skill/resource_genesyscloud_routing_skill_schema.go
```diff
@@ -56,6 +57,11 @@ func ResourceRoutingSkill() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 			},
+			"divisionId": {
```
sureshperiyappan: "divisionId": {
The field must be named division_id (snake_case). Every single resource in this provider (30+ resources) uses division_id. Terraform convention mandates snake_case for schema attributes. This will break exporter reference resolution and confuse users who work with other resources.

Also applies to all references in resource_genesyscloud_routing_skill.go:

d.Get("divisionId") → d.Get("division_id")
d.Set("divisionId", ...) → d.Set("division_id", ...)

genesyscloud/routing_skill/resource_genesyscloud_routing_skill_schema.go
```diff
@@ -56,6 +57,11 @@ func ResourceRoutingSkill() *schema.Resource {
...
+				Description: "Division of a skill. Changing the division will cause the skill object object to dropped and recreated",
+				Type:        schema.TypeString,
+				ForceNew:    true,
```
sureshperiyappan: If divisionId is ForceNew: true, Terraform will destroy and recreate the resource when the division changes. The UpdateContext handler will never be triggered for this field. You need to pick one approach:

Option A (division change = recreation): Keep ForceNew: true, remove UpdateContext entirely (since name is also ForceNew, there's nothing to update).
Option B (division change = in-place update via PATCH): Remove ForceNew: true, add Optional: true, Computed: true (matching the routing_wrapupcode pattern), and keep the UpdateContext.
Given this PR adds PatchRoutingSkill support, Option B seems to be the intent. Please clarify and fix.

genesyscloud/routing_skill/resource_genesyscloud_routing_skill.go
```diff
@@ -121,3 +150,14 @@ func GenerateRoutingSkillResource(
 	}
 	`, resourceLabel, name)
 }
+
+func GenerateUpdateRoutingSkillResource(
+	resourceLabel string,
+	id string,
+	divisionId string) string {
+	return fmt.Sprintf(`resource "genesyscloud_routing_skill" "%s" {
```
sureshperiyappan: return fmt.Sprintf(`resource "genesyscloud_routing_skill" "%s" {
    id = "%s",
    divisionId = "%s"
}
`, resourceLabel, id, divisionId)
Problems:

id cannot be set in a resource block — it's always a computed attribute in Terraform.
Commas are not valid HCL separators between attributes (this is not JSON).
The name field is Required but missing from this config.
This will produce a terraform config validation error at test time. The test helper should look like:


func GenerateRoutingSkillResourceWithDivision(resourceLabel, name, divisionId string) string {
    return fmt.Sprintf(`resource "genesyscloud_routing_skill" "%s" {
        name        = "%s"
        division_id = "%s"
    }
    `, resourceLabel, name, divisionId)
}

genesyscloud/routing_skill/resource_genesyscloud_routing_skill_test.go
```diff
@@ -17,6 +17,8 @@ func TestAccResourceRoutingSkillBasic(t *testing.T) {
 	var (
 		skillResourceLabel1 = "test-skill1"
 		skillName1          = "Terraform Skill" + uuid.NewString()
+		skillId1            = uuid.NewString()
```
sureshperiyappan: skillId1 = uuid.NewString()
skillDivision1 = uuid.NewString()
The test creates random UUIDs for id and divisionId that don't correspond to any real resource or division. The update step will fail with an API error since the division doesn't exist.
The update step no longer includes name (a Required field), so Terraform validation will fail before even reaching the API.
Checking resource.TestCheckResourceAttr(..., "id", skillId1) makes no sense — the id is assigned by the API on create, not by the test.
The test should:

Create a real skill in step 1 (already done).
Create a real division (or use a data source for an existing one).
Update the skill to move it to that division in step 2.
Verify division_id is set correctly.

genesyscloud/routing_skill/resource_genesyscloud_routing_skill_schema.go
sureshperiyappan: The exporter needs a RefAttrs entry so that terraform export properly resolves division references:


func RoutingSkillExporter() *resourceExporter.ResourceExporter {
        GetResourcesFunc: provider.GetAllWithPooledClient(GetAllRoutingSkills),
            "division_id": {RefType: "genesyscloud_auth_division"},
        },
    }
}
Every other resource with a division_id has this. Without it, exported configs will contain hardcoded division UUIDs instead of references.

genesyscloud/routing_skill/resource_genesyscloud_routing_skill_schema.go
```diff
@@ -56,6 +57,11 @@ func ResourceRoutingSkill() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 			},
+			"divisionId": {
+				Description: "Division of a skill. Changing the division will cause the skill object object to dropped and recreated",
```
sureshperiyappan: "Division of a skill. Changing the division will cause the skill object object to dropped and recreated"

Double "object" and missing "be". Should be:

"The division to which this skill belongs. If not set, the home division will be used."

(Also, if you go with Option B and remove ForceNew, the "dropped and recreated" language is wrong.)

genesyscloud/routing_skill/resource_genesyscloud_routing_skill_schema.go
eclancy404: I've gone with option B

genesyscloud/routing_skill/resource_genesyscloud_routing_skill_test.go
eclancy404: I've updated based on help from kiro, hopefully this makes more sense now

sureshperiyappan (approved):

PR #2346, opened 2026-05-28
DEVTOOLING-1571 - When trying to export "journey" object an exception is no longer raised

daragh-kearney:
- Fixes crash when reading/exporting journey action maps with Open action media type
- open_action was set to *[]map instead of a value compatible with TypeSet
- Use stringmap.SetValueIfNotNil like other nested fields in this resource
- Added unit test TestFlattenActionMapOpenAction
- Also relaxed TestMain auth handling to match architect_ivr (unit tests run without credentials)


sureshperiyappan:
Merge the dev branch latest ode into you branch and run the make docs and update the document.

sureshperiyappan (commented):
update the doc files

sureshperiyappan (approved):

PR #2347, opened 2026-06-01
OBR-723: Add time-based conditions to campaign rules

obolobova:
Adds time-based and duration-based condition types to `genesyscloud_outbound_campaignrule`.

These conditions require Advanced Rules (`campaign_rule_processing = "v2"`) and are used inside `condition_groups`.

## New fields

**On the resource:**
- `time_zone_id` — timezone for evaluating date/time conditions (defaults to UTC)

**On conditions (`condition_groups.conditions`):**
- `date_time_parameters` — configures timeOfDay, dayOfWeek, dayOfMonth, specificDate, weekDayOfMonth sub-conditions
- `campaign_run_time_settings` — controls whether waiting time counts toward campaign run duration
- `campaign_wait_time_settings` — specifies which wait state to monitor (Agents/Contacts/Lines)

**On condition parameters:**
- `for_duration` — requires the condition to remain true for N seconds before triggering

## Design notes

- Condition and action parameters use a shared helper (`campaignRuleParameterSchema`) with options to control which fields are included. Actions do not expose `for_duration` or date/time operators.
- Provider validates condition configurations at plan time (CustomizeDiff) and apply time (Create/Update). Invalid combinations — such as date/time operators on non-date conditions, or v2-only blocks in legacy mode — produce clear errors before reaching the API.
- `for_duration` inside TypeSet `parameters` is not visible to CustomizeDiff and is validated in Create/Update instead.

## Testing

- 1 positive acceptance test (create + import)
- 7 negative tests covering invalid operator/field/block combinations

Resolves: OBR-723
Depends on: OBR-213 (PR #2185)


sureshperiyappan:
@obolobova can you please update your branch with latest dev branch, run make docs and update PR for review.

obolobova:
> @obolobova can you please update your branch with latest dev branch, run make docs and update PR for review.

@sureshperiyappan done.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_schema.go
```diff
@@ -128,6 +128,255 @@ var (
 	}
 )
 
+var (
+	outboundCampaignRuleWeekDayOfMonth = &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			`day_of_week`: {
```
sureshperiyappan, line 144: resource_genesyscloud_outbound_campaignrule_schema.go — Line 134, 139, 144

day_of_week, month, and occurrence are missing ValidateFunc.

day_of_week is Required and the valid range is 1–7, but there's nothing
stopping a user from passing 0 or 8 at plan time.

month should be validation.IntBetween(1, 12)
day_of_week should be validation.IntBetween(1, 7)
occurrence is trickier — valid values are 1, 2, 3, 4, or -1 (last).
Use a custom ValidateFunc for that one.

Without these, the API returns the error instead of Terraform catching it
at plan time.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_utils.go
```diff
@@ -423,5 +527,242 @@ func flattenRuleParameters(params *platformclientv2.Campaignruleparameters) []in
...
+
+func buildWeekDayOfMonth(m map[string]interface{}) *platformclientv2.Campaignruleweekdayofmonth {
+	sdk := &platformclientv2.Campaignruleweekdayofmonth{}
+	if v, ok := m["day_of_week"].(int); ok && v != 0 {
+		sdk.DayOfWeek = &v
+	}
+	if v, ok := m["month"].(int); ok && v != 0 {
+		sdk.Month = &v
+	}
+	if v, ok := m["occurrence"].(int); ok && v != 0 {
```
sureshperiyappan, line 657: resource_genesyscloud_outbound_campaignrule_utils.go — Line 643–644
if v, ok := m["occurrence"].(int); ok && v != 0 {

occurrence = -1 means "last occurrence" and is a valid API value.
The v != 0 guard correctly passes -1 through, but there's no validation
to prevent someone setting occurrence = 0, which is invalid and would be
silently dropped rather than errored.

This is fine if you add ValidateFunc on the schema field (see schema.go
comment above), but leaving both without it is a silent data loss bug.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_utils.go
```diff
@@ -10,6 +11,66 @@ import (
...
+
+	// Reuse the same validation as CustomizeDiff for legacy conditions
+	conditions := d.Get("campaign_rule_conditions").([]interface{})
+	return validateNoTimeBasedConditionsInLegacyFromResourceData(conditions)
+}
+
+// validateNoTimeBasedConditionsInLegacyFromResourceData is the apply-time version that handles
+// ResourceData types (*schema.Set for parameters) vs CustomizeDiff types ([]interface{}).
+func validateNoTimeBasedConditionsInLegacyFromResourceData(conditions []interface{}) error {
+	v2OnlyConditionTypes := map[string]bool{
```
sureshperiyappan: resource_genesyscloud_outbound_campaignrule_utils.go — Line 32 and resource_genesyscloud_outbound_campaignrule_schema.go — Line 445

v2OnlyConditionTypes is defined twice — once in schema.go (line 445)
inside validateNoTimeBasedConditionsInLegacy, and again in utils.go
(line 32) inside validateNoTimeBasedConditionsInLegacyFromResourceData.

They have to stay in sync. If a new v2-only condition type gets added
later, someone will update one and miss the other.

Extract it to a package-level var so both functions reference the same
source of truth.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_schema.go
```diff
@@ -449,6 +1069,12 @@ func ResourceOutboundCampaignrule() *schema.Resource {
 				Default:     false,
 				Type:        schema.TypeBool,
 			},
+			`time_zone_id`: {
```
sureshperiyappan, line 1087: resource_genesyscloud_outbound_campaignrule_schema.go — Line 1072
`time_zone_id`: {
    Optional:    true,
    Computed:    true,
},

No format validation on time_zone_id. Any string is accepted at plan
time — the API will reject an invalid IANA timezone but only at apply
time, which is a poor UX.

Even a simple ValidateFunc that checks the string isn't empty and
contains a "/" (e.g. "America/New_York") would catch most typos early.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_test.go
```diff
@@ -1270,6 +1271,438 @@ data "genesyscloud_auth_division_home" "home" {}
 	})
 }
 
+func TestAccResourceOutboundCampaignRuleTimeBasedConditions(t *testing.T) {
```
sureshperiyappan, line 1274: resource_genesyscloud_outbound_campaignrule_test.go — Line 1274
func TestAccResourceOutboundCampaignRuleTimeBasedConditions(t *testing.T) {
Missing t.Parallel(). Every other new test in this PR has it. If this
one is intentionally serial (e.g. shared org resource conflict), add a
comment explaining why. Otherwise add t.Parallel().

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_test.go
```diff
@@ -1270,6 +1271,438 @@ data "genesyscloud_auth_division_home" "home" {}
...
+
+		// Campaign trigger condition (required alongside time-based)
+		condProgressOp  = "greaterThan"
+		condProgressVal = "0.5"
+
+		timeZoneId = "America/New_York"
+	)
+
+	// Time-based condition: timeOfDay between 09:00 and 17:00
+	timeOfDayCondition := `
```
sureshperiyappan: resource_genesyscloud_outbound_campaignrule_test.go — Line 1317
timeOfDayCondition := `
    conditions {
        condition_type = "timeOfDay"
        ...
    }`

This is the only acceptance test for all the new date/time condition
types, and it only covers timeOfDay + between operator.

The new code adds 7 condition types (timeOfDay, dayOfWeek, dayOfMonth,
specificDate, weekDayOfMonth, campaignRunTime, campaignWaitTime) and
for_duration. None of the others are acceptance-tested.

weekDayOfMonth has the most complex nested schema (3 levels deep) and
is the highest risk. At minimum add a test for it, and one positive test
that shows for_duration working end-to-end.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_test.go
```diff
@@ -1270,6 +1271,438 @@ data "genesyscloud_auth_division_home" "home" {}
...
+		conditions {
```
sureshperiyappan: resource_genesyscloud_outbound_campaignrule_test.go — Line 1317
timeOfDayCondition := `
    conditions {
        condition_type = "timeOfDay"

Raw HCL string here while everything else in the test file uses
generate helper functions. This should either use an existing helper
or have a new generateDateTimeConditionBlock(...) helper added
alongside the others at the bottom of the file.

Consistency matters — the raw string is easy to typo and hard to reuse
in future tests.

genesyscloud/outbound_campaignrule/resource_genesyscloud_outbound_campaignrule_test.go
obolobova: The positive for_duration end-to-end test is written but blocked and commented out.
The test is blocked by a Go SDK bug (Duration serialized as object, not ISO-8601 - results in 400 API error).
Tracked in [DEVTOOLING-1759](https://inindca.atlassian.net/browse/DEVTOOLING-1759), commented out until fixed. PR moved to draft.

All other comments addressed.

sureshperiyappan:
@obolobova can you please update your branch with latest dev branch, run make docs and update PR for review. Since there are many changes there including the platform version, so please update it.

sureshperiyappan (approved):

PR #2353, opened 2026-06-04
MRMO-422 export source

jacob-shaw-genesys:
Changes

Moves from singleton export configuration to per worker configuration for MRMO exports


public/data/resource_permissions-latest.json
bbbco, line 1: Please use `git rm --cached` to remove this file from your branch, as it is being generated with each new release now.

genesyscloud/tfexporter/mr_mo_utils.go
```diff
@@ -198,11 +277,18 @@ func (g *GenesysCloudResourceExporter) ExportForMrMo(resType, resourceId string,
...
+	if g.resourcesExportedForMrMo == nil {
+		resourceDataList = make([]*schema.ResourceData, 0)
+	} else {
+		resourceDataList = (*g.resourcesExportedForMrMo)[resType]
+	}
+
+	return &MrMoExportByTypeResponse{
 		Config: util.JsonMap{
-			"resource": g.resourceTypesMaps,
+			"resources": g.resourceTypesMaps,
```
charliecon: ExportForMrMo returns this as `resource` , not `resources`. Just mentioning that in case this isn't an intentional difference

charliecon (approved):

BrianMoyles (approved):

PR #2355, opened 2026-06-04
DEVTOOLING-1565 - exporting genesyscloud_routing_email_route now produces export that is plannable 

daragh-kearney:
reply_email_address is no longer malformed:

- When a route uses itself as the reply address, export was emitting both domain_id and self_reference_route = true, which fails terraform plan
- Updated ReplyEmailAddressSelfReferenceRouteExporterResolver to clear domain_id (and route_id) when a self-reference is detected, leaving only self_reference_route = true
- Registered the resolver on reply_email_address.route_id as well as self_reference_route, so it runs after route_id is resolved (avoids map iteration ordering issues)
- Added/updated unit tests


genesyscloud/resource_exporter/resource_exporter_custom_test.go
```diff
@@ -121,3 +122,33 @@ func TestUnitRuleSetPropertyGroup(t *testing.T) {
 	}
 
 }
+
+// TestUnitReplyEmailAddressSelfReferenceRouteExporterResolver verifies DEVTOOLING-1565: when a route
+// self-references as its reply address, the export resolver must set self_reference_route=true and
+// clear route_id and domain_id. Leaving domain_id in place conflicts with self_reference_route
+// and causes terraform plan to fail.
+func TestUnitReplyEmailAddressSelfReferenceRouteExporterResolver(t *testing.T) {
+	resourceLabel := "procurementoperations_dell_com"
```
bbbco: Please remove any customer specific references/values

bbbco (changes requested):

docs/resources/greeting_user.md
sureshperiyappan, line 24: The doc adds a duplicate "Permissions and Scopes" section. The file already has "Permissions and Scopes" content above (lines with users, users:readonly). Now there are two separate sections — one with users/users:readonly and another with greetings/greetings:readonly. This might be intentional (adding missing scopes), but having two identically-titled sections looks like a doc formatting issue. Recommend merging into a single section or clarifying the distinction.

bbbco: Rebasing against the `dev` branch and running `make docs` should take care of this issue

sureshperiyappan (approved):

PR #2367, opened 2026-06-10
RULES-1687: improve rollback after terraform timeout, increase default timeout

brlenn:
Found a couple of issues around timeouts and rollback while creating large decision tables

Default terraform timeout of 20 mins is not enough for large tables, this had a knock on effect that cleanup of failed resources would not run due to `giving up after 1 attempt(s): context deadline exceeded`.


rmctigue (approved):

genesyscloud/business_rules_schema/resource_genesyscloud_business_rules_schema.go
```diff
@@ -105,7 +105,12 @@ func readBusinessRulesSchema(ctx context.Context, d *schema.ResourceData, meta i
 		}
 
 		resourcedata.SetNillableValue(d, "name", schema.Name)
-		resourcedata.SetNillableValue(d, "description", schema.JsonSchema.Description)
+		// Only set description when there is a real value; SetNillableValue's nil branch
+		// (d.Set(key, nil)) would coerce an absent description to "" for a TypeString,
```
bbbco: There are likely other cases where a `SetStringValueIfNotNil()` might be the correct approach to take for a given field. Could you extract that logic to a new shared function in `resourcedata` so it can be useful elsewhere?

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_schema.go
```diff
@@ -148,15 +148,21 @@ func valueSchemaFunc() *schema.Resource {
 func defaultsToSchemaFunc() *schema.Resource {
 	return &schema.Resource{
 		Schema: map[string]*schema.Schema{
+			// value/values/special are mutually exclusive; a column sets exactly one.
+			// Computed lets the provider own the unused siblings (which the legacy SDK
+			// materializes as empty) so plan(null) -> apply("") is not flagged as an
+			// inconsistency and does not produce a phantom ForceNew diff.
```
bbbco, line 154: Should you use a `ConflictsWith` attribute to enforce the mutual exclusivity?

docs/resources/business_rules_decision_table.md
```diff
@@ -623,5 +623,8 @@ Optional:
 
 Optional:
 
+- `create` (String)
```
bbbco, line 629: Don't modify the docs directly, as these are auto generated. Instead, use the `apis.md`, `resource.tf` and `notes.md` files in the examples directory to generate these docs.

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_schema.go
brlenn: We tried `ConflictsWith` back when the resource was first added, and for completeness we tried again here - it still doesn't work. It fails `InternalValidate`:
`ConflictsWith: value configuration block reference (columns.0.inputs.0.defaults_to.0.values) can only be used with TypeList and MaxItems: 1 configuration blocks`.
The SDK resolves `ConflictsWith` paths from the resource root and only through `TypeList` + `MaxItems:1` blocks, but `inputs`/`outputs` are variable-length lists, and `defaults_to` is a shared `Elem` used by both - so there's no valid static path to express the constraint. Mutual exclusivity is enforced server-side by the API, so we kept the `Optional+Computed` approach.

docs/resources/business_rules_decision_table.md
brlenn: This had been modified previously by https://github.com/MyPureCloud/terraform-provider-genesyscloud/pull/2314/changes I was following that,  I had ran the `go generate` command and it adds this

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_schema.go
bbbco: Oh right. I forgot about that constraint when I suggested it. Sounds good to keep it enforced server-side by the API. Can we update the attribute descriptions to be explicit about these known constraints?

docs/resources/business_rules_decision_table.md
bbbco: Should not run `go generate` on its own. Should be running `make docs` which runs `go generate` with other stuff.

brlenn: Wasn't aware of `make docs` previously told go generate was sufficent

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_schema.go
brlenn: The parent object has this set - https://github.com/MyPureCloud/terraform-provider-genesyscloud/blob/main/genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_schema.go#L190

bbbco: Understood. But its also good to have it clarified within the Attribute description that one field conflicts with another on the API side.

docs/resources/business_rules_decision_table.md
bbbco: https://github.com/MyPureCloud/terraform-provider-genesyscloud/#documentation-generation

brlenn: Thanks for the link, these lines thought aren't generated from apis.md/resource.tf/notes.md - they come from the Go schema via `{{ .SchemaMarkdown }}` in templates/resources.md.tmpl. The examples files only feed the Example Usage and API Usage sections; the entire Schema section (attributes, nested blocks, and the timeouts block) is derived directly from the schema. So the correct place to change them is the schema, which I did, then regenerated with `make docs`.

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_schema.go
brlenn: I have updated the descriptions and regenerated the docs

rmctigue (approved):

docs/resources/business_rules_decision_table.md
bbbco: Exactly. 

Oh, I see why these were updated now. They are displaying the Timeouts section, and since you added explicit timeouts per action type in the schema, it updated these docs.

bbbco (approved):

sanjaysg21 (approved):

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table.go
```diff
@@ -222,8 +292,15 @@ func addRowsToVersion(ctx context.Context, proxy *BusinessRulesDecisionTableProx
 		}
 
 		// Add the row to the version
-		_, err = proxy.createDecisionTableRow(ctx, tableId, version, &sdkRow)
+		resp, err := proxy.createDecisionTableRow(ctx, tableId, version, &sdkRow)
 		if err != nil {
+			if isDuplicateRowAtIndex(resp, i+1) {
```
sureshperiyappan, line 306: Nice fix for the 504-ghost-row scenario. One thing — applyRowChanges in the utils file also calls proxy.createDecisionTableRow when adding new rows during an update, but doesn't have this same isDuplicateRowAtIndex guard. The same gateway-timeout-then-409 situation can happen there too. Could we apply this guard to that path as well, or add a comment explaining why it's not needed there?

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table.go
```diff
@@ -109,6 +119,20 @@ func createBusinessRulesDecisionTable(ctx context.Context, d *schema.ResourceDat
 	return readBusinessRulesDecisionTable(ctx, d, meta)
 }
 
+// rollbackDecisionTable deletes a partially-created decision table on a fresh,
+// detached context. The create request context may already be expired/cancelled
+// (e.g. on a create timeout), which would prevent the cleanup DELETE from being
+// sent and leave the table orphaned in the org.
+func rollbackDecisionTable(tableId string, proxy *BusinessRulesDecisionTableProxy) {
```
sureshperiyappan: Minor suggestion: if the rollback DELETE fails, the user currently only sees the original error ("Failed to add rows" / "Failed to publish version") with no indication that a partial table was left behind in their org. The [WARN] log is good for debug logs, but won't surface in normal Terraform output.

Would it be worth returning an error from this function so the caller can enrich the diagnostic? Something like:

func rollbackDecisionTable(tableId string, proxy *BusinessRulesDecisionTableProxy) error {
    ...
    if _, derr := proxy.deleteBusinessRulesDecisionTable(cleanupCtx, tableId); derr != nil {
        log.Printf("[WARN] rollback delete failed for decision table %s: %s", tableId, derr)
        return derr
    }
    ...
    return nil
}
Then at the call site:

if rollbackErr := rollbackDecisionTable(tableId, proxy); rollbackErr != nil {
    return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf(
        "Failed to add rows: %s; additionally, cleanup of table %s failed — manual deletion may be required: %s",
        err, tableId, rollbackErr), nil)
}
Not a blocker - just thinking about the ops experience when this runs in CI and nobody's checking debug logs.

genesyscloud/business_rules_schema/resource_genesyscloud_business_rules_schema.go
```diff
@@ -105,7 +105,9 @@ func readBusinessRulesSchema(ctx context.Context, d *schema.ResourceData, meta i
 		}
 
 		resourcedata.SetNillableValue(d, "name", schema.Name)
-		resourcedata.SetNillableValue(d, "description", schema.JsonSchema.Description)
+		// Preserve null for an absent description; SetNillableValue would coerce it
+		// to "" for a TypeString and cause a null -> "" plan inconsistency.
+		resourcedata.SetStringValueIfNotNil(d, "description", schema.JsonSchema.Description)
```
sureshperiyappan: This will panic if schema.JsonSchema is nil — the if schema.JsonSchema != nil check on line 94 only guards the properties extraction block above, but this line sits outside that guard.

Since the code already acknowledges JsonSchema can be nil (the else branch on line 103 sets schemaPropsPtr = nil), this needs protection too:

if schema.JsonSchema != nil {
    resourcedata.SetStringValueIfNotNil(d, "description", schema.JsonSchema.Description)
}
Or move it inside the existing if schema.JsonSchema != nil block above.

brlenn: Done

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table.go
brlenn: Done, added this log message,

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table.go
brlenn: Done added the duplicate guard

sureshperiyappan (approved):

PR #2368, opened 2026-06-11
SQUONK-941: Launcher button changes in webdeployments_configuration

Ajith-26:
**Summary**
Add support for display_type and icon fields on the launcher_button block in genesyscloud_webdeployments_configuration messenger settings. Also add MessengerLauncherButtonText as a valid key for localized_labels.

**Changes**
Schema: Added display_type (enum: Icon, Text, IconAndText) and icon (with nested url) as optional fields on launcher_button
Build function: Updated buildMessengerSettings to pass display_type and icon to the SDK struct when provided by the customer
Flatten function: Updated flattenLauncherButton to read DisplayType and Icon from the API response
Localized labels: Added MessengerLauncherButtonText to the allowed values for the key field in localized_labels
Example: Updated resource.tf with display_type and icon usage

**Behavior**
All new fields are optional. If not set by the customer, they are not sent to the backend — no defaults are applied by the provider.

**⚠️ SDK Version Dependency**
This PR does not build on the current SDK version (platform-client-sdk-go/v188). The new fields (Launcherbuttonsettings.DisplayType, Launcherbuttonsettings.Icon, MessengerLauncherButtonText) are available in v191.

**Requesting the terraform team to bump the SDK to github.com/mypurecloud/platform-client-sdk-go/v191 v191.0.0** - once that's done, will verify below commands

- go vet
- go mod tidy
- go generate


genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
```diff
@@ -205,9 +220,21 @@ func flattenLauncherButton(settings *platformclientv2.Launcherbuttonsettings) []
...
-	}}
+	}
+
+	if settings.DisplayType != nil {
+		ret["display_type"] = *settings.DisplayType
+	}
+
+	if settings.Icon != nil {
+		ret["icon"] = []interface{}{map[string]interface{}{
+			"url": settings.Icon.Url,
```
sureshperiyappan: settings.Icon.Url is stored as a raw pointer here. Since Url is a *string in the SDK, this could panic if the API returns an Icon struct with a nil Url. You already nil-check Icon itself — just add the same safety for Url and dereference it, like you did for display_type above:


if settings.Icon != nil && settings.Icon.Url != nil {
    ret["icon"] = []interface{}{map[string]interface{}{
        "url": *settings.Icon.Url,
    }}
}

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
```diff
@@ -32,21 +32,21 @@ var (
...
 						"key": {
-							Description:  "Contains localized label key used in messenger homescreen or push notification.",
+							Description:  "Contains localized label key used in messenger homescreen, push notification, or messenger launcher button.",
 							Type:         schema.TypeString,
 							Required:     true,
-							ValidateFunc: validation.StringInSlice([]string{"MessengerHomeHeaderTitle", "MessengerHomeHeaderSubTitle", "PushNotificationTitle", "PushNotificationBody"}, false),
+							ValidateFunc: validation.StringInSlice([]string{"MessengerHomeHeaderTitle", "MessengerHomeHeaderSubTitle", "PushNotificationTitle", "PushNotificationBody", "MessengerLauncherButtonText"}, false),
 						},
 						"value": {
 							Description: "Contains localized label value used in messenger homescreen or push notification",
```
sureshperiyappan: You updated the descriptions for language, localized_labels, and key to mention "messenger launcher button", but the value field description still says "messenger homescreen or push notification". Should be consistent.

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
```diff
@@ -205,9 +220,21 @@ func flattenLauncherButton(settings *platformclientv2.Launcherbuttonsettings) []
 		return nil
 	}
 
-	return []interface{}{map[string]interface{}{
+	ret := map[string]interface{}{
 		"visibility": settings.Visibility,
```
sureshperiyappan: visibility is stored as the raw pointer (settings.Visibility) while display_type on line 228 is nil-checked and dereferenced. Both are *string fields from the SDK. If there's a reason visibility is always guaranteed non-nil that's fine, but it'd be safer to handle them the same way — either both as pointers or both with nil-check + dereference.

sureshperiyappan:
Missing test coverage for the new fields. The existing acceptance test only asserts on launcher_button.0.visibility. Could you add TestCheckResourceAttr assertions for display_type, icon.0.url, and a localized_labels entry with key MessengerLauncherButtonText? Even extending the existing test config would be enough — just want to make sure these fields survive a round-trip through create/read.

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
Ajith-26: Made consistent with adding nil-checks + dereference on both fields (visibility and display_type)

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
Ajith-26: Sure, updated the value field description

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
Ajith-26: Earlier, it was wrongly set icon/url field as "Required", but now made it "Optional", also the api accepts/returns in below formats:
a. "icon":{}
b. "icon":{"url":""}
c. "icon":{"url":"https://www.example.com/image1"}
Hence altered build and flatten functions accordingly

Ajith-26:
Done testing in local with new fields.
Able to create configuration with new fields (displayType, launcher button icon url, launcher button text)
image

sureshperiyappan (approved):

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_messenger.go
```diff
@@ -132,11 +132,26 @@ func buildMessengerSettings(d *schema.ResourceData) *platformclientv2.Messengers
 
 	if launchers, ok := cfg["launcher_button"].([]interface{}); ok && len(launchers) > 0 && launchers[0] != nil {
 		launcher := launchers[0].(map[string]interface{})
+		launcherSettings := &platformclientv2.Launcherbuttonsettings{}
+
 		if visibility, ok := launcher["visibility"].(string); ok {
```
sanjaysg21: In buildMessengerSettings, display_type has an empty-string guard but visibility doesn't. Can we add && visibility != "" for consistency? Without it, an unset visibility could send an empty string to the API.

Ajith-26: Correct, added empty-string guard for visibility field as well

sanjaysg21 (approved):

PR #2374, opened 2026-06-15
RULES-1399 Update CX as code to use bulk operations

Jx-hou:
Replace the per-row decision table row add/update/delete with v191 bulk APIs to improve performance.


genesyscloud/business_rules_decision_table/decision_table_bulk_http.go
```diff
@@ -0,0 +1,74 @@
...
+	maxBulkDecisionTableRowsAdd    = 15
+	maxBulkDecisionTableRowsUpdate = 15
+	maxBulkDecisionTableRowsRemove = 49
+)
+
+func getBulkChunkLimits() (add, update, remove int) {
+	return maxBulkDecisionTableRowsAdd, maxBulkDecisionTableRowsUpdate, maxBulkDecisionTableRowsRemove
+}
+
+func bulkAddDecisionTableRowsFn(ctx context.Context, p *BusinessRulesDecisionTableProxy, tableId string, version int, rows []platformclientv2.Createdecisiontablerowrequest) (*platformclientv2.APIResponse, error) {
```
brlenn, line 23: These functions could be added to the proxy file and replace the deleted functions

PR #2375, opened 2026-06-15
NO-JIRA: BCP Exporter improvements

bbbco:
Improvements to the BCP Exporter tool fixes some proxy handling and concurrency related issues.

### DIFF
--- a/genesyscloud/bcp_tf_exporter/genesyscloud_bcp_tf_exporter_proxy.go
+++ b/genesyscloud/bcp_tf_exporter/genesyscloud_bcp_tf_exporter_proxy.go
@@ -11,41 +11,38 @@ import (
 	"github.com/mypurecloud/platform-client-sdk-go/v191/platformclientv2"
 )
 
-type BcpExporterProxy struct {
-	ClientConfig               *platformclientv2.Configuration
-	GetFlowDependenciesAttr    getFlowDependenciesFunc
-	GetAllWithPooledClientAttr getPooledClientFunc
+type bcpExporterProxy struct {
+	clientConfig               *platformclientv2.Configuration
+	getFlowDependenciesAttr    getFlowDependenciesFunc
+	getAllWithPooledClientAttr getPooledClientFunc
 }
 
-type getFlowDependenciesFunc func(ctx context.Context, p *BcpExporterProxy, resourceInfo resourceExporter.ResourceInfo) (resourceExporter.ResourceIDMetaMap, *resourceExporter.DependencyResource, error)
+type getFlowDependenciesFunc func(ctx context.Context, p *bcpExporterProxy, resourceInfo resourceExporter.ResourceInfo) (resourceExporter.ResourceIDMetaMap, *resourceExporter.DependencyResource, error)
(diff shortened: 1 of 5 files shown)


genesyscloud/bcp_tf_exporter/resource_genesyscloud_bcp_tf_exporter.go
```diff
@@ -104,47 +116,76 @@ func createBcpTfExporter(ctx context.Context, d *schema.ResourceData, meta inter
...
-				tflog.Warn(ctx, "Resource type not registered, skipping", map[string]interface{}{
-					"resource_type": resourceType,
-					"id":            id,
-				})
-				continue
+			items = append(items, workItem{id: id, resMeta: resMeta})
+		}
+
+		// process items with bounded concurrency
+		workerCount := min(maxWorkers, len(items), 1)
```
sureshperiyappan: This doesn't look right — min(maxWorkers, len(items), 1) will always return 1 since the built-in min picks the smallest value among all arguments, and 1 will always be ≤ the other two. The concurrency introduced in this PR won't actually kick in.

Did you mean something like:

workerCount := max(1, min(maxWorkers, len(items)))
That gives you the smaller of maxWorkers and len(items), with a floor of 1.

genesyscloud/bcp_tf_exporter/genesyscloud_bcp_tf_exporter_proxy.go
```diff
@@ -11,41 +11,38 @@ import (
...
-func GetBcpExporterProxy(ClientConfig *platformclientv2.Configuration) *BcpExporterProxy {
-	return newBcpExporterProxy(ClientConfig)
-}
-
-func newBcpExporterProxy(ClientConfig *platformclientv2.Configuration) *BcpExporterProxy {
-	if InternalProxy == nil {
-		InternalProxy = &BcpExporterProxy{
-			GetAllWithPooledClientAttr: getPooledClientFn,
-		}
+func getBcpExporterProxy(ClientConfig *platformclientv2.Configuration) *bcpExporterProxy {
```
sureshperiyappan, line 25: Minor: once the concurrency is actually working (after the min fix), multiple goroutines processing flow resources will hit getBcpExporterProxy simultaneously. The check-then-write on internalProxy is a data race that'll trip go test -race. It won't cause wrong behavior since the proxy is stateless, but it's easy to clean up — you could just drop the singleton and return newBcpExporterProxy(clientConfig) directly, which is what the group package does.

genesyscloud/bcp_tf_exporter/resource_genesyscloud_bcp_tf_exporter.go
bbbco: Yep! Meant to do `max(min(maxWorkers, len(items)), 1)`. Nice catch.

genesyscloud/bcp_tf_exporter/genesyscloud_bcp_tf_exporter_proxy.go
bbbco: I'll refactor the handling so flows are processed sequentially so it won't hit this issue.

sureshperiyappan (approved):

PR #2377, opened 2026-06-17
fixed backslashes not accepted in windows

sanjaysg21:


genesyscloud/responsemanagement_responseasset/genesyscloud_responsemanagement_responseasset_proxy.go
```diff
@@ -124,6 +124,14 @@ func (p *responsemanagementResponseassetProxy) uploadRespManagementRespAsset(ctx
 		name = localFilePath
 	}
 
+	// Normalize the asset name for cross-platform compatibility.
```
sureshperiyappan, line 127: This normalization logic is duplicated in three places (here, and both branches of updateRespManagementRespAsset). Consider extracting a shared helper to reduce maintenance risk:

func normalizeAssetName(name string) string {
    name = strings.ReplaceAll(name, "\\", "/")
    if len(name) >= 3 && name[1] == ':' && name[2] == '/' {
        name = name[3:]
    }
    return name
}
Then each call site becomes just name = normalizeAssetName(name). This also makes it straightforward to add unit tests and handle future edge cases (e.g., UNC paths, empty result after stripping) in one place.

genesyscloud/responsemanagement_responseasset/resource_genesyscloud_responsemanagement_responseasset.go
```diff
@@ -117,9 +118,20 @@ func updateRespManagementRespAsset(ctx context.Context, d *schema.ResourceData,
 	divisionId := d.Get("division_id").(string)
 
 	var bodyRequest platformclientv2.Responseassetrequest
+
+	// Normalize the asset name for cross-platform compatibility.
+	// Convert backslashes to forward slashes (Windows paths use \ which the API rejects).
 	if name != "" {
```
sureshperiyappan, line 120: The normalization logic lives in two different layers now — here in the resource layer for updates, and in the proxy for creates (uploadRespManagementRespAsset). That split makes it easy to miss one if a new code path gets added later.

Since the proxy is the last stop before the SDK call, would it make more sense to consolidate all normalization there? You could add it to the updateRespManagementRespAsset proxy method (normalize respAsset.Name before calling the SDK), and then this resource function can just pass the raw values through — same as createRespManagementRespAsset already does.

That way there's one place to maintain it and no caller can accidentally send an un-normalized name to the API.

genesyscloud/responsemanagement_responseasset/resource_genesyscloud_responsemanagement_responseasset_schema.go
```diff
@@ -85,6 +85,11 @@ func ResourceResponseManagementResponseAsset() *schema.Resource {
 				Optional:    true,
 				Computed:    true,
 				ForceNew:    true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
sanjaysg21, line 88: Added DiffSuppressFunc on the name attribute because normalizing the name before sending to the API causes the value stored in state (e.g., "terraform-test/test.png") to differ from the user's config (e.g., "terraform-test\\test.png"). Without it, the consistency checker detects a mismatch between the original plan value and the API-returned value, retries for 5 minutes, and fails. The DiffSuppressFunc tells both Terraform and the consistency checker (which uses SimpleDiff internally) that these path variants are semantically equivalent.

genesyscloud/responsemanagement_responseasset/genesyscloud_responsemanagement_responseasset_proxy.go
sanjaysg21: declared the function in utils file and have used it here. also since proxy the last end point before sdk removed the backslash conversions in resource layer.

genesyscloud/responsemanagement_responseasset/resource_genesyscloud_responsemanagement_responseasset_utils.go
```diff
@@ -14,6 +14,17 @@ import (
 	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util/files"
 )
 
+// normalizeAssetName converts Windows-style path separators to forward slashes
+// and strips drive letter prefixes (e.g., "C:/") to ensure cross-platform
+// compatibility with the Genesys Cloud API, which rejects backslashes and colons.
+func normalizeAssetName(name string) string {
```
bbbco, line 20: I think we have to do this in several other resource types too (contact lists, flows, user prompts, etc). Can we ensure we are handling path normalization the same across the board?

sanjaysg21: Sure ill do it

sureshperiyappan (approved):

PR #2380, opened 2026-06-18
DEVTOOLING-1619 - Fixed genesyscloud_outbound_digitalruleset export issues 

daragh-kearney:
- Export now results in references instead of GUIDs
- Export no longer leaves raw GUIDs for missing/deleted refs like contact_list_id
- SMS and email template IDs are optional in the schema (at least one required via validation)
- Added unit tests and updated docs
- Manually tested export with a digital ruleset — references look correct and unresolved contact_list_id is omitted


genesyscloud/resource_exporter/resource_exporter_custom.go
```diff
@@ -104,6 +104,32 @@ func isValidGuid(id string) bool {
 	return matched
 }
 
+// OmitUnresolvedGuidAttributeResolver removes attribute values that could not be resolved to a Terraform reference.
+// Optional reference attributes that still contain a raw GUID after export resolution are omitted from the config.
```
bbbco, line 116: I'm not sure whether this is desired blanket behavior, as it removes configuration that likely exists for a reason. I understand why you would want this, but I think we need to confirm this is desired behavior before making it default.

Can you add either a feature flag or an attribute on the exporter to toggle this behavior (where the default is the current behavior and gives plenty of time and opportunity to test this behavior out before toggling the default in a future release?). See the `export_computed` and `export_deprecated` attributes on the exporter schema.

daragh-kearney: I've updated this so omitting unresolved GUIDs is opt-in rather than default.

Added export_omit_unresolved_refs on genesyscloud_tf_export (defaults to false). Resources that want this mark attributes with OmitUnresolvedRef on their exporter config; the omit logic only runs when that flag is true.

`genesyscloud_outbound_digitalruleset` uses this for contact_list_id and the content template IDs.

sureshperiyappan (approved):

PR #2382, opened 2026-06-22
added retryable error login for building dependency

sanjaysg21:


genesyscloud/knowledge_category/resource_genesyscloud_knowledge_category.go
```diff
@@ -150,21 +150,25 @@ func deleteKnowledgeCategory(ctx context.Context, d *schema.ResourceData, meta i
...
-		if strings.Contains(err.Error(), "in use by Bot flow status unknown") {
-			rebuildErr := rebuildDatabase()
-			if rebuildErr != nil {
-				log.Printf("Failed to rebuild knowledge category database: %v", rebuildErr)
-			}
-			time.Sleep(10 * time.Second)
-			_, resp, err = proxy.deleteKnowledgeCategory(ctx, knowledgeBaseId, knowledgeCategoryId)
-			if err != nil {
-				return util.BuildAPIDiagnosticError("genesyscloud_knowledge_category", fmt.Sprintf("Failed to delete knowledge category %s error: %s", id, err), resp)
+
```
sanjaysg21, line 153: Knowledge base resources (knowledgebase, category, label, document, document_variation)
fail to delete when the Architect dependency tracking system has stale "Bot flow status
unknown" references.

The previous logic attempted one dependency tracking rebuild followed by a single delete
retry, which was insufficient when the stale reference took longer to clear.

Replaced with a retry loop (2-minute timeout) that repeatedly rebuilds dependency tracking
and reattempts deletion until the stale reference is purged, rather than giving up after
one attempt.

genesyscloud/knowledge_category/resource_genesyscloud_knowledge_category.go
sureshperiyappan, line 189: rebuildDatabase() is copy-pasted across 5 packages (knowledge_category, knowledge_document, knowledge_document_variation, knowledge_knowledgebase, knowledge_label) and they have subtle differences in error handling — for example, the knowledgebase version treats GetArchitectDependencytrackingBuild errors as fatal and returns diag.Errorf, while the others just log and break. Since this PR touches all 5 instances and they're doing the exact same thing (triggering and polling the architect dependency tracking build), would be worth extracting this into a shared utility — maybe something like a knowledge_util package or even adding it to the existing dependency tracking proxy. Would reduce the maintenance surface and make the behavior consistent. Fine as a follow-up if you don't want to bloat this PR

genesyscloud/knowledge_category/resource_genesyscloud_knowledge_category.go
```diff
@@ -150,21 +150,25 @@ func deleteKnowledgeCategory(ctx context.Context, d *schema.ResourceData, meta i
...
+	// Retry the delete operation to handle stale "Bot flow status unknown" dependency references.
+	diagErr := util.WithRetries(ctx, 2*time.Minute, func() *retry.RetryError {
```
sureshperiyappan: The 2-minute timeout here might be too tight. rebuildDatabase() polls the dependency tracking status in a loop with 10-second sleeps, so a single rebuild cycle can easily take 30-60+ seconds. If the first delete attempt triggers a rebuild and it takes a while, there's not much budget left for a second attempt. Other places in the repo use 5 minutes for similar patterns (e.g. outbound campaign stop, auth division delete). I'd suggest bumping this to 5*time.Minute to give comfortable room for 1-2 full rebuild cycles.

genesyscloud/knowledge_document/resource_genesyscloud_knowledge_document.go
```diff
@@ -200,25 +200,29 @@ func deleteKnowledgeDocument(ctx context.Context, d *schema.ResourceData, meta i
...
-			if rebuildErr != nil {
-				log.Printf("Failed to rebuild knowledge document database: %v", rebuildErr)
-			}
-			time.Sleep(10 * time.Second)
-			resp, err = proxy.deleteKnowledgeKnowledgebaseDocument(ctx, knowledgeBaseId, knowledgeDocumentId)
-			if err != nil {
-				return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to delete knowledge document %s error: %s", knowledgeDocumentId, err), resp)
+
+	// Retry the delete operation to handle stale "Bot flow status unknown" dependency references.
+	diagErr := util.WithRetries(ctx, 2*time.Minute, func() *retry.RetryError {
```
sureshperiyappan: Same as above — 2 minutes feels tight given rebuildDatabase() blocks for 30-60+ seconds per iteration. Would bump to 5*time.Minute to stay consistent with what we do elsewhere in the codebase.

genesyscloud/knowledge_document_variation/resource_genesyscloud_knowledge_document_variation.go
```diff
@@ -283,21 +283,24 @@ func deleteKnowledgeDocumentVariation(ctx context.Context, d *schema.ResourceDat
...
-			rebuildErr := rebuildDatabase()
-			if rebuildErr != nil {
-				log.Printf("Failed to rebuild knowledge base database: %v", rebuildErr)
-			}
-			time.Sleep(10 * time.Second)
-			resp, err = variationProxy.deleteVariationRequest(ctx, ids.knowledgeDocumentVariationID, ids.knowledgeDocumentID, ids.knowledgeBaseID)
-			if err != nil {
-				return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to delete knowledge document variation %s error: %s", ids.knowledgeDocumentVariationID, err), resp)
+	// Retry the delete operation to handle stale "Bot flow status unknown" dependency references.
+	diagErr := util.WithRetries(ctx, 2*time.Minute, func() *retry.RetryError {
```
sureshperiyappan: Same here — consider 5*time.Minute instead of 2 to account for the rebuild time inside the loop.

genesyscloud/knowledge_knowledgebase/resource_genesyscloud_knowledge_knowledgebase.go
```diff
@@ -181,21 +181,26 @@ func deleteKnowledgeKnowledgebase(ctx context.Context, d *schema.ResourceData, m
...
-				log.Printf("Failed to rebuild knowledge base database: %v", rebuildErr)
-			}
-			time.Sleep(10 * time.Second)
-			_, resp, err = knowledgebaseProxy.deleteKnowledgebase(ctx, d.Id())
-			if err != nil {
-				return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to delete knowledge base %s error: %s", name, err), resp)
+
+	// Retry the delete operation to handle stale "Bot flow status unknown" dependency references.
+	// The dependency tracking system may need multiple rebuild cycles to clear ghost entries.
+	diagErr := util.WithRetries(ctx, 2*time.Minute, func() *retry.RetryError {
```
sureshperiyappan: Same feedback — 5*time.Minute would be safer. The comment even says "may need multiple rebuild cycles" which reinforces that 2 minutes is likely not enough.

genesyscloud/knowledge_label/resource_genesyscloud_knowledge_label.go
```diff
@@ -154,21 +154,25 @@ func deleteKnowledgeLabel(ctx context.Context, d *schema.ResourceData, meta inte
...
-			if rebuildErr != nil {
-				log.Printf("Failed to rebuild knowledge base database: %v", rebuildErr)
-			}
-			time.Sleep(10 * time.Second)
-			_, resp, err = proxy.deleteKnowledgeLabel(ctx, knowledgeBaseId, knowledgeLabelId)
-			if err != nil {
-				return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to delete knowledge label %s error: %s", id, err), resp)
+
+	// Retry the delete operation to handle stale "Bot flow status unknown" dependency references.
+	diagErr := util.WithRetries(ctx, 2*time.Minute, func() *retry.RetryError {
```
sureshperiyappan: Same — bump to 5*time.Minute.

genesyscloud/knowledge_category/resource_genesyscloud_knowledge_category.go
```diff
@@ -150,21 +150,25 @@ func deleteKnowledgeCategory(ctx context.Context, d *schema.ResourceData, meta i
...
+		_, resp, err := proxy.deleteKnowledgeCategory(ctx, knowledgeBaseId, knowledgeCategoryId)
+		if err != nil {
+			if strings.Contains(err.Error(), "in use by Bot flow status unknown") {
+				log.Printf("Knowledge category %s in use by Bot flow with status unknown. Rebuilding dependency tracking and retrying.", knowledgeCategoryId)
+				rebuildErr := rebuildDatabase()
+				if rebuildErr != nil {
```
sureshperiyappan, line 161: If rebuildDatabase() fails with a terminal error (like 403 or a network failure), we still return RetryableError and keep hammering the delete for the full timeout. Might be worth distinguishing — if the rebuild itself can't succeed, retrying the delete won't help either. Not blocking but would give users a clearer error faster. Could be addressed in the follow-up when this gets extracted to a shared utility.

genesyscloud/knowledge_category/resource_genesyscloud_knowledge_category.go
sanjaysg21: The repo follows a pattern where each resource package is self-contained ,there's no existing shared utility between knowledge resource packages for domain-specific logic like this. The genesyscloud/util package is only for generic helpers (retries, diagnostics, etc.), not for Architect API calls. Introducing a new shared package just for rebuildDatabase() would break the established convention. The duplication and inconsistency is pre-existing and wasn't introduced by this PR, so I'd prefer to keep this out of scope in this PR

genesyscloud/knowledge_category/resource_genesyscloud_knowledge_category.go
sanjaysg21: In practice, even if the rebuild call fails, the stale reference can still clear on its own over time since the platform's dependency tracker runs periodically ,so retrying the delete still has a chance of succeeding. The 5-minute timeout keeps it bounded either way.

genesyscloud/knowledge_category/resource_genesyscloud_knowledge_category.go
sanjaysg21: agreed, added this

sureshperiyappan (approved):

sureshperiyappan (approved):

PR #2383, opened 2026-06-22
Ep 370 routing email domain graph imap settings

SvetaB11:
**Fixes [EP-370]**(https://inindca.atlassian.net/browse/EP-370): genesyscloud_routing_email_domain could not read, export, or apply Microsoft Graph / IMAP connector settings. Domains promoted via CX as Code were created as plain Custom domains because graphApiSettings was omitted from Terraform state and PATCH payloads.

This PR adds optional nested blocks that map to the Platform API `graphApiSettings` and `imapSettings` objects to: `graph_api_settings` and `imap_settings` 
Writable surface is integration_id only (same idea as custom_smtp_server_id). status is computed from the API.

**Changes**
Schema: graph_api_settings and imap_settings blocks
Read: Flatten GraphApiSettings / ImapSettings from GET into state (fixes export)
Create: Deferred PATCH when Graph/IMAP blocks are set (same pattern as mail_from_domain / custom_smtp_server_id)
Update: Include GraphApiSettings / ImapSettings on Inbounddomainpatchrequest when configured
Docs: Regenerated docs/resources/routing_email_domain.md
Tests: Unit tests for flattenGraphApiSettings and flattenImapSettings

[EP-370]: https://inindca.atlassian.net/browse/EP-370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_schema.go
```diff
@@ -75,8 +119,14 @@ func DataSourceRoutingEmailDomain() *schema.Resource {
 func RoutingEmailDomainExporter() *resourceExporter.ResourceExporter {
 	return &resourceExporter.ResourceExporter{
 		GetResourcesFunc: provider.GetAllWithPooledClient(getAllRoutingEmailDomains),
+		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
```
sanjaysg21: graph_api_settings and imap_settings are listed in both RefAttrs and UnResolvableAttributes, but these two mechanisms conflict. During export, the exporter first recurses into the block and resolves integration_id via RefAttrs, but then UnResolvableAttributes matches the parent key (graph_api_settings) and overwrites the entire block with a ${var.…} placeholder ,discarding the resolved reference. This makes the RefAttrs entries effectively dead code that never produces output. Since the PR description says this follows the same pattern as custom_smtp_server_id (which is only in UnResolvableAttributes), I'd suggest removing the RefAttrs entries for graph_api_settings.integration_id and imap_settings.integration_id to keep the behavior consistent and avoid confusion.

genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_unit_test.go
```diff
@@ -75,3 +75,59 @@ func TestUnitGetRoutingEmailDomainIdByName_SubdomainPrefixMatch(t *testing.T) {
 		t.Fatalf("expected id %q, got %q", fullID, gotID)
 	}
 }
+
```
sanjaysg21, line 83: Could we get matching tests for expandGraphApiSettings and expandImapSettings as well? They have several early-return nil paths (block absent, empty list, empty integration_id) that would be good to pin down. A table-driven test with cases for nil/empty/valid input would round out the coverage and guard against regressions if the schema or SDK types change later.

docs/resources/routing_email_domain.md
```diff
@@ -53,10 +53,35 @@ resource "genesyscloud_routing_email_domain" "example_domain_com" {
 ### Optional
 
 - `custom_smtp_server_id` (String) The ID of the custom SMTP server integration to use when sending outbound emails from this domain.
+- `graph_api_settings` (Block List, Max: 1) The Graph API server integration and settings to use for processing inbound and outbound emails. (see [below for nested schema](#nestedblock--graph_api_settings))
```
sureshperiyappan, line 56: are the doc changes are manual update or by running make docs command?

SvetaB11: Generated by make docs, not manual.

genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_unit_test.go
SvetaB11: I added matching tests for expandGraphApiSettings and expandImapSettings.

genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_schema.go
SvetaB11: Fixed

sanjaysg21 (approved):

sureshperiyappan (approved):

PR #2384, opened 2026-06-23
[MRMO-357] Improve export performance by eliminating redundant API calls + improved caching

charliecon:
During CX as Code export, the provider repeatedly fetched the same related data (queue members, wrapup codes, team members, group members, user policies, etc.) because caches were either missing or scoped to individual proxy instances — so concurrent export workers never shared hits. 

**Resources updated:**

| Resource | What's cached | Invalidation on mutation |
|---|---|---|
| `genesyscloud_routing_queue` | Queue list, per-queue wrapup codes, per-queue members | Yes |
| `genesyscloud_team` | Team list, team by ID, team members | Yes |
| `genesyscloud_group` | Group members, voicemail group policy | Yes |
| `genesyscloud_user` | Voicemail user policy, routing utilization (raw response) | Yes |
| `genesyscloud_routing_skill_group` | **Not cached** — list vs. get-by-ID return incompatible SDK types | N/A |

**Notable queue-specific optimizations:**
- Skip member API calls when `UserMemberCount == 0` (uses count already on the queue object).
- Wrapup code fetch page size increased to 500.
- Removed broken per-proxy `wrapupCodeCache`; replaced with shared `queueWrapupCodesCache`.
- Extracted `setRoutingQueueStateFromQueue` for reuse between read and partial-failure sync paths.

---

### Performance impact

Benchmarks were run against inintesting, which held about 2500 queues and the maximum of 200 teams.

#### Queue export (`genesyscloud_routing_queue`)

queue_export_duration

| Branch | Avg duration | Member GET requests |
|---|---|---|
| v1.83.0 | 113.3s | 2,594 |
| dev | 107.5s | 2,582 |
| **MRMO-357** | **54.8s** | **152** |

- **~2.07× faster** wall-clock vs v1.83.0
- **94.1% fewer** `GET /routing/queues/{id}/members` calls during export

queue_member_api_calls

#### Team export (`genesyscloud_team`)

teams_export_duration

| Branch | Avg duration | Member GET requests |
|---|---|---|
| v1.83.0 | 11.7s | 194 |
| **MRMO-357** | **5.8s** | **36** |

- **~2.02× faster** vs v1.83.0
- **81.4% fewer** `GET /teams/{id}/members` calls

teams_member_api_calls

### Package-level proxy caches

Moved ResourceCache maps from per-instance allocation in `new*Proxy` to package-level vars across 48 proxy files. Previously, concurrent pagination and any code path calling `new*Proxy` instead of the singleton got a fresh empty cache, so export workers never shared hits. Package-level caches make caching correct regardless of which constructor is used, matching the existing pattern in routing_skill, team, group, and routing_queue. Also fixed cache invalidation in user (extension pools) and telephony_providers_edges_site (managed vs unmanaged site lists) to clear the shared package var, not just a local or struct field reference.


genesyscloud/routing_wrapupcode/genesyscloud_routing_wrapupcode_proxy.go
```diff
@@ -162,7 +180,7 @@ func getAllRoutingWrapupcodeFn(ctx context.Context, p *routingWrapupcodeProxy) (
 	ctx = provider.EnsureResourceContext(ctx, ResourceType)
 
 	var allWrapupcodes []platformclientv2.Wrapupcode
-	const pageSize = 100
+	const pageSize = 500
```
harter-genesys: I think I mentioned this in another pr, but having this configurable would be nice

harter-genesys (approved):

genesyscloud/routing_wrapupcode/genesyscloud_routing_wrapupcode_proxy.go
charliecon: I added a new package inside util that allows us to configure this with env vars 

e.g.

```bash
export GENESYSCLOUD_PAGE_SIZE_genesyscloud_group="500"
```

To keep the PR size from exploding, I only implemented this for each resource we cranked up to 500. If we're going to enable this for all resources, I think we should work with the CX as Code team and see about extending `genesyscloud_tf_export` to support per-resource configuration, as opposed to going the env var route.

sureshperiyappan (approved):

PR #2386, opened 2026-06-23
Bug/devtooling 1413  phone cannot be created for a user error

shrutisuryawanshigenesys:


genesyscloud/telephony_providers_edges_phone/genesyscloud_telephony_providers_edges_phone_proxy.go
```diff
@@ -61,8 +62,9 @@ type phoneProxy struct {
 	getAllPhonesAttr   getAllPhonesFunc
 	createPhoneAttr    createPhoneFunc
 	getPhoneByIdAttr   getPhoneByIdFunc
-	getPhoneByNameAttr getPhoneByNameFunc
-	updatePhoneAttr    updatePhoneFunc
+	getPhoneByNameAttr         getPhoneByNameFunc
```
sureshperiyappan, line 65: run go fmt

genesyscloud/telephony_providers_edges_phone/genesyscloud_telephony_providers_edges_phone_proxy.go
```diff
@@ -287,6 +295,38 @@ func getPhoneByNameFn(ctx context.Context, p *phoneProxy, phoneName string) (pho
...
+	if phones.Entities == nil || len(*phones.Entities) == 0 {
+		return nil, resp, nil
+	}
+
+	var deletedPhone *platformclientv2.Phone
+	for _, phone := range *phones.Entities {
+		if phone.WebRtcUser == nil || phone.WebRtcUser.Id == nil || *phone.WebRtcUser.Id != webRtcUserId {
+			continue
+		}
+		if phone.State != nil && *phone.State == "deleted" {
```
sureshperiyappan, line 318: getPhoneByWebRtcUserIdFn falls back to returning a deleted phone (line 327) when there's no active one found. Downstream in adoptExistingWebRtcPhone (utils line 102), we only check for nil — so if the API returns a phone in "deleted" state, we'll happily set the resource ID to it and call updatePhone on a dead resource. That's going to blow up with a 404 or 400, and the user gets a confusing error instead of the clean create they're expecting.

This is a pretty real scenario too — if someone deletes the phone from the UI and re-runs terraform apply, they'd hit exactly this path.

I think the simplest fix is either:

Return nil instead of deletedPhone in the proxy (line 327), or
Add a state check in adoptExistingWebRtcPhone before adopting:
if existingPhone.State != nil && *existingPhone.State == "deleted" {
    return false, nil
}
Either way it's a one-liner. Without this, the fix works for the active-phone case but still breaks for the deleted-phone case, which is arguably the more common scenario when people are re-running applies after manual cleanup.

genesyscloud/telephony_providers_edges_phone/resource_genesyscloud_telephony_providers_edges_phone_utils.go
```diff
@@ -87,6 +87,27 @@ func getPhoneFromResourceData(ctx context.Context, pp *phoneProxy, d *schema.Res
...
+	}
+	return strings.Contains(strings.ToLower(err.Error()), "web rtc phone has already been assigned")
+}
+
+func adoptExistingWebRtcPhone(ctx context.Context, d *schema.ResourceData, meta interface{}, pp *phoneProxy, webRtcUserId string) (bool, diag.Diagnostics) {
+	existingPhone, resp, err := pp.getPhoneByWebRtcUserId(ctx, webRtcUserId)
+	if err != nil {
+		return false, util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to lookup existing WebRTC phone for user %s error: %s", webRtcUserId, err), resp)
+	}
+	if existingPhone == nil || existingPhone.Id == nil {
```
sureshperiyappan, line 102: getPhoneByWebRtcUserIdFn falls back to returning a deleted phone (line 327) when there's no active one found. Downstream in adoptExistingWebRtcPhone (utils line 102), we only check for nil — so if the API returns a phone in "deleted" state, we'll happily set the resource ID to it and call updatePhone on a dead resource. That's going to blow up with a 404 or 400, and the user gets a confusing error instead of the clean create they're expecting.

This is a pretty real scenario too — if someone deletes the phone from the UI and re-runs terraform apply, they'd hit exactly this path.

I think the simplest fix is either:

Return nil instead of deletedPhone in the proxy (line 327), or
Add a state check in adoptExistingWebRtcPhone before adopting:
if existingPhone.State != nil && *existingPhone.State == "deleted" {
    return false, nil
}
Either way it's a one-liner. Without this, the fix works for the active-phone case but still breaks for the deleted-phone case, which is arguably the more common scenario when people are re-running applies after manual cleanup.

sureshperiyappan (approved):

PR #2388, opened 2026-06-24
Fix/devtooling 1561

sanjaysg21:


genesyscloud/architect_grammar_language/resource_genesyscloud_architect_grammar_language_utils.go
```diff
@@ -17,14 +17,25 @@ import (
...
-	"github.com/mypurecloud/platform-client-sdk-go/v192/platformclientv2"
+	"github.com/mypurecloud/platform-client-sdk-go/v191/platformclientv2"
 )
 
 /*
 The resource_genesyscloud_architect_grammar_language_utils.go file contains various helper methods to marshal
 and unmarshal data into formats consumable by Terraform and/or Genesys Cloud.
 */
 
+// normalizeAssetPath converts Windows-style path separators to forward slashes
```
sanjaysg21, line 28: added a normalizeAssetName helper in the proxy layer to convert \ to / and strip drive letter prefixes before sending the asset name to the API. Other file-uploading resources (genesyscloud_flow, genesyscloud_script, genesyscloud_architect_user_prompt, etc.) are not affected because they only use file paths locally to read content and never send the path itself to the API as a name or identifier.

sureshperiyappan (approved):

PR #2390, opened 2026-06-24
DEVTOOLING-1698 - Fixed issues with genesyscloud_task_management_worktype_status and status_transition: auto_terminate_workitem error and 400 "No change" error on apply

daragh-kearney:
- status_transition: Clearing default_destination_status_id, status_transition_delay_seconds, or status_transition_time no longer fails with “No change for the record is obtained.” Create/update now use the same custom PATCH path as delete (Workitemstatusupdate + customapi), with a MarshalJSON fix so cleared fields are sent as null.

- worktype_status: auto_terminate_workitem = false now works. SDK omitempty was dropping explicit false; we send it via a targeted custom PATCH when the field changes, and tolerate SDK “No change” when that’s the only update.


genesyscloud/task_management_worktype_status/genesyscloud_task_management_worktype_status_proxy.go
```diff
@@ -155,6 +164,19 @@ func updateTaskManagementWorktypeStatusFn(ctx context.Context, p *taskManagement
...
 }
 
+type workitemStatusAutoTerminatePatch struct {
+	AutoTerminateWorkitem bool `json:"autoTerminateWorkitem"`
+}
+
+// patchTaskManagementWorktypeStatusAutoTerminateFn uses customapi so explicit false is sent for auto_terminate_workitem.
+func patchTaskManagementWorktypeStatusAutoTerminateFn(ctx context.Context, p *taskManagementWorktypeStatusProxy, worktypeId string, statusId string, autoTerminate bool) (*platformclientv2.Workitemstatus, *platformclientv2.APIResponse, error) {
+	ctx = provider.EnsureResourceContext(ctx, ResourceType)
+	c := customapi.NewClient(p.taskManagementApi.Configuration, ResourceType)
```
sureshperiyappan: The custom_api_client package docs say to store the client as a field on the proxy struct and initialize it in the constructor. Creating a new client on every call works functionally but doesn't follow the repo convention — and it makes this function harder to mock in unit tests (you can't swap the callAPIAttr without access to the client instance).

Consider adding a customApiClient *customapi.Client field to the taskManagementWorktypeStatusProxy struct (around line 35) and initializing it in newTaskManagementWorktypeStatusProxy. Then reference p.customApiClient here instead.

genesyscloud/task_management_worktype_status/genesyscloud_task_management_worktype_status_proxy.go
```diff
@@ -155,6 +164,19 @@ func updateTaskManagementWorktypeStatusFn(ctx context.Context, p *taskManagement
 	return p.taskManagementApi.PatchTaskmanagementWorktypeStatus(worktypeId, statusId, *taskManagementWorktypeStatus)
 }
 
+type workitemStatusAutoTerminatePatch struct {
+	AutoTerminateWorkitem bool `json:"autoTerminateWorkitem"`
+}
+
+// patchTaskManagementWorktypeStatusAutoTerminateFn uses customapi so explicit false is sent for auto_terminate_workitem.
+func patchTaskManagementWorktypeStatusAutoTerminateFn(ctx context.Context, p *taskManagementWorktypeStatusProxy, worktypeId string, statusId string, autoTerminate bool) (*platformclientv2.Workitemstatus, *platformclientv2.APIResponse, error) {
+	ctx = provider.EnsureResourceContext(ctx, ResourceType)
```
sureshperiyappan: This call is redundant — customapi.Do already calls provider.EnsureResourceContext(ctx, c.resourceType) internally (see custom_api_client.go line ~150). Double-calling it doesn't break anything but is misleading for future readers who might think callers are expected to do this. I'd remove it to keep things consistent with how the custom_api_client package is designed to be consumed.

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status.go
```diff
@@ -124,6 +123,13 @@ func createTaskManagementWorktypeStatus(ctx context.Context, d *schema.ResourceD
 
 	d.SetId(worktypeId + "/" + *workitemStatus.Id)
 
+	if autoTerminate := resourcedata.GetNillableBool(d, "auto_terminate_workitem"); autoTerminate != nil {
```
sureshperiyappan, line 126: Every other API call in this resource's create function is wrapped in util.WithRetries handling 400 "Database transaction was cancelled" and 409 as retryable. This patch call isn't. The same race condition applies here — if two statuses are being created simultaneously on the same worktype, the auto-terminate patch could hit the same transient DB error.

Should this be wrapped in the same retry pattern for consistency? Something like:


diagErr = util.WithRetries(ctx, 60*time.Second, func() *retry.RetryError {
    workitemStatus, resp, err = proxy.patchTaskManagementWorktypeStatusAutoTerminate(...)
    if err == nil {
        return nil
    }
    builtDiagErr := util.BuildWithRetriesApiDiagnosticError(...)
    if util.IsStatus400(resp) && strings.Contains(resp.ErrorMessage, "Database transaction was cancelled") {
        return retry.RetryableError(builtDiagErr)
    }
    if util.IsStatus409(resp) {
        return retry.RetryableError(builtDiagErr)
    }
    return retry.NonRetryableError(builtDiagErr)
})

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status.go
```diff
@@ -239,6 +244,13 @@ func updateTaskManagementWorktypeStatus(ctx context.Context, d *schema.ResourceD
 		if err == nil {
 			return nil
 		}
+		// When only auto_terminate_workitem changed, the SDK update omits false and the API may return no change.
+		if util.IsStatus400(resp) && strings.Contains(resp.ErrorMessage, "No change for the record is obtained") && d.HasChange("auto_terminate_workitem") {
+			workitemStatus, resp, err = proxy.getTaskManagementWorktypeStatusById(ctx, worktypeId, statusId)
```
sureshperiyappan: The resp and err variables get reassigned by the GET call here. If the GET fails, you fall through to the builtDiagErr construction below which now uses the GET's resp and err, not the original PATCH response. This could produce a confusing error message like "Failed to update status: 404 not found" when the actual issue was the PATCH returning "no change".

Suggest using separate variables for the fallback GET:


if util.IsStatus400(resp) && strings.Contains(resp.ErrorMessage, "No change for the record is obtained") && d.HasChange("auto_terminate_workitem") {
    // SDK omits auto_terminate_workitem=false, so API reports no change. Verify current state.
    currentStatus, _, getErr := proxy.getTaskManagementWorktypeStatusById(ctx, worktypeId, statusId)
    if getErr == nil && currentStatus != nil {
        workitemStatus = currentStatus
        return nil
    }
}

genesyscloud/task_management_worktype_status/resource_genesyscloud_task_management_worktype_status_test.go
```diff
@@ -172,3 +172,107 @@ func testVerifyTaskManagementWorktypeStatusDestroyed(state *terraform.State) err
...
+						util.NullValue,
+						"",
+						"default = false",
+						"auto_terminate_workitem = false",
+					),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(ResourceType+"."+statusResourceLabel2, "auto_terminate_workitem", util.FalseValue),
+				),
+			},
+		},
```
sureshperiyappan, line 280: The existing TestAccResourceTaskManagementWorktypeStatus includes an ImportState step to verify round-trip correctness. Can we add one here too? It would catch any issues with auto_terminate_workitem not being read back correctly from the API:


{
    ResourceName:      ResourceType + "." + statusResourceLabel2,
    ImportState:       true,
    ImportStateVerify: true,
},

daragh-kearney: all comments now addressed in latest push. Let me know if anything else needs a look.

sureshperiyappan (approved):

PR #2400, opened 2026-06-30
DEVTOOLING-1617: Add regression test for stable integration action ex…

Teja-genesys:
Title: DEVTOOLING-1617: Add regression test for stable integration action export labels

**Summary:**
Adds a regression test (TestSameNameDifferentCategoryProducesStableDistinctLabels) that verifies the fix for non-deterministic export labels when two integration actions share the same name but belong to different categories.

**Problem:**
In v1.75.1, the exporter used only action.Name as the BlockLabel. When two actions had the same name (e.g., "Log Call") in different categories, the sanitizer would randomly assign which one got the bare label vs a hash suffix — causing destructive plan changes on every export/apply cycle.

**Fix (already shipped):**
The BlockLabel now includes the category: category + "_" + name, ensuring same-named actions in different categories always produce distinct, stable labels.

**What this PR adds:**
A unit test that runs the full buildIntegrationActionBlockLabel → sanitizer pipeline 100 times and asserts:
Labels for same-named actions in different categories are always distinct
Labels are stable (identical across all iterations)

**Testing:**
go test -v -run TestSameNameDifferentCategoryProducesStableDistinctLabels ./genesyscloud/integration_action/

### DIFF
--- a/docs/resources/routing_email_route.md
+++ b/docs/resources/routing_email_route.md
@@ -50,6 +50,12 @@ resource "genesyscloud_routing_email_route" "example_route" {
     name  = "Supervisors"
     email = "support_supervisors@${genesyscloud_routing_email_domain.example_domain_com.domain_id}"
   }
+  signature {
+    enabled            = false
+    canned_response_id = genesyscloud_responsemanagement_response.example_signature_response.id
+    always_included    = false
+    inclusion_type     = "FirstResponseOnly"
+  }
 }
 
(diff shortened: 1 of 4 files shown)


sureshperiyappan (approved):

sureshperiyappan:
address the merge conflict, pull the latest dev.

Teja-genesys:
> address the merge conflict, pull the latest dev.

done, resolved merge conflicts

PR #2404, opened 2026-07-01
DEVTOOLING-1497 team_ids issue fix for incorrect tfstate

shrutisuryawanshigenesys:


genesyscloud/recording_media_retention_policy/resource_genesyscloud_recording_media_retention_policy.go
```diff
@@ -97,14 +97,19 @@ func createMediaRetentionPolicy(ctx context.Context, d *schema.ResourceData, met
 	policy, resp, err := pp.createPolicy(ctx, &reqBody)
 	log.Printf("Media retention policy creation status %#v", resp.Status)
 	if err != nil {
+		d.Partial(true)
```
sureshperiyappan, line 102: d.Partial(true) is a no-op in terraform-plugin-sdk v2.x. It compiles fine but does nothing at runtime. No other resource in this repo uses it. It's not harmful, but wanted to flag it so future readers don't assume it's providing some partial-state protection. The real fix here is the TypeList migration, which is solid.

Up to you whether to keep or remove — just want to make sure it's an intentional choice rather than something inherited from a v1 example.

(Same applies to lines 110, 203, 211)

genesyscloud/recording_media_retention_policy/resource_genesyscloud_recording_media_retention_policy_schema_upgrade.go
```diff
@@ -0,0 +1,132 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func upgradeMediaRetentionPolicyStateV1(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
```
sureshperiyappan, line 70: func upgradeMediaRetentionPolicyStateV1(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {

Minor: Would be nice to add a short doc comment here explaining the why, like:

// upgradeMediaRetentionPolicyStateV1 converts team_ids from TypeSet to sorted TypeList.
// This addresses DEVTOOLING-1497 where the SDK gRPC layer was forcing planned TypeSet
// values into state on apply errors, causing state corruption.
Helps future readers understand the migration context without digging through Jira.

genesyscloud/recording_media_retention_policy/resource_genesyscloud_recording_media_retention_policy.go
shrutisuryawanshigenesys: I tested removing Partial(true) and it does change behavior, the failed update still wrote invalid team_ids into state. 
From [resource_data.go](https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.40.0/helper/schema/resource_data.go#L298): Partial preserves prior state on apply error, and [State()](https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.40.0/helper/schema/resource_data.go#L459) uses getSourceState when d.partial is set

sureshperiyappan: thanks for the clarification

sureshperiyappan (approved):

PR #2412, opened 2026-07-06
Ep 370 routing email domain graph imap settings

SvetaB11:
Re-create EP-370 after it was merged to dev (#2383) and reverted (#2383 revert, 3810aa948).
Adds `graph_api_settings` and `imap_settings` to `genesyscloud_routing_email_domain` so Graph/IMAP email domains are preserved on CX as Code export and cross-org promotion (customer issue: promoted domains lost Graph type and became Custom).

The first merge caused test-cx-as-code failures: cross-org export included computed graph_api_settings.status and imap_settings.status in generated Terraform config. Apply then failed with:
`Error: Value for unconfigurable attribute
Can't configure a value for "graph_api_settings.0.status": its value will be decided automatically`

**Fix in this PR**
Use export-only UnResolvableAttributes schemas (graphApiSettingsExportSchema / imapSettingsExportSchema) that include integration_id only, not status.

**Export / promotion:** only integration_id in genesyscloud.tf and terraform.tfvars
**Resource schema unchanged:** status remains computed; Read still populates it in state after apply
**Unit test:** TestUnitRoutingEmailDomainExporter_UnResolvableSchemasOmitComputedStatus
Branch rebased on latest dev (includes DEVTOOLING-1605 subdomain tests and DEVTOOLING-1724 create race handling).


sanjaysg21 (approved):

genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_schema.go
```diff
@@ -53,6 +83,20 @@ func ResourceRoutingEmailDomain() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 			},
+			"imap_settings": {
+				Description: "The IMAP server integration and settings to use for processing inbound emails.",
+				Type:        schema.TypeList,
+				Optional:    true,
```
sureshperiyappan, line 89: The imap_settings and graph_api_settings blocks contain a Computed-only child attribute (status) but the parent block is only Optional. Per this provider's established convention (see user.employer_info, user.routing_utilization, integration.config), blocks with server-populated children also set Computed: true on the parent. While SDKv2 tolerates this via its legacy compatibility layer (demoting the error to a warning), it can cause issues when status is referenced by other resources and creates tech debt for any future framework migration. Recommend adding Computed: true to both parent blocks.

genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_schema.go
```diff
@@ -53,6 +83,20 @@ func ResourceRoutingEmailDomain() *schema.Resource {
...
+				MaxItems:    1,
+				Elem:        imapSettingsResource,
+			},
+			"graph_api_settings": {
+				Description: "The Graph API server integration and settings to use for processing inbound and outbound emails.",
+				Type:        schema.TypeList,
+				Optional:    true,
```
sureshperiyappan, line 97: The imap_settings and graph_api_settings blocks contain a Computed-only child attribute (status) but the parent block is only Optional. Per this provider's established convention (see user.employer_info, user.routing_utilization, integration.config), blocks with server-populated children also set Computed: true on the parent. While SDKv2 tolerates this via its legacy compatibility layer (demoting the error to a warning), it can cause issues when status is referenced by other resources and creates tech debt for any future framework migration. Recommend adding Computed: true to both parent blocks.

genesyscloud/routing_email_domain/resource_genesyscloud_routing_email_domain_schema.go
SvetaB11: Great catch, thanks for the review. Fixed both.

sureshperiyappan (approved):

PR #2423, opened 2026-07-09
Bug/devtooling-1714: Update genesyscloud_quality_forms_evaluation and genesyscloud_quality_forms_survey resources

shrutisuryawanshigenesys:


genesyscloud/quality_forms_evaluation/resource_genesyscloud_quality_forms_evaluation_utils.go
```diff
@@ -312,6 +483,12 @@ func flattenQuestions(questions *[]platformclientv2.Evaluationquestion) []interf
 			questionMap["multiple_select_option_questions"] = flattenMultipleSelectOptionQuestions(question.MultipleSelectOptionQuestions)
 			log.Printf("flattenQuestions: Question '%s' has %d multiple_select_option_questions", questionText, len(*question.MultipleSelectOptionQuestions))
 		}
+		if question.DefaultAnswer != nil && question.DefaultAnswer.Id != nil {
```
sanjaysg21, line 513: This stores the answer option's Id, but answer_options.id is in ExcludedAttributes in the exporter so those IDs are stripped during export. When a user exports from one environment and imports to another, default_answer_id will contain a UUID that doesn't exist in the target environment. I beleive we should use context_id instead of id(Not sure please cross verify). 
Same change would be needed in the build function where it reads default_answer_id and similarly for the multiple_select_option_questions variant of this code below.

genesyscloud/quality_forms_evaluation/resource_genesyscloud_quality_forms_evaluation.go
```diff
@@ -176,6 +194,16 @@ func updateEvaluationForm(ctx context.Context, d *schema.ResourceData, meta inte
 			QuestionGroups: buildSdkQuestionGroups(d),
 		}
 
+		if evaluationSettings, ok := d.Get("evaluation_settings").([]interface{}); ok && len(evaluationSettings) > 0 {
```
sanjaysg21: If a user removes the evaluation_settings block from their config after initially setting it, this condition is false and nothing is sent to the API, the previous settings (revisions/disputes enabled) will remain active on the form. This breaks the Terraform contract that config represents desired state. We might need an else branch that sends disabled defaults to explicitly clear the settings. Same applies in both createEvaluationForm and updateEvaluationForm.

sureshperiyappan (approved):

PR #2424, opened 2026-07-09
RULES-1399 Update CX as code to use bulk operations

Jx-hou:
Use Genesys Cloud bulk APIs for business_rules_decision_table row create/update/delete instead of per-row requests. Rows are chunked to API limits (add/update: 15, remove: 49) to improve performance and reliability on large tables.


genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table.go
```diff
@@ -235,50 +233,6 @@ func readBusinessRulesDecisionTable(ctx context.Context, d *schema.ResourceData,
...
-// reported duplicate index matches the 1-based position of the row currently
-// being added. This indicates the row was already created server-side by a prior
-// attempt - typically a POST that returned 504 at the gateway but succeeded on
-// the backend, whose non-idempotent retry then collides with the ghost row. In
-// that case the add is safe to treat as a no-op and continue.
-//
-// The index guard is what keeps this from masking a genuine duplicate in config:
-// a real config duplicate reports the index of the earlier (first) occurrence,
-// which will not equal the position of the row currently being added.
-func isDuplicateRowAtIndex(resp *platformclientv2.APIResponse, rowNumber int) bool {
```
brlenn, line 281: This was added as we were seeing duplicate rows due to timeouts in the sdk, can you confirm we are no longer seeing these?

Jx-hou: `isDuplicateRowAtIndex` is removed because it was tied to the per-row POST loop. The underlying 504 + SDK retry race can still occur at the bulk-chunk level (up to 15 rows per request). The Lexicon API still returns decision.table.duplicate.row 409 on retry-after-success, we no longer treat matching-index 409s as idempotent success. 

When we get a 409 from a bulk request,  we should not treat the whole 15 rows were written correctly right. Should we read back the table rows on bulk-add failure and verify that every row in the chunk is already present with matching content? If they are, we continue, otherwise we fail.

Jx-hou: `isDuplicateRowAtIndex` is removed because it was tied to the per-row POST loop. The underlying 504 + SDK retry race can still occur at the bulk-chunk level (up to 15 rows per request). The Lexicon API still returns decision.table.duplicate.row 409 on retry-after-success, we no longer treat matching-index 409s as idempotent success.

When we get a 409 from a bulk request, we should not treat the whole 15 rows were written correctly right. Should we read back the table rows on bulk-add failure and verify that every row in the chunk is already present with matching content? If they are, we continue, otherwise we fail.

brlenn: For the bulk request it uses a transaction right so if there is even one duplicate the entire batch is not written. Reading back tables rows is not ideal as what if its 000s of rows. I believe our APIs returns a similar response to the single row api for duplicates. 

Could we restore the per-row guard at chunk granularity, on a `decision.table.duplicate.row 409`, accept the chunk only if the reported duplicate index falls in that chunk's expected appended range `[priorRowCount+start+1 .. priorRowCount+start+len(chunk)]`. A genuine duplicate collides with a pre-existing row at a lower index and is outside that range, so it still fails. This needs no extra GETs and scales to large tables.

This must key off the error code, not the 409 status. Bulk add/update can also return `RULE_PERSISTENCE_CONFLICT (409) `for a version-lock/concurrency conflict where nothing was written we should not skip those. 

The one case this doesn't cover is a bare 504 with no response body (no index to check), which we fail closed on, same as now.

genesyscloud/business_rules_decision_table/resource_genesyscloud_business_rules_decision_table_create_unit_test.go
```diff
@@ -129,171 +128,3 @@ func TestUnitRollbackDecisionTableNotBlockedByExpiredRequestContext(t *testing.T
...
-				},
-			},
-		},
-	}
-}
-
-// TestUnitApplyRowChangesSkipsGhostRowDuplicate verifies the update-path add-rows
-// loop treats a matching-index 409 duplicate.row as an already-created ghost row
-// and continues, using priorRowCount+i+1 as the expected index base.
-func TestUnitApplyRowChangesSkipsGhostRowDuplicate(t *testing.T) {
```
brlenn, line 182: Do we need to add back in these tests for bulk

genesyscloud/business_rules_decision_table/decision_table_bulk_ghost_guard.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/mypurecloud/platform-client-sdk-go/v193/platformclientv2"
+)
+
+// decisionTableDuplicateRowCode is the API error code returned when a row being
+// added duplicates a row that already exists on the table version.
+const decisionTableDuplicateRowCode = "decision.table.duplicate.row"
+
+// rulePersistenceConflictCode is returned for version-lock/concurrency conflicts
+// where nothing was written. It must not be treated as a ghost-chunk success.
+const rulePersistenceConflictCode = "RULE_PERSISTENCE_CONFLICT"
```
brlenn: Dont think this is used anywhere

brlenn (approved):

sanjaysg21 (approved):

sureshperiyappan:
@Jx-hou ,
local-test/ fixtures must not be committed — 
main.tf
 (~1.5M lines), .terraform.lock.hcl, and the entire stress-test-8000-rows/ tree. These are local scratch/stress artifacts. A generated 1.5M-line .tf plus a lock file bloats the repo, will wreck diffs/clones, and isn't wired into any acceptance test. Please drop the whole local-test/ directory from the PR (add to .gitignore if useful locally). If the generator is worth keeping, keep only generate.py in an appropriate examples/test location.

sureshperiyappan (commented):
Comment provided

Jx-hou:
Hi @sureshperiyappan , thanks for the review. I have removed the additional test files. Let me know if it looks good now.

sureshperiyappan (approved):

PR #2439, opened 2026-07-13
bug/DEVTOOLING-1689 : remove audio when switching a language from file to TTS

shrutisuryawanshigenesys:


genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
```diff
@@ -319,6 +319,17 @@ func createOrUpdateArchitectUserPromptResourcesFn(ctx context.Context, p *archit
 		return resp, err
 	}
 
+	// Deletes run before creates so a file->TTS language is removed before being recreated.
```
sureshperiyappan, line 322: Can we add an acceptance test that covers the file→TTS switch? Something like: create a prompt with a wav file, then update the same language to TTS only. That way the delete+recreate path has CI coverage and won't regress silently.

genesyscloud/architect_user_prompt/genesyscloud_architect_user_prompt_proxy.go
```diff
@@ -319,6 +319,17 @@ func createOrUpdateArchitectUserPromptResourcesFn(ctx context.Context, p *archit
...
+	for _, language := range resourcesToDelete {
+		log.Printf("Deleting user prompt resource for language: %s", language)
+		resp, err = p.deleteArchitectUserPromptResource(ctx, d.Id(), language)
+		if err != nil {
+			return resp, fmt.Errorf("failed to delete user prompt resource for language '%s': %w", language, err)
+		}
+
+		allLanguages = removeByValue(allLanguages, language)
```
sureshperiyappan: allLanguages is empty at this point so removeByValue is a no-op here. Safe to remove this line.

sureshperiyappan (approved):

PR #2441, opened 2026-07-15
SQUONK-968 add video chat channels resources

fmarc-genesys:
### Add channels attribute to video settings in web deployments configuration

Adds a new channels attribute to the video block of the genesyscloud_webdeployments_configuration resource. This enables CXasCode support for the "Video Chat from Voice Call" feature (PURE-7232), allowing users to configure which channels support video chat escalation (Webmessaging, Voice).

### DIFF
--- a/genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
+++ b/genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
@@ -549,6 +549,15 @@ var (
 				Optional:    true,
 				Computed:    true,
 			},
+			"channels": {
+				Description: "List of channels through which video chat is available",
+				Type:        schema.TypeList,
+				Optional:    true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{"Webmessaging", "Voice"}, false),
+				},
+			},
 			"agent": {
 				Description: "Video Settings for agent",
 				Type:        schema.TypeList,
--- a/genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
+++ b/genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
@@ -553,6 +553,7 @@ var (
 				Description: "List of channels through which video chat is available",
 				Type:        schema.TypeList,
 				Optional:    true,
+				Computed:    true,
 				Elem: &schema.Schema{
 					Type:         schema.TypeString,
 					ValidateFunc: validation.StringInSlice([]string{"Webmessaging", "Voice"}, false),
--- a/genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_test.go
+++ b/genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_test.go
@@ -175,6 +175,9 @@ func TestAccResourceWebDeploymentsConfigurationComplex(t *testing.T) {
 		channels       = []string{strconv.Quote("Webmessaging")}
 		channelsUpdate = []string{strconv.Quote("Webmessaging"), strconv.Quote("Voice")}
 
+		videoChannels       = []string{strconv.Quote("Webmessaging")}
+		videoChannelsUpdate = []string{strconv.Quote("Webmessaging"), strconv.Quote("Voice")}
+
 		authenticationSettings1 = authSettings{
 			enabled:             true,
 			integrationId:       uuid.NewString(),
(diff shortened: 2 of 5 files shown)


genesyscloud/outbound_campaign/resource_genesyscloud_outbound_campaign_unit_test.go
```diff
@@ -1,174 +1,174 @@
...
-                    "diagnostics_settings should be set for %s mode", tc.dialingMode)
-                settingsMap := diagSettings[0].(map[string]interface{})
-                assert.Equal(t, true, settingsMap["report_low_max_calls_per_agent_alert"])
-            } else {
-                assert.Equal(t, 0, len(diagSettings),
-                    "diagnostics_settings should NOT be set for %s mode", tc.dialingMode)
-            }
-        })
-    }
+	t.Parallel()
```
fmarc-genesys, line 15: Unrelated to my changes, but Github Actions wanted me to run `go fmt ./genesyscloud/...`

docs/resources/routing_email_route.md
bbbco, line 1: Please rebase on `dev` branch and rerun `make docs`

fmarc-genesys: Rebased and targeting now dev branch

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
```diff
@@ -549,6 +549,15 @@ var (
 				Optional:    true,
 				Computed:    true,
 			},
+			"channels": {
```
sureshperiyappan, line 552: The enabled attribute just above has Computed: true set alongside Optional: true. If the API ever returns a default value for channels when it's not explicitly set in config, the lack of Computed: true here will cause a perpetual plan diff (Terraform will see a difference between the empty config and what the API returns). Is there a reason this doesn't have Computed: true? If the API can populate this field on its own, you'll need to add it.

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_utils.go
```diff
@@ -269,6 +269,11 @@ func buildVideoSettings(d *schema.ResourceData) *platformclientv2.Videosettings
 		videoSettings.Enabled = &v
 	}
 
+	if v, ok := cfg["channels"]; ok {
```
sureshperiyappan, line 272: If the user omits channels from their config, cfg["channels"] will still exist as a key with an empty []interface{} (because it's a TypeList). This means you'll always set videoSettings.Channels to a pointer to an empty slice &[]string{}, which might cause unintended behavior on the API side. Consider adding a length check:


if v, ok := cfg["channels"]; ok {
    channels := lists.InterfaceListToStrings(v.([]interface{}))
    if len(channels) > 0 {
        videoSettings.Channels = &channels
    }
}

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_test.go
```diff
@@ -175,6 +175,9 @@ func TestAccResourceWebDeploymentsConfigurationComplex(t *testing.T) {
 		channels       = []string{strconv.Quote("Webmessaging")}
 		channelsUpdate = []string{strconv.Quote("Webmessaging"), strconv.Quote("Voice")}
 
+		videoChannels       = []string{strconv.Quote("Webmessaging")}
```
sureshperiyappan, line 178: Both test steps include channels in the video config. There's no test step that omits channels entirely to verify the resource handles the absence gracefully (especially relevant given comment #2 above about empty list handling). Consider adding a step or separate test that creates/updates video settings without specifying channels.

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_schema.go
fmarc-genesys: Indeed, computed field has been added.

genesyscloud/webdeployments_configuration/utils/resource_genesyscloud_webdeployments_configuration_utils.go
fmarc-genesys: Good point, length check added.

genesyscloud/webdeployments_configuration/resource_genesyscloud_webdeployments_configuration_test.go
fmarc-genesys: Tests updated

sureshperiyappan (approved):

PR #2442, opened 2026-07-15
fix(knowledge_document): Export label_names and category_name as refe…

Teja-genesys:
## Summary
Fixes knowledge document export to produce proper Terraform references for
label_names and category_name, and ensures the dependent label and category
resources are exported when enable_dependency_resolution=true.

## Problem
When exporting a knowledge_document, label_names and category_name were exported
as hardcoded strings. The referenced label and category resources were not included
in the export. This caused terraform apply to hang in fresh orgs because the
labels/categories didn't exist.

## Changes
- `resource_genesyscloud_knowledge_document_schema.go`: Added category_name custom
  resolver, top-level label_ids/category_id fields with RefAttrs for dependency
  resolution, and cleanup resolvers
- `resource_genesyscloud_knowledge_document.go`: Set top-level label_ids and
  category_id in Read for export dependency resolution
- `resource_exporter_custom.go`: Added KnowledgeDocumentCategoryNameResolver,
  modified label resolver to use global registry fallback, added cleanup resolvers

## Testing
- Created knowledge base, label, category, and document in test org
- Exported with include_filter_resources and enable_dependency_resolution=true
- Verified exported .tf contains references (not strings) for label_names and category_name
- Verified label and category resources are included in export output
- Verified label_ids and category_id internal fields do not appear in export
- Unit tests pass

### DIFF
--- a/genesyscloud/knowledge_document/resource_genesyscloud_knowledge_document_schema.go
+++ b/genesyscloud/knowledge_document/resource_genesyscloud_knowledge_document_schema.go
@@ -30,9 +30,14 @@ func KnowledgeDocumentExporter() *resourceExporter.ResourceExporter {
 		GetResourcesFunc: provider.GetAllWithPooledClient(getAllKnowledgeDocuments),
 		RefAttrs: map[string]*resourceExporter.RefAttrSettings{
 			"knowledge_base_id": {RefType: "genesyscloud_knowledge_knowledgebase"},
+			"label_ids":         {RefType: "genesyscloud_knowledge_label"},
+			"category_id":       {RefType: "genesyscloud_knowledge_category"},
 		},
 		CustomAttributeResolver: map[string]*resourceExporter.RefAttrCustomResolver{
-			"knowledge_document.label_names": {ResolverFunc: resourceExporter.KnowledgeDocumentLabelNamesResolver},
+			"knowledge_document.label_names":   {ResolverFunc: resourceExporter.KnowledgeDocumentLabelNamesResolver},
+			"knowledge_document.category_name": {ResolverFunc: resourceExporter.KnowledgeDocumentCategoryNameResolver},
+			"label_ids":                        {ResolverFunc: resourceExporter.RemoveLabelIdsResolver},
+			"category_id":                      {ResolverFunc: resourceExporter.RemoveCategoryIdResolver},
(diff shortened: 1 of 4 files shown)


docs/resources/routing_email_route.md
bbbco, line 1: Please rebase on the `dev` branch and rerun `make docs`

genesyscloud/knowledge_document/resource_genesyscloud_knowledge_document_schema.go
bbbco, line 1: This is really a very interesting strategy for handling this sort of scenario, and it could probably come in handy in other use cases.

However, I'm concerned that the `label_ids` and `category_id` fields will still be visible to the user (documentation, etc) and they may think they can/should set them despite the description (i.e., if a user just exports and looks at the configuration data itself).

I wonder if we should have some sort of convention for these types of "Exporter-only" fields that we make explicit to the user that they should not be setting manually. Could we maybe wrap these attribute in a `tf_exporter` map, and update the documentation generator to remove any fields in `tf_exporter` from being generated in the documentation? This could be an explicit indicator to the user not to mess with these fields.

cc @sureshperiyappan @sanjaysg21 for thoughts.

Teja-genesys: Thanks Brian. The label_ids and category_id fields don't appear in exported HCL — the RemoveLabelIdsResolver and RemoveCategoryIdResolver strip them before output. So users won't see them when looking at exported config.

They do show up in the auto-generated docs as Optional fields, but they're functionally harmless — even if a user sets them manually, the provider overwrites them with the real API values on the next read. The description says "Do not set manually."

I tried Computed: true only to make them invisible, but the exporter pipeline strips computed-only fields before RefAttrs can process them — so dependency resolution breaks. Optional + Computed is the minimum needed for the exporter to function.

cc: @sureshperiyappan , @sanjaysg21

sureshperiyappan: I have created https://inindca.atlassian.net/browse/DEVTOOLING-1795 ticket to discuss and work on it.

sureshperiyappan (approved):

PR #2445, opened 2026-07-17
Feat/relate 25224 routing queue identity resolution

BlekiGenesys:
## Summary

- Adds `genesyscloud_routing_queue_identity_resolution` to manage queue identity resolution settings in Terraform (RELATE-25224 pilot).
- Supports create, update, import, export, and destroy (restore to the platform default).
- Includes tests, examples, and docs.


istvan-bihari-genesys (approved):

nduchenk (approved):

genesyscloud/routing_queue_identity_resolution/genesyscloud_routing_queue_identity_resolution_proxy.go
```diff
@@ -0,0 +1,73 @@
+package routing_queue_identity_resolution
```
sureshperiyappan, line 1: Other proxies in the repo (e.g., `flow_loglevel`, `routing_queue`) include a block comment at the top describing the proxy composition pattern and its numbered conventions. Please add a similar doc block for consistency:

/*
The genesyscloud_routing_queue_identity_resolution_proxy.go file contains the proxy structures and methods that interact
with the Genesys Cloud SDK. We use composition here for each function on the proxy so individual functions can be stubbed
out during testing.
*/

genesyscloud/routing_queue_identity_resolution/genesyscloud_routing_queue_identity_resolution_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (p *routingQueueIdentityResolutionProxy) getRoutingQueueById(ctx context.Context, queueId string) (*platformclientv2.Queue, *platformclientv2.APIResponse, error) {
+	return p.getRoutingQueueByIdAttr(ctx, p, queueId)
+}
+
+func getRoutingQueueIdentityResolutionFn(ctx context.Context, p *routingQueueIdentityResolutionProxy, queueId string) (*platformclientv2.Identityresolutionqueueconfig, *platformclientv2.APIResponse, error) {
+	ctx = provider.EnsureResourceContext(ctx, ResourceType)
+	return p.routingApi.GetRoutingQueueIdentityresolution(queueId)
+}
```
sureshperiyappan, line 69: The `ctx` reassignment looks like dead code since the SDK method doesn't accept a context parameter. Other proxies in the repo always annotate this with `// Set resource context for SDK debug logging`. Please add the comment here (and on lines 68 and 73 for the other two Fn implementations) to clarify intent.

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_schema.go
```diff
@@ -0,0 +1,82 @@
...
+	return &schema.Resource{
+		Description:   `Genesys Cloud routing queue identity resolution settings. Destroy restores the queue to its default identity resolution configuration (resolve_identities = true, unassigned division).`,
+		CreateContext: provider.CreateWithPooledClient(createRoutingQueueIdentityResolution),
+		ReadContext:   provider.ReadWithPooledClient(readRoutingQueueIdentityResolution),
+		UpdateContext: provider.UpdateWithPooledClient(updateRoutingQueueIdentityResolution),
+		DeleteContext: provider.DeleteWithPooledClient(deleteRoutingQueueIdentityResolution),
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		SchemaVersion: 1,
```
sureshperiyappan: Question: Why is `SchemaVersion` set to `1`? New resources in this repo typically start at `0` (or omit the field). Version `1` implies a migration from a prior schema, but no `StateUpgraders` are defined. If this is intentional, please add a comment explaining why. Otherwise, change to `SchemaVersion: 0`.

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_schema.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+var callOnBehalfOfQueueSchema = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"resolve_identities": {
+			Description: "Whether identities should be resolved.",
+			Type:        schema.TypeBool,
+			Required:    true,
+		},
+		"division_id": {
```
sureshperiyappan, line 32: Suggestion: Consider adding `Computed: true` alongside `Optional: true` for `division_id`. Since the read function conditionally omits `division_id` from state when the API returns `*` or empty, and import may populate a value the user never set, `Computed: true` makes the schema semantics explicit and avoids potential "value not expected" errors during import scenarios.

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v193/platformclientv2"
+)
+
+func getAllRoutingQueueIdentityResolution(ctx context.Context, clientConfig *platformclientv2.Configuration) (resourceExporter.ResourceIDMetaMap, diag.Diagnostics) {
+	resources := make(resourceExporter.ResourceIDMetaMap)
+	proxy := getRoutingQueueIdentityResolutionProxy(clientConfig)
+
+	queues, resp, err := proxy.routingQueueProxy.GetAllRoutingQueues(ctx, "", false)
```
sureshperiyappan, line 27: Bug: If `GetAllRoutingQueues` returns `(nil, resp, nil)` (no error but nil pointer), the dereference on `*queues` at line 37 will panic. Please add a nil check:

if queues == nil {
    return resources, nil
}

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_test.go
```diff
@@ -0,0 +1,207 @@
...
+	var (
+		identityResolutionResourceLabel = "test-identity-resolution"
+		queueResourceLabel              = "test-queue"
+		queueName                       = "Terraform Test Queue-" + uuid.NewString()
+		homeDivisionConfig              = gcloud.GenerateAuthDivisionHomeDataSource(homeDivisionDataSourceLabel)
+	)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { util.TestAccPreCheck(t) },
+		ProviderFactories: provider.GetProviderFactories(providerResources, providerDataSources),
```
sureshperiyappan, line 33: Suggestion: Most resources in this repo define a `CheckDestroy` function on `resource.TestCase` (see `user`, `team`, `auth_role`). Since this resource's destroy resets to defaults, you could reuse the existing `verifyIdentityResolutionDefault` function:

CheckDestroy: verifyIdentityResolutionDefault("genesyscloud_routing_queue." + queueResourceLabel),

This provides a safety net if a test step fails partway through, ensuring cleanup verification still runs.

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_test.go
```diff
@@ -0,0 +1,207 @@
...
+		expectedDivisionId := ""
+		if divisionStateRef != "" {
+			divisionResource, ok := state.RootModule().Resources[divisionStateRef]
+			if !ok {
+				return fmt.Errorf("failed to find division %s in state", divisionStateRef)
+			}
+			expectedDivisionId = divisionResource.Primary.ID
+		}
+
+		routingApi := platformclientv2.NewRoutingApi()
```
sureshperiyappan: `platformclientv2.NewRoutingApi()` uses the SDK's global/default configuration rather than the provider's configured client. This works because the test framework sets up global auth, but it's fragile and inconsistent with how the resource itself operates (via `meta.(*provider.ProviderMeta).ClientConfig`). Consider using `platformclientv2.NewRoutingApiWithConfig(...)` with the provider config if accessible, or at minimum add a comment explaining why the global client is acceptable here.

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_utils.go
```diff
@@ -0,0 +1,106 @@
...
+	if divisionId, ok := block["division_id"].(string); ok && !isUnassignedDivisionId(divisionId) {
+		config.Division = &platformclientv2.Writablestarrabledivision{
+			Id: platformclientv2.String(divisionId),
+		}
+	}
+
+	return config, nil
+}
+
+func flattenCallOnBehalfOfQueue(config *platformclientv2.Outboundqueueidentityresolutionconfig) []interface{} {
```
sureshperiyappan, line 54: Bug (defensive): If `config.ResolveIdentities` is nil (unexpected API response), the flattened map won't include `resolve_identities`. Since it's a `Required` field in the schema, Terraform will error on the next plan. Add a default to match `buildDefaultIdentityResolutionQueueConfig`:

if config.ResolveIdentities != nil {
    result["resolve_identities"] = *config.ResolveIdentities
} else {
    result["resolve_identities"] = true
}

genesyscloud/routing_queue_identity_resolution/genesyscloud_routing_queue_identity_resolution_init_test.go
```diff
@@ -0,0 +1,47 @@
+package routing_queue_identity_resolution
+
+import (
```
sureshperiyappan, line 3: Go convention (and this repo's style) groups stdlib imports first, then a blank line, then third-party/internal packages. Currently `"sync"` and `"testing"` are mixed in after the aliased imports. Please reorder:

import (
    "sync"
    "testing"

    "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
    gcloud "github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud"
    routingQueue "github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/routing_queue"
)

genesyscloud/routing_queue_identity_resolution/genesyscloud_routing_queue_identity_resolution_proxy.go
BlekiGenesys: comment added

genesyscloud/routing_queue_identity_resolution/genesyscloud_routing_queue_identity_resolution_proxy.go
BlekiGenesys: comments added

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_schema.go
BlekiGenesys: set to 0, had no special reason

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution.go
BlekiGenesys: nil check added

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_utils.go
BlekiGenesys: default true added

genesyscloud/routing_queue_identity_resolution/genesyscloud_routing_queue_identity_resolution_init_test.go
BlekiGenesys: imports reordered

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_test.go
BlekiGenesys: provider's configured client is used now

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_test.go
BlekiGenesys: I wired CheckDestroy, but didn’t reuse verifyIdentityResolutionDefault directly.

That helper assumes the parent queue is still present (it’s also used in the last step to assert reset-to-default while the queue exists). CheckDestroy runs after the final destroy, when the queue is often already gone, so that would fail with a missing queue / 404 even when destroy worked.

Instead I added testVerifyRoutingQueueIdentityResolutionDestroyed, which follows the usual package pattern: iterate IR resources in pre-destroy state, assert the config was reset to default, and treat parent 404 as success. The last step still uses verifyIdentityResolutionDefault for the strict “queue still exists” check.

genesyscloud/routing_queue_identity_resolution/resource_genesyscloud_routing_queue_identity_resolution_schema.go
BlekiGenesys: We intentionally left division_id as Optional only (not Computed).

Read omits * / empty from state and only keeps real UUIDs, and suppressUnassignedDivisionIdDiff treats "" and "*" as equivalent so explicit "*" doesn’t perpetual-plan against omitted state.

Adding Computed: true would change unset semantics: when division_id is omitted from config, Terraform may treat an existing state value as provider-owned instead of planning to clear it back to unassigned. That fights the “omit = unassigned” model we want for this field.

Happy to revisit if you hit a concrete import / “value not expected” case — we just haven’t seen that with the current approach.

sureshperiyappan (approved):

PR #2448, opened 2026-07-21
Devtooling 1731

sanjaysg21:


genesyscloud/agentic_virtual_agent_version/resource_genesyscloud_agentic_virtual_agent_version_schema.go
```diff
@@ -0,0 +1,478 @@
...
+				MaxItems:    1,
+				Elem:        definitionResource(),
+			},
+		},
+	}
+}
+
+func definitionResource() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
```
rbrennangen, line 68: @sanjaysg21 - as discussed there will be a new feature toggled 'model' field coming here also (values: Stable / Preview).
Thanks!

sanjaysg21: Added!

examples/resources/genesyscloud_agentic_virtual_agent/resource.tf
```diff
@@ -0,0 +1,4 @@
+resource "genesyscloud_agentic_virtual_agent" "example_agent" {
```
sureshperiyappan, line 1: The resource examples look great — nice work there. One gap: since the package registers a data source (DataSourceAgenticVirtualAgent), there should be a matching 
data-source.tf
 file. Other resources that expose data sources (e.g. guide) have this. Without it, go generate / make docs will produce incomplete documentation.

examples/data-sources/genesyscloud_agentic_virtual_agent/ (missing)

genesyscloud/agentic_virtual_agent_version/genesyscloud_agentic_virtual_agent_version_proxy.go
```diff
@@ -0,0 +1,130 @@
...
+   that interact with the Genesys Cloud API for AVA versions.
+
+   API Endpoints:
+   - POST   /api/v2/agentic/virtualagents/{virtualAgentId}/versions           → Create version
+   - GET    /api/v2/agentic/virtualagents/{virtualAgentId}/versions/{versionId} → Read version
+   - PATCH  /api/v2/agentic/virtualagents/{virtualAgentId}/versions/{versionId} → Update version
+   - DELETE → Does not exist. Version destroy is state-only (no-op).
+*/
+
+const ResourceType = "genesyscloud_agentic_virtual_agent_version"
```
sureshperiyappan: Per repo convention, const ResourceType belongs in resource_genesyscloud_*_schema.go. I checked every other package in the repo and these two are the only ones that define it in the proxy file. The base agentic_virtual_agent package already does it correctly in its schema file. Could you move these to the respective *_schema.go files for consistency?

genesyscloud/agentic_virtual_agent_version_publish/genesyscloud_agentic_virtual_agent_version_publish_proxy.go
```diff
@@ -0,0 +1,92 @@
...
+   genesyscloud_agentic_virtual_agent_version_publish_proxy.go contains the proxy for
+   managing AVA version publish jobs.
+
+   API Endpoints:
+   - POST  /api/v2/agentic/virtualagents/{agentId}/versions/{versionId}/jobs     → Create publish job
+   - GET   /api/v2/agentic/virtualagents/{agentId}/versions/{versionId}/jobs/{jobId} → Poll job status
+   - GET   /api/v2/agentic/virtualagents/{agentId}/versions/{versionId}           → Read version status
+*/
+
+const ResourceType = "genesyscloud_agentic_virtual_agent_version_publish"
```
sureshperiyappan: Per repo convention, const ResourceType belongs in resource_genesyscloud_*_schema.go. I checked every other package in the repo and these two are the only ones that define it in the proxy file. The base agentic_virtual_agent package already does it correctly in its schema file. Could you move these to the respective *_schema.go files for consistency?

genesyscloud/agentic_virtual_agent_version_publish/resource_genesyscloud_agentic_virtual_agent_version_publish.go
```diff
@@ -0,0 +1,153 @@
...
+		if util.IsStatus404(resp) {
+			log.Printf("Version %s/%s not found, removing publish from state", agentId, versionId)
+			d.SetId("")
+			return nil
+		}
+		return util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to read version %s/%s: %s", agentId, versionId, getErr), resp)
+	}
+
+	currentStatus := ""
+	if versionResp.Status != nil {
```
sureshperiyappan, line 130: If getErr is nil but versionResp itself comes back as nil (edge case with custom API client), this line panics:

if versionResp.Status != nil {
Suggest adding a nil guard:

if versionResp == nil {
    d.SetId("")
    return nil
}

genesyscloud/agentic_virtual_agent_version/resource_genesyscloud_agentic_virtual_agent_version_test.go
```diff
@@ -0,0 +1,189 @@
+package agentic_virtual_agent_version
```
sureshperiyappan, line 1: The tests cover create and import but there's no test step that modifies the definition and verifies the PATCH path. Since UpdateContext is implemented, it should be exercised. Could you add a step to TestAccResourceAgenticVirtualAgentVersionBasic that changes the role or adds an instruction, then asserts the updated values? This is the same pattern the base agent test uses (create → update name → update image → import).

genesyscloud/agentic_virtual_agent_version_publish/resource_genesyscloud_agentic_virtual_agent_version_publish.go
```diff
@@ -0,0 +1,153 @@
...
+			return retry.RetryableError(fmt.Errorf("publish job %s still in progress", jobId))
+		case "Succeeded":
+			log.Printf("Publish job %s succeeded. Version %s/%s is now %s", jobId, agentId, versionId, status)
+			return nil
+		case "Failed":
+			errMsg := "publish job failed"
+			if len(jobStatus.Errors) > 0 {
+				errMsg = fmt.Sprintf("publish job failed: %s (code: %s)", jobStatus.Errors[0].Message, jobStatus.Errors[0].Code)
+			}
+			return retry.NonRetryableError(fmt.Errorf("%s for version %s/%s: %s", errMsg, agentId, versionId, errMsg))
```
sureshperiyappan: errMsg is printed twice here:

return retry.NonRetryableError(fmt.Errorf("%s for version %s/%s: %s", errMsg, agentId, versionId, errMsg))
Produces output like: "publish job failed: X (code: Y) for version abc/1.0: publish job failed: X (code: Y)"

Should be:

return retry.NonRetryableError(fmt.Errorf("publish failed for version %s/%s: %s", agentId, versionId, errMsg))

sureshperiyappan (approved):

PR #2459, opened 2026-07-23
IAM-4784 idp redirect forceauthn cx as code

wwallgenesys:
https://inindca.atlassian.net/browse/IAM-4784

As a customer who uses Terraform to make changes on Genesys Cloud, I want to be able to configure Inactivity Timeout IdP Redirect via Terraform so that I can make all my configuration changes via a single tool.

CX as Code has switched to a contribution model and it is the responsibility of the service team to make sure that their configuration settings are available via Terraform.

### DIFF
--- a/docs/resources/routing_email_route.md
+++ b/docs/resources/routing_email_route.md
@@ -54,7 +54,7 @@ resource "genesyscloud_routing_email_route" "example_route" {
     enabled            = false
     canned_response_id = genesyscloud_responsemanagement_response.example_signature_response.id
     always_included    = false
-    inclusion_type     = "FirstResponseOnly"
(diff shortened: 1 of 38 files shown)


docs/resources/routing_email_route.md
```diff
@@ -54,7 +54,7 @@ resource "genesyscloud_routing_email_route" "example_route" {
     enabled            = false
     canned_response_id = genesyscloud_responsemanagement_response.example_signature_response.id
     always_included    = false
-    inclusion_type     = "FirstResponseOnly"
```
wwallgenesys, line 57: Side effect of go generate

sureshperiyappan (approved):

PR #2460, opened 2026-07-23
RULES-1478: Decision table CSV import/export

brlenn:
- Add CSV-backed rows for genesyscloud_business_rules_decision_table via rows_csv_filepath (plus computed rows_csv_content_hash / rows_record_count), aligned with the outbound contact-list pattern.
- Deprecate nested rows under ExactlyOneOf with CSV; create/update use Lexicon Replace CSV import (provider reinjects empty rowId), then publish.
tf_export writes Populated CSV with rowId stripped and emits rows_csv_filepath instead of nested rows. 
- Skip nested row paging on CSV read; skip nested-row update when migrating to CSV; during export Refresh, skip row fetch without clearing nested rows (avoids a non-empty plan). Drop MinItems: 1 on rows so CSV-only configs validate.
- Docs, examples, and migration notes updated; ACC coverage for CSV happy path, validation, queue names, migrate, and export.

NOTE, For ACC tests permissions will need to be added to test orgs


NaeemTai (approved):

genesyscloud/business_rules_decision_table/decision_table_csv.go
```diff
@@ -0,0 +1,548 @@
...
+		return "", "", fmt.Errorf("input column missing contractual")
+	}
+	key, err = nestedSchemaPropertyKey(contractualList[0])
+	if err != nil {
+		return "", "", err
+	}
+	return key, comparator, nil
+}
+
+func inputColumnSchemaPropertyKey(col interface{}) (string, error) {
```
sanjaysg21: I see this function is not used anywhere. I beleive the purpose of this func would be to extract just the schema_property_key from an input column. If its not used anywhere can we remove this?

brlenn: No this is no longer required, have removed it

sureshperiyappan (approved):

PR #2464, opened 2026-07-28
Bug/devtooling 1762

sureshperiyappan:


genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -310,7 +315,12 @@ func buildSubTypeSettings(d *schema.ResourceData, subTypeList []interface{}) *ma
 		subTypeSetting := platformclientv2.Messagesubtypesettings{
 			EnableAutoAnswer: &enableAutoAnswer,
 		}
-		subTypeSetting.EnableInactivityTimeout = resourcedata.GetNillableBool(d, fmt.Sprintf("media_settings_message.0.sub_type_settings.%d.enable_inactivity_timeout", i))
+		// The API only accepts EnableInactivityTimeout when set to true.
+		// Sending false explicitly causes a 400 error.
+		enableInactivityTimeout := resourcedata.GetNillableBool(d, fmt.Sprintf("media_settings_message.0.sub_type_settings.%d.enable_inactivity_timeout", i))
```
bbbco: Why is this using `"media_settings_message.0.sub_type_settings.%d.enable_inactivity_timeout"` instead of `subTypeMap["enable_inactivity_timeout"].(bool)`?

genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -217,7 +217,12 @@ func buildSdkMediaSettingsMessage(d *schema.ResourceData, settings []any) *platf
 		messageMediaSettings.SubTypeSettings = buildSubTypeSettings(d, subTypeSettingsList)
 	}
 
-	messageMediaSettings.EnableInactivityTimeout = resourcedata.GetNillableBool(d, "media_settings_message.0.enable_inactivity_timeout")
+	// The API only accepts EnableInactivityTimeout when set to true.
+	// Sending false explicitly causes a 400 error.
+	enableInactivityTimeout := resourcedata.GetNillableBool(d, "media_settings_message.0.enable_inactivity_timeout")
```
bbbco: Why is this using `"media_settings_message.0.enable_inactivity_timeout"` instead of `settingsMap["enable_inactivity_timeout"].(bool)`?

genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -217,7 +217,12 @@ func buildSdkMediaSettingsMessage(d *schema.ResourceData, settings []any) *platf
 		messageMediaSettings.SubTypeSettings = buildSubTypeSettings(d, subTypeSettingsList)
 	}
 
-	messageMediaSettings.EnableInactivityTimeout = resourcedata.GetNillableBool(d, "media_settings_message.0.enable_inactivity_timeout")
+	// The API only accepts EnableInactivityTimeout when set to true.
+	// Sending false explicitly causes a 400 error.
```
bbbco: This comment explains the behavior, but the reader has to infer from the code what the resulting logic is. We should be explicit in this comment about the expected logic:
```
// The API only accepts EnableInactivityTimeout when set to true.
// Sending false explicitly causes a 400 error, so we only set EnableInactivityTimeout if the value is true
```

bbbco (changes requested):

genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -217,7 +217,12 @@ func buildSdkMediaSettingsMessage(d *schema.ResourceData, settings []any) *platf
...
+	enableInactivityTimeout := resourcedata.GetNillableBool(d, "media_settings_message.0.enable_inactivity_timeout")
+	if enableInactivityTimeout != nil && *enableInactivityTimeout {
+		messageMediaSettings.EnableInactivityTimeout = enableInactivityTimeout
```
bbbco: Is there still concern about perpetual drift?

bbbco (changes requested):

genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
sureshperiyappan: No perpetual drift concern here. Terraform's zero value for an Optional TypeBool is false. When the API returns nil and the flattener skips the field (via SetMapValueIfNotNil), the state defaults to false, which matches the user's config. So plan won't detect a diff.

That said, if the field is not set in the user's config at all (omitted entirely), there's also no issue — Terraform won't track an attribute the user never specified (since it's not Computed). Both cases are clean.

bbbco (approved):

bbbco:
Thank you for this @sureshperiyappan !

PR #2491, opened 2026-08-11
DEVTOOLING-1757 fix: able to delete bullseye ring even if members are…

hemanth-genesys:
## Summary

Fix routing queue updates to clear member groups before removing bullseye rings, matching the UI behavior. Previously, removing a ring with members returned a 400 error and corrupted tfstate.

## Changes

- Modified `clearBullseyeRingMemberGroups` to issue a preliminary PUT clearing all member groups regardless of whether rings are partially or fully removed
- Simplified the caller in `updateRoutingQueue` to always call the clearing function when `bullseye_rings` changes (removed the `updateQueue.Bullseye == nil` condition)

## Root Cause

The API rejects removing a bullseye ring that has members attached. The provider only cleared member groups when ALL rings were removed (`Bullseye == nil`), not when individual rings were removed. The fix issues a prep PUT to detach members before reducing ring count — the same two-step approach the UI uses.

### DIFF
--- a/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
+++ b/genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
@@ -1148,6 +1148,11 @@ func flattenQueueWrapupCodes(ctx context.Context, queueID string, proxy *Routing
 	return nil, nil
 }
 
+// clearBullseyeRingMemberGroups clears member groups from bullseye rings before the main update.
+// The Genesys Cloud API rejects removing rings that have members attached. This function issues
+// a preliminary PUT that keeps the current ring count but sets MemberGroups to nil on all rings,
+// satisfying the API prerequisite. The caller's updateQueue.Bullseye (which holds the desired
+// final state — nil for all-rings-removed, or non-nil for partial removal) is left untouched.
 func clearBullseyeRingMemberGroups(ctx context.Context, d *schema.ResourceData, updateQueue *platformclientv2.Queuerequest, proxy *RoutingQueueProxy) diag.Diagnostics {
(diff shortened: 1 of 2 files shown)


genesyscloud/routing_queue/resource_genesyscloud_routing_queue_utils.go
```diff
@@ -1177,16 +1182,18 @@ func clearBullseyeRingMemberGroups(ctx context.Context, d *schema.ResourceData,
 		clearedRings[i].MemberGroups = nil
 	}
 
-	updateQueue.Bullseye = &platformclientv2.Bullseye{
-		Rings: &clearedRings,
+	clearQueue := platformclientv2.Queuerequest{
```
sureshperiyappan: The Queuerequest struct uses omitempty on all pointer fields. Your clearQueue only sets Name and Bullseye, which means the serialized JSON will omit mediaSettings, routingRules, acwSettings, skillEvaluationMethod, flows, and every other queue property.

The Genesys Cloud PUT endpoint treats omitted fields as "reset to default" — this is confirmed by the existence of addCGRAndOEAAndCGA (DEVTOOLING-751), which was added specifically to prevent fields managed by independent resources from being wiped during queue updates.

While the main update follows immediately with the full state, if it fails (network error, API 500, rate limit), the queue is left with all settings reset except Name and Bullseye.

Fix (one line change):


clearQueue := *updateQueue
clearQueue.Bullseye = &platformclientv2.Bullseye{
    Rings: &clearedRings,
}
This sends the full desired queue state with member groups cleared — safe regardless of what happens next.

hemanth-genesys: Done

sureshperiyappan (approved):

PR #2507, opened 2026-08-18
fix/DEVTOOLING-1530: Skip unrelated datatables during filtered export of architect datatable rows

hemanth-genesys:
**Summary**
Optimizes genesyscloud_architect_datatable_row export to skip fetching rows from datatables that don't match the include_filter_resources pattern.

Resolves: DEVTOOLING-1530

Problem
When exporting with a filter like:

include_filter_resources = ["genesyscloud_architect_datatable_row::Prioridades.ATR_AGENTES"]
The exporter fetched all rows from every datatable in the org, then applied the regex filter in memory. In orgs with 100+ tables and 40,000+ rows, this caused 10+ minute export times for a single row.

Solution
The export filter is now passed via context to getAllArchitectDatatableRows. The function fetches all datatable metadata (names and IDs , the process is cheap, no rows), then checks each table name against the filter patterns using a prefix match. Only tables whose name is a prefix of a filter pattern have their rows fetched.

For example, filter pattern Prioridades.ATR_AGENTES starts with table name Prioridades, so only that table's rows are fetched. Tables like DecoyTable are skipped entirely.

When no filter is present, behavior is unchanged — all tables and rows are fetched.


genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row.go
```diff
@@ -51,17 +51,64 @@ func getAllArchitectDatatableRows(ctx context.Context, clientConfig *platformcli
...
 		if err != nil {
-			return nil, util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to get architect Datatable Rows error: %s", err), resp)
+			return nil, util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to get architect Datatable Rows for table %q (id=%s): %s", *tableMeta.Name, *tableMeta.Id, err), resp)
+		}
+
+		rowCount := 0
+		if rows != nil {
+			rowCount = len(*rows)
 		}
+		log.Printf("[DEBUG] %s: fetched %d rows from datatable %q (id=%s)", ResourceType, rowCount, *tableMeta.Name, *tableMeta.Id)
```
sureshperiyappan, line 76: The code checks if rows != nil at line 108 for counting, but then unconditionally dereferences *rows at line 113. If rows is nil, this will panic.

// Current code (line 107-113):
rowCount := 0
if rows != nil {
    rowCount = len(*rows)
}
log.Printf(...)

for _, row := range *rows {  // <-- panics if rows is nil
You should guard the iteration with a nil check or use continue when rows is nil:


if rows == nil {
    continue
}
for _, row := range *rows {

genesyscloud/architect_datatable_row/genesyscloud_architect_datatable_row_utils.go
```diff
@@ -132,3 +136,41 @@ func getArchitectDatatableCached(ctx context.Context, tableID string, config *pl
...
+	for _, f := range filter {
+		if !strings.Contains(f, prefix) {
+			continue
+		}
+
+		pattern := f[strings.Index(f, "::")+2:]
+		if pattern == "" {
+			continue
+		}
+
```
sureshperiyappan, line 158: The comment says:

The filter pattern uses "." as the separator between table name and row key

But the actual BlockLabel is constructed as *tableMeta.Name + "_" + keyStr (line 116 in the resource file). Since the export filter uses the BlockLabel (e.g., genesyscloud_architect_datatable_row::MyTable_row1), the "." split won't work correctly. The separator between table name and row key is "_", not ".".

This means extractTableNamesFromFilter will never find a "." in the pattern and will treat the entire "TableName_RowKey" as the table name, which may never match any table from the API. The fallback to full fetch at line 89 saves you, but the optimization is effectively bypassed for most real-world filter values.

You either need to change the BlockLabel format to use "." as separator (e.g., TableName.RowKey), or change the parsing logic to split on "_". However, "_" is problematic because both table names and row keys can contain underscores. Consider using a different separator that's unlikely to appear in names (like ".") and update the BlockLabel construction to match.

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row.go
```diff
@@ -51,17 +51,64 @@ func getAllArchitectDatatableRows(ctx context.Context, clientConfig *platformcli
...
+					processedIds[*t.Id] = true
+					allTables = append(allTables, t)
+				}
+			}
+		}
+	}
+
+	// If name-based search returned 0 tables but we had a filter active, fall back to full fetch.
+	// This handles cases where the table name extraction was imprecise (e.g., table names with underscores).
+	if len(allTables) == 0 && len(filter) > 0 {
```
sureshperiyappan: if len(allTables) == 0 && len(filter) > 0 {
The condition should be checking len(tableNames) > 0 instead of len(filter) > 0. The filter slice contains entries for ALL resource types in the export (e.g., genesyscloud_user::john, genesyscloud_routing_queue::support). If no filter entries match genesyscloud_architect_datatable_row, tableNames will be empty and the code will already have taken the "fetch all" path (line 69: tableNames = []string{""}). But with the current condition, even when we're already fetching everything, if the filter had entries for other resource types, this fallback will re-fetch all tables a second time.

Use len(tableNames) > 0 instead of len(filter) > 0 to only trigger fallback when we actually attempted a narrowed search.

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row.go
```diff
@@ -51,17 +51,64 @@ func getAllArchitectDatatableRows(ctx context.Context, clientConfig *platformcli
...
-	if err != nil {
-		return nil, util.BuildAPIDiagnosticError(ResourceType, fmt.Sprintf("Failed to get architect datatables error: %s", err), resp)
+	// Check if an export filter is available in the context.
+	// If so, extract table name(s) from the filter and only fetch rows from matching tables
+	// to avoid unnecessary API calls to every datatable in the org.
+	filter := resourceExporter.ExportFilterFromContext(ctx)
+	tableNames := extractTableNamesFromFilter(ResourceType, filter)
+
+	if len(tableNames) > 0 {
+		log.Printf("[INFO] Export filter active for %s: narrowing to %d table name(s): %v", ResourceType, len(tableNames), tableNames)
```
sureshperiyappan: log.Printf("[INFO] Export filter active for %s: narrowing to %d table name(s): %v", ResourceType, len(tableNames), tableNames)
The repo convention is to keep INFO-level logs minimal and avoid dumping list contents. Table names could be considered sensitive org-specific information. Consider using [DEBUG] for the detailed list and keep the INFO log to just the count, consistent with how other resources in this repo log during export.

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row.go
hemanth-genesys: addressed

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row.go
hemanth-genesys: addressed

genesyscloud/architect_datatable_row/resource_genesyscloud_architect_datatable_row.go
hemanth-genesys: addressed

genesyscloud/architect_datatable_row/genesyscloud_architect_datatable_row_utils.go
hemanth-genesys: addressed by changing the logic to first get the table names and then match the table name present in the filter with the list of table names, and only fetch the rows of the matched table names. 

The previous approach doesnt work because there is no distinct seperator ('_' in our case that seperates DatatableName_DatatableRowName) that can be used to segregate the Datatable Name.

genesyscloud/architect_datatable_row/genesyscloud_architect_datatable_row_utils.go
```diff
@@ -132,3 +132,36 @@ func getArchitectDatatableCached(ctx context.Context, tableID string, config *pl
...
+		pattern := f[strings.Index(f, "::")+2:]
+		if pattern != "" {
+			patterns = append(patterns, pattern)
+		}
+	}
+
+	return patterns
+}
+
+// tableMatchesFilter checks if a table name could produce a BlockLabel that matches any of the filter patterns.
```
sanjaysg21, line 162: tableMatchesFilter ,strings.HasPrefix(pattern, tableName) will fail when patterns have regex anchors.
Filter like ^DEVTOOLING1530_LargeTable_row_0042$ won't match table name DEVTOOLING1530_LargeTable because of the leading ^. Table gets skipped, nothing gets exported.
Need to strip ^/$ before doing the prefix check.

sanjaysg21 (changes requested):

genesyscloud/architect_datatable_row/genesyscloud_architect_datatable_row_utils.go
hemanth-genesys: addressed using regex

sureshperiyappan (approved):

PR #2516, opened 2026-08-25
DEVTOOLING-1563 : Add genesyscloud_access_policy resource

hemanth-genesys:
## Summary

Adds a new Terraform resource `genesyscloud_access_policy` to manage Genesys Cloud Access Policies (ABAC - Attribute-Based Access Control).

## What's included

- **Resource**: `genesyscloud_access_policy` — Full CRUD support for access policies
- **Data Source**: `genesyscloud_access_policy` — Lookup existing policies by name
- **Exporter**: Support for CX-as-Code export
- **Acceptance Tests**: Resource CRUD and data source tests
- **Unit Tests**: Condition JSON parsing, flatten/build helpers

## Terraform usage

```hcl
resource "genesyscloud_access_policy" "example" {
  name             = "Cannot Grant New Roles"
  description      = "Prevents users from granting new roles"
  target_resource  = "authorization:grant:add"
  effect           = "DENY"
  subject_type     = "ALL"
  enabled          = true
  apply_to_clients = false
}

data "genesyscloud_access_policy" "lookup" {
  name = "Cannot Grant New Roles"
}

### CODE REVIEW

File: genesyscloud/access_policy/resource_genesyscloud_access_policy_schema.go [hunk 1]
```diff
@@ -0,0 +1,141 @@
...
+					// The API returns subject_id as "all" when subject_type is "ALL" and no explicit ID was provided.
+					// Suppress the diff in this case to avoid a perpetual plan.
+					subjectType := d.Get("subject_type").(string)
+					if new == "" && strings.EqualFold(old, subjectType) {
+						return true
+					}
+					return false
+				},
+			},
+			"condition_json": {
```
sanjaysg21, line 91: The API will likely return JSON with different key ordering or whitespace compared to what jsonencode() produces in HCL, you should add a diffsupress function similar to how its handled in business_rules_schema, routing_skill_group,etc.

genesyscloud/access_policy/resource_genesyscloud_access_policy_schema.go
```diff
@@ -0,0 +1,141 @@
...
+				Optional:    true,
+				Default:     false,
+			},
+			"apply_to_clients": {
+				Description: "Whether the policy also applies to OAuth Client (service account) API calls. Defaults to false.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+			},
+			"preset_attributes_json": {
```
sanjaysg21, line 109: follows the same problem as condition_json attribute

genesyscloud/access_policy/genesyscloud_access_policy_proxy.go
```diff
@@ -0,0 +1,200 @@
...
+
+	updatedPolicy, resp, err := p.authorizationApi.PutAuthorizationPolicy(id, *policy, false)
+	if err != nil {
+		return nil, resp, fmt.Errorf("failed to update access policy %s: %s", id, err)
+	}
+	return updatedPolicy, resp, nil
+}
+
+// deleteAccessPolicyFn is an implementation function for deleting a Genesys Cloud access policy
+func deleteAccessPolicyFn(ctx context.Context, p *accessPolicyProxy, id string, targetResource string, subjectId string) (response *platformclientv2.APIResponse, err error) {
```
sanjaysg21, line 194: The id parameter is accepted but unused in the actual API call the delete operates on targetResource + subjectId rather than the policy ID. Could you add a brief comment explaining this is by API design (e.g., target+subject is the natural key for deletion)? Just helps future readers understand why id only appears in the error message.

genesyscloud/access_policy/resource_genesyscloud_access_policy_utils.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mypurecloud/platform-client-sdk-go/v195/platformclientv2"
+)
+
+// buildAccessPolicyFromResourceData constructs an SDK Authorizationpolicy from Terraform ResourceData
+func buildAccessPolicyFromResourceData(d *schema.ResourceData) (*platformclientv2.Authorizationpolicy, error) {
```
sanjaysg21, line 12: d.GetOk("description") returns false for empty strings, so if a user removes their description (or sets it to ""), the field won't be included in the update request and the API will likely retain the old value. Can you use d.Get("description").(string) and always setting policy.Description so that clearing the field actually takes effect.

genesyscloud/access_policy/data_source_genesyscloud_access_policy_test.go
```diff
@@ -0,0 +1,81 @@
...
+					"ALL",
+					"",
+					"true",
+					"",
+				) + generateAccessPolicyDataSource(
+					dataLabel,
+					name,
+					"genesyscloud_access_policy."+resourceLabel,
+				),
+				PreConfig: func() {
```
sanjaysg21: Mostly other data source tests in the repo rely on the retry loop in the data source's ReadContext (120s with retry). The depends_on already forces resource-creation ordering. The sleep is likely unnecessary, the retry loop handles eventual consistency. If it's needed for this specific API, a comment explaining why the standard retry isn't sufficient would help.

sanjaysg21 (changes requested):

genesyscloud/access_policy/resource_genesyscloud_access_policy_schema.go
hemanth-genesys: addressed

genesyscloud/access_policy/resource_genesyscloud_access_policy_schema.go
hemanth-genesys: addressed

genesyscloud/access_policy/genesyscloud_access_policy_proxy.go
hemanth-genesys: added comment

genesyscloud/access_policy/resource_genesyscloud_access_policy_utils.go
hemanth-genesys: addressed

genesyscloud/access_policy/data_source_genesyscloud_access_policy_test.go
hemanth-genesys: Removed the time.Sleep(30 * time.Second) and the PreConfig block entirely.

sureshperiyappan (approved):

Issue #2437, opened 2026-07-13
Duplicate Integrations causing interaction actions to force replace


atreyayelishetti:
Hi,

For some reason Genesys Cloud allows you to create integration with same Name in same Integration Type.
And when Data block is used to query for Id of a integration and reference it in data actions, this some times causes to find different integration than one it was initially referenced and created with, which makes data action to force replace delete and create data action which breaks the flows where they are referenced in.

Ideal fix should be in Integrations, it shouldn't allow to create integration with same name in same integration type and also data query block should ask for Integration Type as well when doing data query block

hemanth-genesys:
Hi @atreyayelishetti 
I have added an optional **integration_type** filter to the genesyscloud_integration data source, which resolves cases where multiple integrations share a name but have different types ( this will be released in the upcoming release version - which I will update you after it is done). For the case where integrations share both the same name and the same type, the data source has no unique signal to distinguish them, so it can't reliably resolve a single one.

I raised this issue with the service team and they have confirmed duplicate names are allowed by design and enforcing uniqueness would be a breaking change. For that specific scenario, the recommended approach is to reference the integration directly by its ID rather than through the name-based data source, since IDs are guaranteed unique.

Let me know if you need any info further !

Thanks
Hemanth

PR #2517, opened 2026-08-25
Add integration_type filter to genesyscloud_integration data source

hemanth-genesys:
Addresses a part of #2437

## Problem
When multiple integrations share the same name (but different types), the
`genesyscloud_integration` data source returns the first match found during
pagination with no way to disambiguate. This can cause data actions to
resolve to the wrong integration, triggering force replacements that break
flows.

## Solution
Added an optional `integration_type` field to the `genesyscloud_integration`
data source schema. When provided, the lookup filters by both name and type,
ensuring the correct integration is returned.

## Changes
- Added `integration_type` optional attribute to `DataSourceIntegration()` schema
- Added `getIntegrationByNameAndTypeFn` in the proxy to look up by name + type
- Updated `dataSourceIntegrationRead` to use the new function when `integration_type` is provided
- Existing behavior (lookup by name only) is unchanged when `integration_type` is not set

### DIFF
--- a/genesyscloud/integration/data_source_genesyscloud_integration.go
+++ b/genesyscloud/integration/data_source_genesyscloud_integration.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/mypurecloud/platform-client-sdk-go/v195/platformclientv2"
 	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/provider"
 	"github.com/mypurecloud/terraform-provider-genesyscloud/genesyscloud/util"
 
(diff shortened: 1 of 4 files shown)


genesyscloud/integration/data_source_genesyscloud_integration.go
```diff
@@ -21,14 +22,26 @@ import (
...
 
 	return util.WithRetries(ctx, 15*time.Second, func() *retry.RetryError {
-		integration, retryable, resp, err := ip.getIntegrationByName(ctx, integrationName)
+		var integration *platformclientv2.Integration
+		var retryable bool
+		var resp *platformclientv2.APIResponse
+		var err error
+
+		if integrationType != "" {
+			integration, retryable, resp, err = getIntegrationByNameAndTypeFn(ctx, ip, integrationName, integrationType)
```
shrutisuryawanshigenesys: this should be called via proxy not the 'Fn' implementation

shrutisuryawanshigenesys (changes requested):

shrutisuryawanshigenesys:
- 'examples/data-sources/genesyscloud_integration/data-source.tf' also needs to be updated with integration_type
- add test coverage for 'name' + 'integration_type' lookup

genesyscloud/integration/data_source_genesyscloud_integration.go
hemanth-genesys: done

hemanth-genesys:
> * 'examples/data-sources/genesyscloud_integration/data-source.tf' also needs to be updated with integration_type
> * add test coverage for 'name' + 'integration_type' lookup

done

PR #2518, opened 2026-08-25
Fix pagination for responsemanagement_response data source with library_id

sanjaysg21:
getAllResponsemanagementResponseFn only fetched page 1 (100 items) when a libraryId was provided, causing lookups to fail for libraries with >100 responses. Added pagination loop matching the existing non-library path.

### DIFF
--- a/genesyscloud/responsemanagement_response/genesyscloud_responsemanagement_response_proxy.go
+++ b/genesyscloud/responsemanagement_response/genesyscloud_responsemanagement_response_proxy.go
@@ -114,17 +114,21 @@ func getAllResponsemanagementResponseFn(ctx context.Context, p *responsemanageme
 	const pageSize = 100
 
 	if libraryId != "" {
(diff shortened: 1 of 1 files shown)


genesyscloud/responsemanagement_response/genesyscloud_responsemanagement_response_proxy.go
```diff
@@ -114,17 +114,21 @@ func getAllResponsemanagementResponseFn(ctx context.Context, p *responsemanageme
 	const pageSize = 100
 
 	if libraryId != "" {
-		responses, resp, getErr := p.responseManagementApi.GetResponsemanagementResponses(libraryId, 1, pageSize, "")
-		if getErr != nil {
-			return nil, resp, fmt.Errorf("Error requesting page of Responsemanagement Response: %s", getErr)
-		}
+		var resp *platformclientv2.APIResponse
```
shrutisuryawanshigenesys, line 117: this loop is duplicated below (~146, ~173); a shared getResponsesForLibrary helper would help.

Issue #2417, opened 2026-07-07
Since 1.82.0, genesyscloud_integration only includes 'config' attribute if export_computed=true


elmaimbo:
I recently upgraded to the latest version of the Genesys Cloud provider, and I've discovered something that looks like a bug, that has been introduced in version 1.82.0...

To demonstrate, this is the configuration I'm testing with:
```
terraform {
  required_providers {
    genesyscloud = {
      source = "MyPureCloud/genesyscloud"
      version = "1.81.0"
    }
  }
}

resource "genesyscloud_tf_export" "export" {
  directory = "./export"
  export_format = "hcl"
  include_state_file = false
  include_filter_resources = [
    "genesyscloud_integration::^XXX_Genesys Cloud Public API Integration$",
  ]
  enable_dependency_resolution = false
  export_computed = false
}
```

Running the above as is (specifying `version = "1.81.0"`) results in the following export being generated (note `X`s used to redact customer-specific info):
```
terraform {
  required_providers {
    genesyscloud = {
      source  = "registry.terraform.io/mypurecloud/genesyscloud"
      version = "1.81.0"
    }
  }
}

resource "genesyscloud_integration" "XXX_Genesys_Cloud_Public_API_Integration" {
  config {
    advanced = jsonencode({		})
    credentials = {
      pureCloudOAuthClient = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
    }
    name       = "XXX_Genesys Cloud Public API Integration"
    properties = jsonencode({		})
  }
  integration_type = "purecloud-data-actions"
  intended_state   = "ENABLED"
}
```

As you can see the `config` attribute is populated with values, including the name of the integration.

However if I change the original snippet to specify `version = "1.82.0"` (instead of `version = "1.81.0"`), and re-run, the export looks like this:

```
terraform {
  required_providers {
    genesyscloud = {
      source  = "registry.terraform.io/mypurecloud/genesyscloud"
      version = "1.82.0"
    }
  }
}

resource "genesyscloud_integration" "XXX_Genesys_Cloud_Public_API_Integration" {
  integration_type = "purecloud-data-actions"
  intended_state   = "ENABLED"
}
```

As you can see the `config` block has disappeared. You can get it to work by setting `export_computed = true`, but given that the `name` and `credentials` aren't 'computed properties', this seems like a bug?

Thanks,
Nick.

PR #2554, opened 2026-09-07
DEVTOOLING-1810: Preserve user config on export when export_computed=…

Teja-genesys:
**Reported externally:** GitHub #2417

A customer exports genesyscloud_integration with export_computed = false and found that the entire config block disappears from the export — including values they set themselves (name, credentials), not just platform-generated values.

How it worked before (1.81.0)
The exporter's computed-attribute removal only ran on the top-level, flat state attributes (delete(instanceState.Attributes, "config")). Since Terraform stores a nested block using compound keys (config.0.name, config.0.notes, …), there is no key literally called config, so this delete was a no-op for any nested block. As a side effect, config and everything inside it survived untouched:

hcl

resource "genesyscloud_integration" "example" {
  config {
    advanced    = jsonencode({})
    credentials = { pureCloudOAuthClient = "..." }
    name        = "My Integration"
    properties  = jsonencode({})
  }
  integration_type = "purecloud-data-actions"
  intended_state   = "ENABLED"
}

What changed (1.82.0, PR #2337)
That PR (unrelated goal: exclude deprecated fields) refactored the exclusion logic into a recursive function (collectSchemaBasedExcludedAttributes) and moved the removal onto the nested config map (removeExcludedAttrsFromMap), which deletes by prefix match (config + config.*). Because config itself is marked Computed: true (Optional+Computed, meaning "the user may omit it and the platform will fill it in"), the new code excluded the whole config path and never looked at its children — so removeExcludedAttrsFromMap wiped the entire block:

hcl

resource "genesyscloud_integration" "example" {
  integration_type = "purecloud-data-actions"
  intended_state   = "ENABLED"
}
name, notes, and credentials are plain user input (confirmed in resource_genesyscloud_integration_utils.go: they're read from the Terraform config and PUT to the API, never generated). Per our own export_computed docs, the flag is only meant to exclude attributes that are "Computed and Optional" — notes and credentials are Optional only, so they should never have been removed.

**What we did**
Two coordinated fixes in the shared exporter code (
genesyscloud_resource_exporter.go
), both gated to export_computed=false so the default (true) export path is completely unchanged:

Don't exclude a nested block wholesale for being Computed. collectSchemaBasedExcludedAttributes now recurses into it and excludes only the individually-computed children. If every child in the block's schema is computed, the whole block is still collapsed (avoids an empty schema-level shell).

Suppress a block that ends up with no real values. After sanitization, removeAllNilNestedBlocks removes any block whose fields are all nil (e.g. user left notes empty and set no credentials) so an empty config {} shell isn't emitted. This is intentionally conservative — it only treats literal nil as empty, so it never touches AllowEmptyArrays non-nil empty slices, RemoveIfMissing handling, or any scalar array. Verified with dedicated unit tests plus the full existing suite (including TestUnitTfExportAllowEmptyArray).

**Before / after examples**
User sets notes, export_computed=false:

hcl

** Before (bug)**
resource "genesyscloud_integration" "example" {
  integration_type = "purecloud-data-actions"
  intended_state   = "DISABLED"
}

** After (fixed)**
resource "genesyscloud_integration" "example" {
  config {
    notes = "user-set notes that should survive export"
  }
  integration_type = "purecloud-data-actions"
  intended_state   = "DISABLED"
}
User sets nothing (all fields end up empty/computed), export_computed=false:

hcl

**# Before (bug): empty shell**
resource "genesyscloud_integration" "example" {
  config {
  }
  integration_type = "purecloud-data-actions"
  intended_state   = "DISABLED"
}

**# After (fixed): block omitted entirely**
resource "genesyscloud_integration" "example" {
  integration_type = "purecloud-data-actions"
  intended_state   = "DISABLED"
}
export_computed=true (default) — unchanged in both before and after:

hcl

resource "genesyscloud_integration" "example" {
  config {
    name       = "example"
    properties = jsonencode({})
    advanced   = jsonencode({})
  }
  integration_type = "purecloud-data-actions"
  intended_state   = "DISABLED"
}
**Blast radius**
This is common exporter code, so it affects every resource with an Optional+Computed nested block (not just genesyscloud_integration) — e.g. routing_settings, outbound_campaign, user, routing_utilization, outbound_callanalysisresponseset. All of them had the same latent bug and are fixed the same way. export_computed=true behavior is untouched for all resources.

**Testing**
Unit tests: full tfexporter package suite passes, including new coverage for both fixes and 9 dedicated cases for the empty-block suppression (real map/string values kept, AllowEmptyArrays respected, scalar arrays untouched, mixed lists, no-block resources unaffected).
go vet: clean.
Manual: reproduced the original bug and the empty-block edge case live against a purecloud-data-actions integration, confirmed both are fixed and export_computed=true output is identical to before.


genesyscloud/tfexporter/genesyscloud_resource_exporter.go
```diff
@@ -2281,12 +2304,44 @@ func (g *GenesysCloudResourceExporter) collectSchemaBasedExcludedAttributes(reso
 			continue
 		}
 		if elem, ok := s.Elem.(*schema.Resource); ok {
-			excludedAttributes = append(excludedAttributes, g.collectSchemaBasedExcludedAttributes(resourceType, elem.Schema, fullPath)...)
+			childExcluded := g.collectSchemaBasedExcludedAttributes(resourceType, elem.Schema, fullPath)
```
sanjaysg21, line 2307: This collapse branch runs for every nested block regardless of exportComputed. When export_computed=true but export_deprecated=false, a block whose children are all deprecated now collapses to excluding the whole block path instead of the individual leaves. That changes the export_computed=true output (previously an empty block {} shell, now the block is omitted), which contradicts the PR statement that the default path is unchanged. Consider gating the collapse on !g.exportComputed, or updating the description + adding a test for the export_computed=true, export_deprecated=false case.

Teja-genesys: Fixed — gated the collapse to !g.exportComputed. Under export_computed=true it now keeps the old behavior (empty shell, only the leaf excluded). Added tests for the deprecated-only and read-only-only cases under both settings, plus nested/TypeSet variants.

PR #2561, opened 2026-09-14
feat: JM-6828 CX as Code - Journey View Chart rank

bencegy:
Add Journey View chart rank to CX as Code


nicob-29 (approved):

docs/resources/user.md
```diff
@@ -52,6 +52,7 @@ The following permissions are required to use this resource:
 * `directory:user:add`
 * `directory:user:delete`
 * `directory:user:edit`
+* `directory:user:hardDelete`
```
bencegy, line 55: this was added by the command "go generate", not by me

PR #2562, opened 2026-09-14
PLUS-2895: Unhide Division ID

eclancy404:


genesyscloud/routing_skill/resource_genesyscloud_routing_skill.go
```diff
@@ -55,20 +60,6 @@ func createRoutingSkill(ctx context.Context, d *schema.ResourceData, meta interf
 
 	d.SetId(*skill.Id)
 	log.Printf("Created skill %s %s", name, *skill.Id)
-
-	// If a division_id is specified, move the skill to that division via PATCH
-	divisionId := d.Get("division_id").(string)
```
eclancy404, line 60: Removing the extra api call now since the public-api has been updated

sanjaysg21 (approved):

PR #2568, opened 2026-09-16
AS-5605: Add all_outbound_email_addresses to genesyscloud_routing_queue

RamKeerthy1006:
[AS-5605](https://inindca.atlassian.net/browse/AS-5605)

## Why
Customers with **multiple outbound email domains/routes** on a queue could not promote it via CX-as-Code. The `genesyscloud_routing_queue` resource only supported a single `outbound_email_address` (`MaxItems: 1`), so on promotion the second address was silently dropped. The API already exposes the full list via `MediaSettings.Email.allOutboundEmailAddresses` (available in Go SDK v199, now on `dev`).

## What
- Added a new top-level **`all_outbound_email_addresses`** block on `genesyscloud_routing_queue`:
  - `TypeList` of `{ domain_id, route_id }`, **`Optional` + `Computed`**, no `MaxItems` (supports multiple).
  - Wired through `MediaSettings.Email.AllOutboundEmailAddresses` on both read and write.
  - Added exporter ref attrs (`domain_id` → email domain, `route_id` → email route) + `RemoveIfMissing`.
- Existing `outbound_email_address` block is **left untouched** (additive, non-breaking).
- Regenerated `docs/resources/routing_queue.md`.

## Impact
- **Non-breaking / additive** — new attribute is optional; existing configs unaffected; no state migration.
- **`Computed`** prevents perpetual plan diffs when the org has API-populated addresses the user didn't set in `.tf`.
- Read is anchored at the top level (mirrors `outbound_email_address`); the shared `queueMediaSettingsResource` for `media_settings_email`/`_call`/`_chat` is untouched.
- **Depends on** the SDK v195→v199 bump (already merged to `dev`).
- **Rollback:** revert this commit; no data/state migration.

## Testing
- `go build ./...` — clean
- `go vet ./genesyscloud/routing_queue/...` — clean
- `gofmt -l` — clean
- `go test ./genesyscloud/routing_queue/ -run TestUnit` — **PASS** (fixture extended with two `AllOutboundEmailAddresses`; flatten round-trip asserted in the state map)
- `go generate ./...` — regenerated docs; only `routing_queue.md` changed

## Files Changed
- `resource_genesyscloud_routing_queue_schema.go` — new `all_outbound_email_addresses` block + exporter refs + RemoveIfMissing
- `resource_genesyscloud_routing_queue_utils.go` — `buildSdkAllOutboundEmailAddresses` (build), `flattenAllOutboundEmailAddresses` (read), wired into `buildSdkMediaSettings`
- `resource_genesyscloud_routing_queue.go` — read path sets the attribute from `MediaSettings.Email.AllOutboundEmailAddresses`
- `resource_genesyscloud_routing_queue_unit_test.go` — fixture + flatten assertion
- `docs/resources/routing_queue.md` — regenerated

[AS-5605]: https://inindca.atlassian.net/browse/AS-5605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
```diff
@@ -814,6 +814,26 @@ func ResourceRoutingQueue() *schema.Resource {
 					},
 				},
 			},
+			"all_outbound_email_addresses": {
+				Description: "The list of all outbound email addresses (domain + route) assigned to the queue. Supports multiple email domains/routes, unlike the deprecated single outbound_email_address block. Requires the multiple outbound email addresses feature to be enabled on the org.",
+				Type:        schema.TypeList,
```
sanjaysg21, line 866: TypeList is order-sensitive. Since this is Computed and populated from an API list, if the API doesn't return addresses in the same order the user declared them in .tf, users will get a permanent plan diff once they set addresses themselves (Computed only masks the empty case). I think the address order is not guaranteed by the API as well. An acceptance test for this would prove it.

sanjaysg21 (approved):

genesyscloud/routing_queue/resource_genesyscloud_routing_queue_schema.go
RamKeerthy1006: Thanks for the catch on the ordering — I added an acceptance test to prove it out (`TestAccResourceRoutingQueueAllOutboundEmailAddresses`) and ran it against a live DCA org.

**Result: the reorder diff does not occur.** The test creates a queue with two addresses, then (a) re-applies the same config and (b) re-applies with the declared order swapped — both steps pass the framework's post-apply no-op plan check, so there's no perpetual diff even when the declared order differs from what the API returns. So `TypeList` is fine here and no `TypeSet`/`DiffSuppressFunc` change is needed. (Worth noting this resource deliberately avoids `TypeSet` elsewhere — `members`/`wrapup_codes` — due to the gRPC TypeSet state-corruption issue in DEVTOOLING-1533, so keeping `TypeList` is also the safer choice.)

Verifying against the API also surfaced a real contract I've now handled: the default `outbound_email_address` must be a member of `all_outbound_email_addresses` (the API returns `400 - Default outbound email address missing from outbound email address list` otherwise). I added a `CustomizeDiff` that catches this at plan time with a clear message, plus a docs note. Also confirmed end-to-end in the DCA portal — both addresses show on the queue with the correct default.

Pushed in the latest commit. Thanks again!

